PR #15, opened 2019-02-07
Add installer flag

philipgough:
@knrc Feel free to dismiss this if not relevant. I was looking for away to make changes to the installer image and test without having to have a generic prefix that was forcing me to build all images.

Maybe I overlooked something and there was already a way.

### DIFF
--- a/cmd/istio-operator/main.go
+++ b/cmd/istio-operator/main.go
@@ -9,9 +9,9 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/util/k8sutil"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 
+	"flag"
 	"github.com/sirupsen/logrus"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
-	"flag"
 )
 
 func printVersion() {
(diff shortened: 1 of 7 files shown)


knrc:
@PhilipGough We made a deliberate decision to tie the installer to the images but we could certainly look at something like this.  You shouldn't need to rebuild all the containers to test though.

philipgough:
@knrc Do they need to be tagged and pushed? So if I override the prefix and run a custom installer image it seemed to deploy all the istio images with the same prefix, can this be avoided?

knrc:
@PhilipGough Are you running istiooc to test?

philipgough:
Nope, workflow was create the cluster with the Maistra release (`istiooc cluster up`), modify the operator image and inject the prefix.

knrc:
@PhilipGough If you are using istiooc then you can do it locally without having to rebuild everything.  The default behaviour of oc is to specify a 'pull always' policy but this can be overridden so your local images take precedence.  I'll send you some details later today.

philipgough:
@knrc Cheers that would be great. FYI this was an ec2 machine and I'm on os x if that changes things.

PR #36, opened 2019-04-12
updates for sidecar patches and 3scale version

rcernich:

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -208,10 +208,12 @@ rules:
   - watch
 - apiGroups:
   - extensions
+  - apps
   resources:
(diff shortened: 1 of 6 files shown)


rcernich:
@knrc, fyi, updates for the sidecar patches and 3scale version update (0.5.0).  Note, updating an Installation results in the same configuration being installed, as the ConfigMap used by the installer does not get update.  I tested with 0.9 and it exhibits the same behavior (fails when trying to create the new ConfigMap).  I left this alone, as it was an existing bug (i.e. not new to the 0.10 operator).

knrc:
@rcernich Updating the installation has worked in previous releases, how are you reproducing this?

PR #38, opened 2019-04-13
change watch namespace variable to istio-system

itsmurugappan:
changing the watch namespace environment variable to istio-system in operator.yaml file .


rcernich:
This has been addressed in subsequent releases.  Thanks for the PR though.

PR #52, opened 2019-05-16
Fix Kiali deployment

jwendell:
Kiali deployment on upstream already contains kiali-secret stuff.

### DIFF
--- a/helm/istio/charts/kiali/templates/secret.yaml
+++ b/helm/istio/charts/kiali/templates/secret.yaml
@@ -10,11 +10,12 @@ metadata:
     chart: {{ template "kiali.chart" . }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
+type: Opaque
 data:
 {{ if .Values.dashboard.user }}
-  {{ .Values.dashboard.usernameKey }}: {{ b64enc .Values.dashboard.user }}
+  username: {{ b64enc .Values.dashboard.user }}
 {{ end }}
 {{ if .Values.dashboard.passphrase }}
-  {{ .Values.dashboard.passphraseKey }}:  {{ b64enc .Values.dashboard.passphrase }}
+  passphrase: {{ b64enc .Values.dashboard.passphrase }}
 {{ end }}
 {{ end }}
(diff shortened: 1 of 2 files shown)


jwendell:
Not sure if it resolves completely MAISTRA-351, but at least now Kiali deploys successfully and dashboard shows up on `https://<route url>/kiali`

helm/istio/charts/kiali/templates/secret.yaml
```diff
@@ -10,11 +10,12 @@ metadata:
     chart: {{ template "kiali.chart" . }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
+type: Opaque
 data:
 {{ if .Values.dashboard.user }}
-  {{ .Values.dashboard.usernameKey }}: {{ b64enc .Values.dashboard.user }}
```
rcernich, line 15: We should only be doing this if we also hard code the `username` value in the deployment.yaml (or kiali CR).

helm/istio/charts/kiali/templates/secret.yaml
```diff
@@ -10,11 +10,12 @@ metadata:
...
+  username: {{ b64enc .Values.dashboard.user }}
 {{ end }}
 {{ if .Values.dashboard.passphrase }}
-  {{ .Values.dashboard.passphraseKey }}:  {{ b64enc .Values.dashboard.passphrase }}
```
rcernich, line 18: Same as above, but for `passphrase`.

tmp/build/patch-charts.sh
```diff
@@ -355,23 +355,14 @@ function patchKialiOpenShift() {
...
-\          mountPath: "/kiali-secret"\
-\{\{- end \}\}
+\          mountPath: "/kiali-cert"
     }
     /configMap:/ {
       N
       a\
 \      - name: kiali-cert\
 \        secret:\
-\          secretName: kiali-cert-secret\
```
rcernich, line 369: We should not be doing this as `oauth` method is configured if either of these are missing.  If they're both specified, we use `login` configuration.  Changing this will alter how auth is configured.

rcernich (changes requested):

PR #54, opened 2019-05-21
MAISTRA-425 Add PodLocality controller

luksa:
Istio needs to know the zone and region of each pod. Since that
information is not stored on the Pod, but on the Node object, it would
need read privileges to the cluster-scoped Node object.

The idea of the PodLocality controller is to copy the region and zone
labels from the Node to the Pod, allowing Istio to get the info directly
from the Pod.


pkg/controller/podlocality/controller.go
```diff
@@ -0,0 +1,164 @@
...
+
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("podlocality-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	err = c.Watch(&source.Kind{Type: &v1.Pod{}}, &handler.EnqueueRequestForObject{}, predicate.Funcs{
```
rcernich, line 57: Should this be scoped to only istio related pods?

luksa: The controller will now only process pods with the `sidecar.istio.io/status` annotation.

pkg/controller/podlocality/controller.go
```diff
@@ -0,0 +1,181 @@
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object
+		return reconcile.Result{}, err
+	}
+
+	if _, hasSidecar := pod.Annotations[IstioSidecarStatusAnnotation]; !hasSidecar {
```
rcernich: Should this be moved into the filter predicate funcs?

luksa: Would still need to be here because I'm enqueuing pods whenever a node's zone/region labels change. But I could add the same check to the filter predicates.

rcernich: You could just not queue those pods, placing the annotation check in the node watcher too.

pkg/controller/podlocality/controller.go
```diff
@@ -0,0 +1,189 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	if pod.Labels[NodeRegionLabel] == node.Labels[NodeRegionLabel] && pod.Labels[NodeZoneLabel] == node.Labels[NodeZoneLabel] {
+		reqLogger.Info("Pod's locality labels match the node's. Nothing to do.")
+		return reconcile.Result{}, nil
+	}
+
+	pod.Labels[NodeRegionLabel] = node.Labels[NodeRegionLabel]
+	pod.Labels[NodeZoneLabel] = node.Labels[NodeZoneLabel]
```
luksa, line 179: @rcernich Do you think I should add these to the pod as annotations instead of labels?

pkg/controller/podlocality/controller.go
luksa: Ah, of course. I forgot I was already listing pods there, so I was able to easily add that there, yes.

pkg/controller/podlocality/controller.go
rcernich: good question.  they'd be hidden a bit more as annotations, but labels are easier to use in selectors.  i don't know.

labels do have limitations on values, but if these are coming from other labels, it's not really a concern.

rcernich (approved):

luksa:
Squashed & resolved conflict. I've now removed the `istio-pilot-mesh-` cluster role, since it is now empty.

rcernich:
Do we also need to remove the mesh cluster role/binding for galley as well?

rcernich:
And we should remove the requirement for nodes from the operator's cluster role.

luksa:
The pod locality controller is in the operator, so it still needs access to nodes.

As for Galley, it looks like it doesn't even read nodes yet. It'll probably read them when Pilot starts reading nodes from Galley instead of from K8s directly.

rcernich:
Got it, thanks!

pkg/controller/podlocality/controller.go
```diff
@@ -0,0 +1,189 @@
...
+
+	err = r.Client.Update(context.TODO(), pod)
```
dgn, line 181: This will sometimes fail trying to update a stale resource. I'm still trying to figure out what's changing the Pod inbetween the `Get()` and `Update()`

dgn: I suppose the only way to be 100% sure is to implement a Get()-Update()-Retry() loop as described here: https://github.com/kubernetes/client-go/blob/396a06da3bd7e3b0cd19b64a1da70fe1142fbff2/examples/create-update-delete-deployment/main.go#L117

This is a minor issue though, as we're not logging as error and the controller immediately applies the labels in the next reconcile:
```
2019-08-09T08:41:52.753Z	INFO	controller_podlocality	podlocality/controller.go:143	Processing Pod	{"Request.Namespace": "test", "Request.Name": "sleep-84d98cdb98-2ktqm"}
2019-08-09T08:41:52.782Z	INFO	controller_podlocality	podlocality/controller.go:183	Error updating pod's labels: Operation cannot be fulfilled on pods "sleep-84d98cdb98-2ktqm": the object has been modified; please apply your changes to the latest version and try again	{"Request.Namespace": "test", "Request.Name": "sleep-84d98cdb98-2ktqm"}
2019-08-09T08:41:53.783Z	INFO	controller_podlocality	podlocality/controller.go:143	Processing Pod	{"Request.Namespace": "test", "Request.Name": "sleep-84d98cdb98-2ktqm"}
2019-08-09T08:41:53.798Z	INFO	controller_podlocality	podlocality/controller.go:187	Successfully added zone and region labels to pod.	{"Request.Namespace": "test", "Request.Name": "sleep-84d98cdb98-2ktqm"}
```

luksa: Keep in mind that `get()` doesn't go to the API server - it only gets the object from the local cache. `Update()`, on the other hand, does go straight to the API Server (but *doesn't* update the local cache!). 

So, there's always a chance for discrepancies. 

Here, some other controller (most likely the kubelet) is also updating the pod (since this is happening during pod creation), so there's a race between them. Getting a 409 conflict is therefore perfectly normal and needs to be handled properly. 

On the get-update-retry loop: most core k8s controllers just try to retry once. IMO, in our case, there's no need to do that. We return the error and the reconciler gets called again (with proper backoff). This is superior because it doesn't block the reconciler from reconciling other pods. If another pod was enqueued in the meantime, it gets reconciled first, and only then is the first pod reconciled again. The get-update-retry loop in reconcile() itself would prevent other pods from being reconciled while waiting for the backoff delay to expire. 

And there's one other reason why using get-update-retry in our operator doesn't work: the controller-runtime framework we're using directs all reads to the cache. This means that performing a get after a failed update won't help, since 99% of the time, we'll get the same stale version of the object (unless the cache is updated by an incoming watch event in the meantime).

dgn:
In the YAML files under `deploy/` we are adding the `node` permissions, but there's a comment saying:
```
# to be removed
- apiGroups:
  - ""
  resources:
  - nodes
```

I think this should be changed to `# required by podlocality controller` or similar

luksa:
> I think this should be changed to `# required by podlocality controller` or similar

Done. Thanks!

pkg/controller/podlocality/controller.go
dgn: Makes sense. Thank you for the explanation!

dgn (approved):

PR #56, opened 2019-05-21
Updated Kiali version

brian-avery:


brian-avery:
@rcernich Looks like patch.sh is used by the rpm script and that this version change requires no changes to the charts here. Please correct me if I'm wrong.

PR #72, opened 2019-05-28
Update maistra-operator.yaml

brian-avery:


rcernich (changes requested):
This also needs to be done for the servicemesh-operator.yaml

brian-avery:
Copied. Thanks. Thoughts? Right name, wrong name? ServiceMesh... gets pretty long to type.

rcernich (approved):

PR #74, opened 2019-05-29
MAISTRA-467 Prevent reconciling namespace in mesh when it's already in another mesh

luksa:


knrc:
This needs to be done as part of the validation of the SMMR

knrc:
-1 to merge

rcernich (approved):

rcernich:
@knrc, while this should be handled through validation, is it still a good idea to prevent the operator from trying to reconcile two meshes into the same namespace?  I think this hardens the operator.

jwendell:
Note that the base branch of this PR is 0.11

rcernich:
@jwendell, thanks for catching that!

luksa:
@knrc I've resolved the conflicts.

knrc:
@luksa Thanks, I'll merge it now

PR #78, opened 2019-05-30
MAISTRA-470 ignore control plane namespace, if specified, when proces…

rcernich:
…sing ServiceMeshMemberRoll


rcernich:
this needs to be targeted to maistra-0.12 once that branch is created

PR #87, opened 2019-06-14
MAISTRA-369 use Kiali Operator

jmazzitelli:
https://issues.jboss.org/browse/MAISTRA-369

Have Maistra use the Kiali Operator when installing Kiali.


jmazzitelli:
This is only a WIP right now. I am creating a draft PR now so I can collaborate with the Maistra team to find out if I'm hitting all the touch points. I have not even tried running or compiling yet.

One thing that I know is missing is the Kiali CR needs its `deployment.accessible_namespaces` list updated when the ServiceMesh CR is created with the list of namespaces in the mesh.

jmazzitelli:
The one question I have is how are we expecting dependent operators to be installed?
This PR currently just has the Kiali CR in the kiali helm chart templates directory (along with the secret). The Kiali Operator CRD and its required roles, service accounts, etc are next to the maistra-operator/servicemesh-operator yaml files in the deploy/ directory - see: https://github.com/Maistra/istio-operator/pull/87/files#diff-4a3563d84d1834c255a7c807f9bfa122

But I don't know how/who will be installing these - I assume the person/process that installs the maistra operator CRD yaml will also install all other CRDs that are needed (which now include the Kiali CRD for the operator CR and for kiali dashboard CR).

If that is not how it is going to work, then this PR will need to change - the `charts/kiali/templates` directory will need to include the necessary clusterroles, service accounts, etc needed for the Kiali *operator* (not Kiali itself, those yaml template files have already been removed via this PR). But based on how I understand how things are going to work, I don't think we want to do it this way. I think we want to manage the kiali operator CRD in the same way as the maistra CRD.

I'll need input from the Maistra team here as for the proper direction this needs to go.

jmazzitelli:
cc @rcernich

jmazzitelli:
As the PR exists right now, Kiali Operator is integrated and working with Maistra's istio operator - still need some things to do (need Istio operator to edit the `accessible_namespaces` setting in the Kiali CR for one).

From the root dir, I run the build with my own IMAGE name:

```
IMAGE=jmazzitelli/istio-operator ./tmp/build/docker_build.sh 
```
I then edit `deploy/maistra-operator.yaml` so it pulls in my own local image for testing:

```
- name: istio-operator
  image: jmazzitelli/istio-operator:latest
...
  imagePullPolicy: Never
```
    
I then run this from the `deploy/` directory:

```
oc create namespace kiali-operator
oc create namespace istio-operator
oc create namespace istio-system
oc create -f kiali-operator.yaml -n kiali-operator
oc create -f maistra-operator.yaml -n istio-operator
oc create -f examples/maistra_v1_servicemeshcontrolplane_cr_basic.yaml -n istio-system
```

rcernich:
@jmazzitelli, @dmitri-d  is working on integrating installation of the operators.  My two cents is that we should not be copying the kiali operator install files, but should instead update the maistra docs to specify how to deploy the kiali operator, possibly just linking to the kiali docs.

jmazzitelli:
This PR needs the latest kiali operator (not the 0.21 release). Also needs this (which as of my typing this hasn't been merged yet) : https://github.com/kiali/kiali/pull/1171

UPDATE: that PR has been merged into Kiali master. The latest image should now work with this PR.

jmazzitelli:
My testing shows that this is now working. Here's the testing I have been doing. Note that in this PR I include some example dev scripts in `deploy/examples/kiali` to make this easier.

1. Start with a clean cluster - no Maistra or anything. I started with an OpenShift 3.11 cluster.
2. `cd deploy/examples/kiali` (make that your cwd)
3. Run `build-maistra.sh` which just runs `cd ../../.. && IMAGE=${USER}/istio-operator ./tmp/build/docker_build.sh`
4. Now deploy the Kiali and Maistra operators and the Maistra SMCP CR. NOTE! You must use a latest build of kiali op (v0.21 won't work). I build a local one and tell my script to use it. Or use the published "latest" image - that should work, too. This script does everything you need:
```
KIALI_VERSION_TO_USE=latest ./deploy-maistra.sh
```
5. See that you can access Kiali. Because I did not install a SMMR and MT is off, Kiali should be able to access everything. `oc get kiali kiali -n istio-system -o yaml` should show `deployment.accessible_namespaces` set to `**`. (so should the configmap - `oc get cm kiali -n istio-system -o yaml`)
6. Install bookinfo
7. Install SMMR: `oc apply -f ../maistra_v1_servicemeshmemberroll_cr.yaml -n istio-system`
8. Kiali operator will now only give Kiali access to the bookinfo namespace. `oc get kiali kiali -n istio-system -o yaml` should show `deployment.accessible_namespaces` to include `bookinfo` (the configmap - `oc get cm kiali -n istio-system -o yaml` - will show `[ "istio-system", "bookinfo" ]`)
9. Remove SMMR: `oc delete -f ../maistra_v1_servicemeshmemberroll_cr.yaml -n istio-system`
10: Kiali should be able to access everything again. `oc get kiali kiali -n istio-system -o yaml` should show `deployment.accessible_namespaces` set to `**`.  (so should the configmap - `oc get cm kiali -n istio-system -o yaml`)

===

Some other things I tested:

* Editing the SMCP to go from `multitenant: false` to `true`.. .you will see maistra operator change the Kiali CR's (`oc get kiali kiali -n istio-system -o yaml`) value of `deployment.accessible_namespaces`  which in turn causes the kiali operator to change the kiali configmap from this:

```
deployment:
  accessible_namespaces:
  - '**'
```

to this

```
api:
  namespaces:
    label_selector: kiali.io/member-of=istio-system
deployment:
  accessible_namespaces:
  - istio-system
```

* If you comment out from the SMCP the kiali credentials (i.e. don't set `kiali.dashboard.user/passphrase`), you will see maistra operator change the Kiali CR (`oc get kiali kiali -n istio-system -o yaml`) which in turn causes Kiali operator change the kiali configmap such that `auth.strategy` goes from `login` to `openshift` in both the CR and the configmap.

pkg/controller/servicemesh/controlplane/hooks.go
```diff
@@ -68,134 +53,9 @@ func (r *ControlPlaneReconciler) processNewObject(object *unstructured.Unstructu
...
-	}
 	return nil
 }
 
-var (
-	grafanaRegexp = regexp.MustCompile("(grafana:\\s*url:).*?\n")
-	jaegerRegexp  = regexp.MustCompile("(jaeger:\\s*url:).*?\n")
-)
-
-func (r *ControlPlaneReconciler) patchKialiConfig(object *unstructured.Unstructured) error {
```
rcernich: Don't we still need to do this, so Kiali has the correct route names?

jmazzitelli: Kiali now has auto-detection via routes built into the server. These settings in the Kiali CR are still there for people that have custom URLs that cannot be autodetected... but this auto-detection by looking at the routes is now done inside the Kiali Server so we don't need to do it here. We just leave the settings unset or empty and that tells Kiali to auto-detect. See:

https://github.com/kiali/kiali/pull/1022

rcernich: does it look for a route named `jaeger` or `tracing`, or does it just use the internal services?

jmazzitelli: By default (if you don't tell it what service it is), it will attempt to look for either jaeger-query or tracing.. see:

https://github.com/kiali/kiali/blob/master/status/discover.go#L18

Here is where it does the lookup:

https://github.com/kiali/kiali/blob/master/status/discover.go#L98-L110

rcernich: So that means it will default to tracing.  Currently, that route is in flux, and jaeger exposes the jaeger-query service through a route named `jaeger`.

jmazzitelli: We can tell @aljesusg if there is a problem we find in the kiali auto-detection. Or if we need to change that algorithm to look for different names.

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -400,9 +404,60 @@ func (r *ReconcileMemberList) Reconcile(request reconcile.Request) (reconcile.Re
...
+	// just get an array of strings consisting of the list of namespaces to be accessible to Kiali
+	var accessibleNamespaces []string
+	if len(requiredMembers) == 0 {
+		// we are not in multitenency mode - kiali can access the entire cluster
+		accessibleNamespaces = []string{"**"}
+	} else {
+		// we are in multitenency mode
+		accessibleNamespaces = make([]string, 0, len(requiredMembers))
+		for key := range requiredMembers {
+			accessibleNamespaces = append(accessibleNamespaces, key)
```
gbaufake: Would that work when the namespace is removed from requiredMembers?

jmazzitelli: Yes. This is happening in the function "reconcileKiali" - it is getting called during a reconciliation ... so when you change the SMMR by adding or removing from the spec.members list, this gets called with the new list "requiredMembers".

I tested this part and it seems to work pretty well.

jmazzitelli: Hmm... one thing I did not test is if you remove ALL the namespaces from spec.members (when spec.members is empty, I think Maistra will assume only the control plane namespace is in the mesh). However, this should still work since `accessibleNamespaces` will be of type `[]string]` but will be empty - so the line down below that calls `SetNestedStringSlice` should be setting the spec.deployment.accessible_namespaces yaml correctly (i.e. it should be set to `[ ]` - an empty list).

pkg/controller/servicemesh/controlplane/hooks.go
jmazzitelli: From Gary Brown of the Jaeger team:

> the 'tracing' service is intended as a tracing
> provider independent UI endpoint, so users did not have to be aware of the
> implementation being used. As service mesh will only support Jaeger, I
> think the 'tracing' service is not relevant.

Then perhaps we should remove or change the order here (you see here the tracing route is looked at first when Kiali is auto-detecting the jaeger endpoint).

https://github.com/kiali/kiali/blob/master/status/discover.go#L18

If tracing is not relevant when a mesh only supports Jaeger, then we should be looking for the jaeger route first, and only then look for tracing if we don't find it.

@aljesusg

aljesusg: Hi, Kiali looks in a list of services that we set in the code if the user not set the tracing service in the configmap.

```go
var tracingLookupRoutes = [...]string{"tracing", "jaeger-query"}
```

If there is a service in the Istio namespace installation with this name kiali will assume it as the tracing service, after that kiali looks for the route if the environment is openshift.

So we can change the order.

mwringe:
If you don't have Grafana installed (its disabled in the servicemeshcontrolplane) then we are still adding in Grafana information to the kiali CR. This causes issues as Kiali thinks Grafana should be available as its configured in the kiali CR when its not installed. This causes error messages to be shown to users.

I believe it acts the same for Jaeger.

If Jaeger or Grafana are not installed in the servicemeshcontrolplane resource then we shouldn't be configuring the kiali resource with this information.

rcernich:
Hey @mwringe, this sounds like a general design consideration.  Which components are required to be installed for Kiali, which are optional, and what functionality will be missing if those components aren't installed?  Keep in mind, we want to have an opinionated install, so what's your opinion?

jmazzitelli:
I'll also bring up the fact that there is no way to tell kiali that grafana is disabled (that I can see).
We have an enabled flag for tracing: https://github.com/kiali/kiali/blob/v0.21.0/config/config.go#L156 but we have nothing like this for grafana: https://github.com/kiali/kiali/blob/v0.21.0/config/config.go#L146-L150

So we can't tell Kiali that grafana isn't available even if we wanted to.

mwringe:
@jmazzitelli Just don't add the external_services section for grafana and jaeger if they are not available (which the maistra operator I guess should if they are installed or not). If the configuration is not there and we can't detect that the services are running, then they wont be shown in the UI.

@rcernich Kiali only really needs access to p8s and the k8s master. Things like grafana and jaeger are optional. If they are not installed (or configured) then they will not be shown in the Kiali UI.

jmazzitelli:
> ust don't add the external_services section for grafana and jaeger if they are not available (which the maistra operator I guess should if they are installed or not). If the configuration is not there and we can't detect that the services are running, then they wont be shown in the UI.

@mwringe That doesn't help. As a quick test, I just removed the "grafana" and "tracing" sections entirely from the configmap, I then restarted the kiali pod... I still get errors in the Kiali UI:

```
Could not fetch Grafana info. Turning off links to Grafana., Error: [ grafana URL is not set in Kiali configuration ]
6/26/201910:52:06 AM
Could not fetch Jaeger info. Turning off Jaeger integration., Error: [ Get https://jaeger-query-istio-system.192.168.1.19.nip.io/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers) ] 
6/26/201910:52:07 AM
```

This is basically the same behavior if I simply set the URL to "".

What we need is to add a new "external_services.grafana.enabled" flag - we already have that for tracing. Then we need to have the istio-operator set those two to false in the Kiali CR when they are not installed:

```
external_services:
  grafana:
    enabled: false
  tracing:
    enabled: false
```

This requires a KIALI JIRA - we need to implement the Grafana enabled flag in the Kiali Server. As it is right now, I don't think there is any way we can avoid having an error message show in Kiali UI because AFAIK there is no way to turn off the auto-detection of Grafana.

jmazzitelli:
BTW: I think this is a minor issue because, as I understand, Grafana and Jaeger come with Maistra by default. User has to actually flip the bits in the istio CR to turn them off. In other words, I don't think this issue blocks this PR from being merged.

jmazzitelli:
Created this: https://issues.jboss.org/browse/KIALI-3066 "need external_services.grafana.enabled flag"

deploy/examples/kiali/kiali-operator.yaml
```diff
@@ -0,0 +1,398 @@
...
+  scope: Namespaced
+  version: v1alpha1
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: kialis.kiali.io
+  labels:
+    app: kiali-operator
+    version: v0.21.0
```
gbaufake: Is this parameter going to be passed through maistra CR?

deploy/examples/kiali/kiali-operator.yaml
```diff
@@ -0,0 +1,398 @@
...
+        - /tmp/ansible-operator/runner
+        - stdout
+        image: quay.io/kiali/kiali-operator:v0.21.0
+        imagePullPolicy: IfNotPresent
+        volumeMounts:
+        - mountPath: /tmp/ansible-operator/runner
+          name: runner
+          readOnly: true
+      - name: operator
+        image: quay.io/kiali/kiali-operator:v0.21.0
```
gbaufake: Is this going to be passed via maistra CR?

deploy/examples/kiali/kiali-operator.yaml
```diff
@@ -0,0 +1,398 @@
...
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups: ["rbac.authorization.k8s.io"]
+  resources:
+  - clusterrolebindings
+  - clusterroles
```
gbaufake: @jmazzitelli 
Is this going to be always created on maistra case?

jmazzitelli: Yes, which shouldn't be a problem because Maistra's own operator requires these same permissions:

https://github.com/Maistra/istio-operator/blob/maistra-0.12/deploy/maistra-operator.yaml#L128-L129

rcernich: Actually, it's less of a problem because all istio-operator needs to be able to do is create the kiali CR now.

deploy/examples/kiali/kiali-operator.yaml
rcernich: This is the operator install, which should be moved out of istio-operator.  istio-operator should document how to install the kiali operator.  Eventually, we should be using OLM for the installs.

pkg/controller/servicemesh/controlplane/hooks.go
rcernich: I think we should keep this code, but adapt it to modify the kiali CR.  This is where you could disable missing dependencies (e.g. jaeger, grafana).

jmazzitelli: we can't disable missing grafana today - kiali needs this: see https://issues.jboss.org/browse/KIALI-3066

deploy/examples/kiali/kiali-operator.yaml
jmazzitelli: i will remove all of these deploy/example files - they were just here to help people test/review the PR. But I will remove if no one is using them.

pkg/controller/servicemesh/controlplane/hooks.go
jmazzitelli: What I am going to do is - I will try to put this code back - but obviously change it so it modifies the Kiali CR (this code here is modifying the configmap which we should not be doing anymore since the kiali operator will deal with configmap updates).

I will have this code set the "external_services.tracing.enabled" flag to false if Jaeger is not available. That is a valid setting.

I will have this code set the "external_services.grafana.enabled" flag to false if Grafana is not available. This is not used today - its just a flag that will be ignored, but once https://issues.jboss.org/browse/KIALI-3066 is implemented, it will take effect. Kiali won't burp if we give it an unknown setting, it will just blindly ignore it.

rcernich: Sounds good.  I think it's probably safe to assume that if the routes are missing, the functionality is missing too.

jmazzitelli:
The latest force push includes a rebase on the maistra-0.12 branch as well as code changes that were discussed - i.e. should be setting the Kiali CR's url and enabled flags for tracing and grafana. This is not ready for merge yet. I need to test.

deploy/maistra-operator.yaml
```diff
@@ -198,6 +198,18 @@ rules:
...
   verbs:
   - '*'
+- apiGroups:
+  - kiali.io
+  resources:
+  - kialis
+  verbs:
+  - '*'
+- apiGroups:
+  - monitoring.kiali.io
```
jmazzitelli: I'm not sure if istio-operator even needs this monitoring.kiali.io apiGroup perms.. I'll test that soon. I think this can be removed.

deploy/servicemesh-operator.yaml
```diff
@@ -198,6 +198,18 @@ rules:
...
   verbs:
   - '*'
+- apiGroups:
+  - kiali.io
+  resources:
+  - kialis
+  verbs:
+  - '*'
+- apiGroups:
+  - monitoring.kiali.io
```
jmazzitelli: I'm not sure if istio-operator even needs this monitoring.kiali.io apiGroup perms.. I'll test that soon. I think this can be removed.

jmazzitelli:
OK, this is again ready for review/approval.

I removed the example scripts.

I also put back the auto-detection of jaeger and grafana URLs. Note: auto-detection is not performed if the user explicitly tells kiali what URLs to use - usually done when the user explicitly sets the helm chart values of `.Values.dashboard.grafanaURL`  and/or `.Values.dashboard.jaegerURL`.  Note also: if the auto-detection fails, maistra operator will tell kiali operator to disable the pieces that failed (e.g. if jaeger route cannot be found, kiali CR will have the external_services.jaeger.enable flag set to false - same for grafana).

cc @rcernich @gbaufake

jmazzitelli:
One other thing you will notice in here - there were a whole bunch of superfluous Kiali rules in the operator Role. Now that we are using kiali operator, the istio operator doesn't need those kiali permissions. However, I went through each one just in case. I did find one apiGroup in that kiali section of permissions that the istio operator actually needs and didn't have it elsewhere in the Role - the batch apiGroup (istio operator does something with jobs so it needs that apiGroup permissions).

All the rest of the permissions/apiGroups listed under that kiali section have been removed - they are mostly duplicates (the only thing not a duplicate was the monitoring.kiali.io permissions but it doesn't need that - the kiali operator will handle the monitoring dashboards).

jmazzitelli:
BTW: I want to mention this here.

Right now, if there is no SMMR (no multitenancy), Kiali is installed in the same mode as it is in upstream Istio (which is determined by the setting accessible_namespaces=**). This behavior is "Kiali can access any namespace, including namespaces created in the future". However, this requires Kiali be given cluster roles, which in turn means the kiali operator needs to be given 'create-cluster-role' permissions.

So, to work with Maistra, the Kiali Operator must be installed with "create-cluster-role(binding" permissions.

I just wanted to point that out. We could change the defaults such that Kiali in non-MT environments is given access to all namespaces, but not namespaces that are created in the future - this would not require a cluster role nor require the operator to be given create-cluster-role(binding) permissions. To do this, we simply change the code to remove accessible_namespaces so Kiali picks up the default - which is every namespace minus some internal ones like all the kube-* ones. But if we do this, Kiali won't be able to see new namespaces created in the future.

rcernich:
> I just wanted to point that out. We could change the defaults such that Kiali in non-MT environments is given access to all namespaces, but not namespaces that are created in the future - this would not require a cluster role nor require the operator to be given create-cluster-role(binding) permissions. To do this, we simply change the code to remove accessible_namespaces so Kiali picks up the default - which is every namespace minus some internal ones like all the kube-* ones. But if we do this, Kiali won't be able to see new namespaces created in the future.

For now, we should probably use the cluster-wide version, to simplify things.  For service mesh, we key off the `multitenent` flag and switch things up (cluster role binding if false, namespace role binding if true).  That said, there are still a few components that have cluster perms, although we're working to eliminate those.

@knrc, FYI.

pkg/controller/servicemesh/controlplane/hooks.go
```diff
@@ -69,146 +58,111 @@ func (r *ControlPlaneReconciler) processNewObject(object *unstructured.Unstructu
...
+		// jaeger is the name of the Jaeger CR
+		err = r.Client.Get(context.TODO(), client.ObjectKey{Name: "jaeger", Namespace: object.GetNamespace()}, jaegerRoute)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				r.Log.Error(err, "error retrieving jaeger route - will disable it in Kiali")
+				// we aren't going to return here - Jaeger is optional for Kiali; Kiali can still run without it
+			}
+			jaegerEnabled = false
+		} else {
+			jaegerURL, _, _ = unstructured.NestedString(jaegerRoute.UnstructuredContent(), "spec", "host")
```
rcernich, line 92: Can we use the route you retrieved above, to save a call to the api server?  Same applies to grafana code below.

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -217,6 +217,10 @@ func (r *ReconcileMemberList) Reconcile(request reconcile.Request) (reconcile.Re
 			}
 			break
 		}
+
+		// tell Kiali that MT mode is disabled. This allows Kiali to see the full cluster.
+		r.reconcileKiali(instance.Namespace, nil, reqLogger)
```
rcernich: I think this is an incorrect assumption.  Just because the smmr is being deleted does not mean the control plane is becoming cluster wide.  I think two things are true here:

1. the user deleted the smmr
2. the user deleted the smcp (we add an owner ref to the smcp, so cleanup of smmr occurs when the control plane is deleted).

Furthermore, there's a fringe case where the user may convert a multitenant install into a cluster wide install (although, this is probably something we should prevent in the validator).

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -400,9 +404,60 @@ func (r *ReconcileMemberList) Reconcile(request reconcile.Request) (reconcile.Re
 		}
 	}
 
+	// tell Kiali about all the namespaces in the mesh
+	r.reconcileKiali(instance.Namespace, requiredMembers, reqLogger)
```
rcernich: I think you should be using `Status.ConfiguredMembers` here, as those are the namespaces that have been configured for use with the mesh.  It's possible that the required list contains namespaces that don't exist (although, the validator should catch those, but not if the namespace is deleted after creation).

pkg/controller/servicemesh/controlplane/hooks.go
jmazzitelli: This may be my misunderstanding of this unstructured API - where is this extra call to the API server?

I thought `jaegerRoute.UnstructuredContent()` merely obtains the json that was retrieved already from the r.Client.Get() call from above. Are you saying `unstructured.NestedString` or `jaegerRoute.UnstructuredContent()` makes another call to the API server?

I see only two calls to the API server - one to get jaeger route, one to get grafana route.

pkg/controller/servicemesh/memberroll/controller.go
jmazzitelli: So what is the semantics here wrt how Kiali should behave/be configured? Do we just tell Kiali it cannot see any namespaces (other than the control plane namespace)? That would be a simple fix.

>  there's a fringe case where the user may convert a multitenant install into a cluster wide install

OK, that is what this code (line 222) is thinking this means. So you are saying this is NOT converting a multi-tenant install to a cluster-wide install and that we should prevent that from ever happening. Then that means I'll need an answer to the above question - what is the semantics as it is related to Kiali? I think  you are saying if the SMMR is deleted (thus no mesh namespaces are provided) then Kiali should not have any access to any namespace outside the control plane namespace.

pkg/controller/servicemesh/controlplane/hooks.go
rcernich: Never mind.  I don't know what I was seeing before.  This is fine as is.

pkg/controller/servicemesh/memberroll/controller.go
rcernich: > So what is the semantics here wrt how Kiali should behave/be configured? Do we just tell Kiali it cannot see any namespaces (other than the control plane namespace)? That would be a simple fix.

Yes.

> OK, that is what this code (line 222) is thinking this means. So you are saying this is NOT converting a multi-tenant install to a cluster-wide install and that we should prevent that from ever happening. Then that means I'll need an answer to the above question - what is the semantics as it is related to Kiali? I think you are saying if the SMMR is deleted (thus no mesh namespaces are provided) then Kiali should not have any access to any namespace outside the control plane namespace.

Adding some color here: it is possible that the user can convert (although we should prevent in the future) a multitenant install into a cluster wide install.  In that case, the smmr reconciler will remove any configured members from the mesh (this is really just cleaning up the work done when it was multitenant, e.g. removing role bindings, etc.).  For kiali, you can use the `isMultitenant` flag and set the namespaces to `**` if it's false, or you could do nothing (assumption being the kiali resource was already updated to be cluster wide by the smcp reconciliation process).

pkg/controller/servicemesh/memberroll/controller.go
jmazzitelli: Instead of "requiredMembers" getting passed in, this now gets instance.Status.ConfiguredMembers (type changes from map to array - so the code changes to reflect this type change). See: https://github.com/Maistra/istio-operator/pull/87/files#diff-3029a0e7a4af7d79dcfe22782cf7dbbaR408

pkg/controller/servicemesh/memberroll/controller.go
jmazzitelli: Rather than passing in an nil value, code now passes in the control plan namespace. This makes it clear what the semantics are here - at this point in the reconciliation process, we are telling kiali to only have access to the control plane namespace and no others. See: https://github.com/Maistra/istio-operator/pull/87/files#diff-3029a0e7a4af7d79dcfe22782cf7dbbaR222

jmazzitelli:
@gbaufake we made some changes here that changes behavior. If you delete the SMMR, it no longer moves kiali into "non-multitenancy mode" (aka cluster mode, or accessibile_namespaces =**). Instead, what should happen is accessible_namespaces should be set to just the control plane namespace ("istio-system" for example)

gbaufake:
@jmazzitelli Ok. I will retest.

Just to document the corner case that I faced:

I enabled Jaeger Operator on the CR, but I didn't install the Jaeger operator and then I added two meshes (1 on control-plane a, 2 on control plane b). The namespaces were not reconciled because of an error on Kiali.

Some logs:

`{"level":"error","ts":1561995668.398791,"logger":"kubebuilder.controller","caller":"controller/controller.go:209","msg":"Reconciler error","Controller":"servicemeshcontrolplane-controller","Request":"istio-system/basic-install-istio-system","error":"[no matches for kind \"Jaeger\" in version \"jaegertracing.io/v1\",`


`{"level":"error","ts":1561995547.20894,"logger":"controller_servicemeshmemberroll","caller":"memberroll/controller.go:285","msg":"skipping reconciliation because mesh is not in a known good state","Request.Namespace":"istio-system2","Request.Name":"default","stacktrace":"github.com/maistra/istio-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/vendor/github.com/go-logr/zapr/zapr.go:128\ngithub.com/maistra/istio-operator/pkg/controller/servicemesh/memberroll.(*ReconcileMemberList).Reconcile\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/pkg/controller/servicemesh/memberroll/controller.go:285\ngithub.com/maistra/istio-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:207\ngithub.com/maistra/istio-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:157\ngithub.com/maistra/istio-operator/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133\ngithub.com/maistra/istio-operator/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134\ngithub.com/maistra/istio-operator/vendor/k8s.io/apimachinery/pkg/util/wait.Until\n\t/Users/`.


Installing Jaeger Operator afterwards, didn't change nothing:

`{"level":"error","ts":1562014376.8188071,"logger":"controller_servicemeshcontrolplane","caller":"controlplane/pruner.go:95","msg":"Error retrieving resources to prune","type":"jaegertracing.io/v1, Kind=jaeger","error":"no matches for kind \"jaeger\" in version \"jaegertracing.io/v1\"","stacktrace":"github.com/maistra/istio-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/vendor/github.com/go-logr/zapr/zapr.go:128\ngithub.com/maistra/istio-operator/pkg/controller/servicemesh/controlplane.(*ControlPlaneReconciler).pruneResources\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/pkg/controller/servicemesh/controlplane/pruner.go:95\ngithub.com/maistra/istio-operator/pkg/controller/servicemesh/controlplane.(*ControlPlaneReconciler).prune\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/pkg/controller/servicemesh/controlplane/pruner.go:75\ngithub.com/maistra/istio-operator/pkg/controller/servicemesh/controlplane.(*ControlPlaneReconciler).Reconcile\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/pkg/controller/servicemesh/controlplane/reconciler.go:131\ngithub.com/maistra/istio-operator/pkg/controller/servicemesh/controlplane.(*ReconcileControlPlane).Reconcile\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/pkg/controller/servicemesh/controlplane/controller.go:179\ngithub.com/maistra/istio-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:207\ngithub.com/maistra/istio-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:157\ngithub.com/maistra/istio-operator/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133\ngithub.com/maistra/istio-operator/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134\ngithub.com/maistra/istio-operator/vendor/k8s.io/apimachinery/pkg/util/wait.Until\n\t/Users/gbaufake/golang/src/github.com/maistra/istio-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88"}`
so I think the order of the operators matters for this case.

Deleting Maistra Operator Pod and restarting it, didn't change a thing, so I guess it will need a complete wipeout to everything on proper state.
[istio-operator-54df8b7598-dwstp-istio-operator.log](https://github.com/Maistra/istio-operator/files/3347309/istio-operator-54df8b7598-dwstp-istio-operator.log)

.

jmazzitelli:
> I enabled Jaeger Operator on the CR, but I didn't install the Jaeger operator  ...
> The namespaces were not reconciled because of an error on Kiali.

As I mentioned earlier, I don't see how this is an error with Kiali. This has nothing to do with Kiali or even this PR. It has to do with the helm charts that Maistra is running and the fact that one helm chart (the Jaeger helm chart) wants to create a Jaeger CR but it cannot because there is no Jaeger Kind  (because the Jaeger CRD hasn't been installed).

At that point, the helm installation is not 100% successful and istio most likely will be in a bad state (as the operator says in its logs - "skipping reconciliation because mesh is not in a known good state".

rcernich:
@gbaufake, that is a different issue, which has since been fixed #104.

rcernich:
@jmazzitelli, great work!  Can I ask two more things (sorry):

1.  Please add instructions for installing the Kiali operator (feel free to link to existing Kiali operator docs, but try to ensure it's constrained to a particular release version), e.g. like the [Jaeger Operator](https://github.com/Maistra/istio-operator#installing-the-jaeger-operator) (don't forget to update the uninstall section too).
2.  Please add the kiali resource to the prune code [here](https://github.com/Maistra/istio-operator/blob/maistra-0.12/pkg/controller/servicemesh/controlplane/pruner.go#L59).  This will ensure that the istio operator properly cleans up any kiali install (uninstall will work fine, because of the owner reference, but if a user sets `kiali.enabled: false`, kiali won't be removed if the type isn't in the types list for the pruner).

Thanks again.

jmazzitelli:
> 2\. Please add the kiali resource to the prune code [here](https://github.com/Maistra/istio-operator/blob/maistra-0.12/pkg/controller/servicemesh/controlplane/pruner.go#L59).  This will ensure that the istio operator properly cleans up any kiali install (uninstall will work fine, because of the owner reference, but if a user sets `kiali.enabled: false`, kiali won't be removed if the type isn't in the types list for the pruner).

@rcernich That is already part of this PR - see https://github.com/Maistra/istio-operator/pull/87/files#diff-09a8490f8c6b56a38fd78ed7ed8a7fb0R22

rcernich (approved):

PR #88, opened 2019-06-14
MAISTRA-506: Add a namespace label for deployment specs

jwendell:


brian-avery (approved):

tmp/build/patch-charts.sh
```diff
@@ -123,6 +123,10 @@ function patchTemplates() {
     xargs sed -i -e 's/^\(.*\)release:\(.*\)$/\1maistra-version: '${MAISTRA_VERSION}'\
 \1release:\2/'
 
+# MAISTRA-506 add a namespace label for deployment specs
+find ${HELM_DIR} -name "*.yaml" -o -name "*.yaml.tpl" | xargs grep -Hl 'kind: Deployment' |\
+  xargs sed -i -e '/^spec:/,$ { /template:$/,$ { /metadata:$/,$ { /labels:$/,$ s/^\(.*\)release:\(.*\)$/\1namespace: {{ .Release.Namespace }}\n\1release:\2/ } } }'
+
```
jwendell, line 129: @knrc what about `control-plane` for the label name?

knrc: @jwendell +1 although it might be better to prefix it with maistra so we don't accidentally clash in future, i.e. maistra-control-plane

knrc (changes requested):

tmp/build/patch-charts.sh
jwendell: @knrc done

knrc (approved):
lgtm, thanks very much

PR #95, opened 2019-06-17
[MAISTRA-462] Put the 'member-of' label in the control plane namespace

jwendell:
So that Kiali knows about it.

### DIFF
--- a/pkg/controller/servicemesh/controlplane/reconciler.go
+++ b/pkg/controller/servicemesh/controlplane/reconciler.go
@@ -68,16 +68,24 @@ func (r *ControlPlaneReconciler) Reconcile() (reconcile.Result, error) {
 			namespace.Labels = map[string]string{}
 		}
 
-		// Label the control plane namespace so that Kiali knows about it
-		r.Log.Info(fmt.Sprintf("Adding %s label to namespace %s", common.MemberOfKey, r.Instance.Namespace))
-		namespace.Labels[common.MemberOfKey] = r.Instance.Namespace
+		needsUpdate := false
+
+		if label, ok := namespace.Labels[common.MemberOfKey]; !ok || label != r.Instance.Namespace {
+			// Label the control plane namespace so that Kiali knows about it
+			r.Log.Info(fmt.Sprintf("Adding %s=%s label to namespace %s", common.MemberOfKey, r.Instance.Namespace, r.Instance.Namespace))
+			namespace.Labels[common.MemberOfKey] = r.Instance.Namespace
+			needsUpdate = true
+		}
 
 		if label, ok := namespace.Labels[common.IgnoreNamespace]; !ok || label != "ignore" {
 			r.Log.Info(fmt.Sprintf("Adding %s=ignore label to namespace %s", common.IgnoreNamespace, r.Instance.Namespace))
 			namespace.Labels[common.IgnoreNamespace] = "ignore"
+			needsUpdate = true
 		}
 
-		err = r.Client.Update(context.TODO(), namespace)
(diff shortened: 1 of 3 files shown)


pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -65,11 +66,16 @@ func (r *ControlPlaneReconciler) Reconcile() (reconcile.Result, error) {
 		if namespace.Labels == nil {
 			namespace.Labels = map[string]string{}
 		}
+
+		// Label the control plane namespace so that Kiali knows about it
+		namespace.Labels[common.MemberOfKey] = r.Instance.Namespace
+
```
rcernich, line 72: We should probably remove these labels during uninstall/delete.

jwendell: Done

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -65,11 +67,17 @@ func (r *ControlPlaneReconciler) Reconcile() (reconcile.Result, error) {
...
-			r.Log.Info("Adding maistra.io/ignore-namespace=ignore label to Request.Namespace")
-			namespace.Labels["maistra.io/ignore-namespace"] = "ignore"
-			err = r.Client.Update(context.TODO(), namespace)
+
+		// Label the control plane namespace so that Kiali knows about it
+		r.Log.Info(fmt.Sprintf("Adding %s label to namespace %s", common.MemberOfKey, r.Instance.Namespace))
+		namespace.Labels[common.MemberOfKey] = r.Instance.Namespace
+
+		if label, ok := namespace.Labels[common.IgnoreNamespace]; !ok || label != "ignore" {
+			r.Log.Info(fmt.Sprintf("Adding %s=ignore label to namespace %s", common.IgnoreNamespace, r.Instance.Namespace))
```
rcernich, line 81: this test prevents updating the namespace if the label already exists.  this code will be executed for every update to the CR, so we save an api call if the labels already exist.

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -65,11 +67,17 @@ func (r *ControlPlaneReconciler) Reconcile() (reconcile.Result, error) {
...
 			namespace.Labels = map[string]string{}
 		}
-		if label, ok := namespace.Labels["maistra.io/ignore-namespace"]; !ok || label != "ignore" {
-			r.Log.Info("Adding maistra.io/ignore-namespace=ignore label to Request.Namespace")
-			namespace.Labels["maistra.io/ignore-namespace"] = "ignore"
-			err = r.Client.Update(context.TODO(), namespace)
+
+		// Label the control plane namespace so that Kiali knows about it
+		r.Log.Info(fmt.Sprintf("Adding %s label to namespace %s", common.MemberOfKey, r.Instance.Namespace))
+		namespace.Labels[common.MemberOfKey] = r.Instance.Namespace
```
rcernich: This should be part of the test below too.  We don't want to update the namespace if we don't have to.

pkg/controller/servicemesh/controlplane/reconciler.go
jwendell: Thanks for the clarification

pkg/controller/servicemesh/controlplane/reconciler.go
jwendell: Done, thanks.

jwendell:
Closing as absolete

PR #100, opened 2019-06-25
MAISTRA-542 configure NetworkPolicy and NetNamespace resources for mesh members

rcernich:


rcernich:
This should be good to go.  Verified network connectivity using bookinfo.

Here are the *default* `NetworkPolicy` resources I was using for testing:

```yaml
- apiVersion: networking.k8s.io/v1
  kind: NetworkPolicy
  metadata:
    name: allow-from-same-namespace
  spec:
    ingress:
    - from:
      - podSelector: {}
    podSelector: null

- apiVersion: networking.k8s.io/v1
  kind: NetworkPolicy
  metadata:
    name: allow-from-default-namespace
  spec:
    ingress:
    - from:
      - namespaceSelector:
          matchLabels:
            name: default
    podSelector: null

- apiVersion: networking.k8s.io/v1
  kind: NetworkPolicy
  metadata:
    name: deny-by-default
  spec:
    ingress: []
    podSelector: null
```

When bookinfo was not a member, pinging product page through the ingress gateway resulted in 503.  After isolating bookinfo (either `oc adm pod-network isolation-projects bookinfo` or `oc delete networkpolicy/istio-mesh -n bookinfo`), the result was a 504 (server timed out).  Also, the proxies fail to start if the networking is not setup correctly, as they can't get their initial config from pilot.

rcernich:
Added a check to the smmr controller to only process members if the mesh is configured for multitenancy.  If the mesh is not configured for multitenancy and a member roll is created, the smmr controller will effectively remove the members from the mesh.  This could be a problem if ovs_multitenant is being used, as it will isolate the member namespaces from the mesh's network.

helm/istio/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if .Values.global.multitenant }}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
```
brian-avery, line 3: is the goal here to make it so that only pods in the mesh can talk to other pods in the mesh? Would this break any migration strategies? People can modify this, but is it too tight by default?

brian-avery (approved):

helm/istio/templates/networkpolicy.yaml
rcernich: The goal is to ensure all members of the mesh can talk to all other members and to the control plane.  This is effectively white listing communication between them.  It is not restricting any other communication (i.e. it doesn't black list anything).  Users are free to constrain their projects however they want.  That said, if they wanted to constrain what services were available to the mesh, they should create a separate project for those and expose them to mesh member, e.g. by giving specific access to the private pod from a member pod/namespace.  (This is similar to policy optimization, which you can read about [here](https://docs.openshift.com/container-platform/3.9/admin_guide/managing_networking.html#admin-guide-networking-using-networkpolicy-efficiently).

PR #101, opened 2019-06-26
MAISTRA-536 Add istio-cni to the operator

luksa:


rcernich:
Do we need to add the new chart to the list of requirements?

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -624,6 +649,68 @@ func (r *ReconcileMemberList) reconcileRoleBindings(namespace string, mesh *v1.S
 	return utilerrors.NewAggregate(allErrors)
 }
 
+func (r *ReconcileMemberList) addNetworkAttachmentDefinition(namespace, meshNamespace string, reqLogger logr.Logger) error {
+	name := fmt.Sprintf("%s-%s", meshNamespace, "istio-cni") // must match name of .conf file in multus.d
+
+	netAttachDef := &unstructured.Unstructured{}
+	netAttachDef.SetGroupVersionKind(schema.GroupVersionKind{
```
rcernich, line 813: Should we add labels, similar to the other created resources (owner labels get copied over from resources in mesh namespace on role bindings), which allows us to delete the ones we actually created, e.g. https://github.com/Maistra/istio-operator/blob/maistra-0.12/pkg/controller/servicemesh/memberroll/controller.go#L461

rcernich (approved):

PR #102, opened 2019-06-27
MAISTRA-548 Drop CAP_MKNOD from sidecar container

luksa:
The anyuid SCC requires the MKNOD kernel capability to be dropped. If
the container's spec doesn't drop it, the SCC admission controller tries
to add it, but isn't allowed to, because it tries to do that during the
validation phase (as the sidecar container wasn't injected prior to the
SCC controller running during the mutation phase). Because it can't
mutate the pod, the SCC admission controller rejects it. By dropping the
capability in the sidecar template, the mutation is not required and
thus the pod is marked as valid by the SCC admission controller.


rcernich (approved):
lgtm

PR #104, opened 2019-06-28
MAISTRA-554 Don't error when pruning object that can't even exist

luksa:
For example, if there's no Jaeger CRD in the system, deleting a Jaeger
resource will fail with NoKindMatchError. The pruner should not treat
this as an error, because the object can't exist if the CRD isn't there.

### DIFF
--- a/pkg/controller/servicemesh/controlplane/pruner.go
+++ b/pkg/controller/servicemesh/controlplane/pruner.go
@@ -2,6 +2,7 @@ package controlplane
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


jwendell (approved):

aljesusg:
This should be fixed too https://github.com/Maistra/istio-operator/blob/maistra-0.12/deploy/maistra-operator.yaml#L198 

jaegers => jaeger

luksa:
@aljesusg No, `jaegers` is correct (must be plural).

aljesusg:
ok Thanks @luksa

PR #105, opened 2019-06-28
[MAISTRA-544] Fix the link between Grafana <-> Prometheus OAuth

jwendell:


pkg/controller/servicemesh/controlplane/grafana.go
```diff
@@ -0,0 +1,128 @@
...
+	dsYaml, found, err := unstructured.NestedString(object.UnstructuredContent(), "data", "datasources.yaml")
+	if err != nil {
+		r.Log.Error(err, "could not parse istio-grafana ConfigMap")
+		return err
+	} else if !found {
+		r.Log.Error(nil, "failed to patch istio-grafana ConfigMap")
+		return nil
+	}
+
+	d := &GrafanaDatasources{}
```
rcernich: I think you should be operating against the yaml directly.  While it is easier to access the fields using a typed object, anything that's not defined in the type will be lost, which means round-tripping will fail.  This is why we don't use a typed version of values.yaml in the operator itself.

jwendell: @rcernich Done

pkg/controller/servicemesh/controlplane/grafana.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (r *ControlPlaneReconciler) patchGrafanaConfig(object *unstructured.Unstructured) error {
+	r.Log.Info("patching Grafana-Prometheus link", object.GetKind(), object.GetName())
+
+	dsYaml, found, err := unstructured.NestedString(object.UnstructuredContent(), "data", "datasources.yaml")
+	if err != nil {
+		r.Log.Error(err, "could not parse istio-grafana ConfigMap")
+		return err
+	} else if !found {
+		r.Log.Error(nil, "failed to patch istio-grafana ConfigMap")
```
rcernich: I think we should log this as `Info`, if not debug.  If no data sources are configured, then we don't need to update them.  I also don't think it "failed," as there's nothing to do.  Maybe something like, "skipping configuration of prometheus basic auth: no prometheus datasource defined."

pkg/controller/servicemesh/controlplane/grafana.go
```diff
@@ -0,0 +1,95 @@
...
+	// Retrieve the raw password created when processing Prometheus charts, then get rid of it
+	htSecret := &corev1.Secret{}
+	err = r.Client.Get(context.TODO(), client.ObjectKey{Namespace: object.GetNamespace(), Name: "prometheus-htpasswd"}, htSecret)
+	if err != nil {
+		r.Log.Error(err, "error retrieving prometheus-htpasswd Secret")
+		return err
+	}
+	rawPassword := string(htSecret.Data["rawPassword"])
+	delete(htSecret.Data, "rawPassword")
+	err = r.Client.Update(context.TODO(), htSecret)
```
rcernich: Is this necessary?  It might be best if it's simply left alone.

pkg/controller/servicemesh/controlplane/grafana.go
```diff
@@ -0,0 +1,95 @@
...
+	"fmt"
+	"regexp"
+
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func (r *ControlPlaneReconciler) patchPrometheusHtpasswd(object *unstructured.Unstructured) error {
```
rcernich, line 17: Have you tested this with an update?  I assume it would generate a new password.  In that case, do grafana and prometheus pickup the new password?  You can test by verifying grafana can continue to access prometheus after the update, then again after bouncing prometheus (in case the first test succeeded because prometheus didn't pull in the new htpasswd file).

tmp/build/patch-prometheus.sh
```diff
@@ -23,12 +23,15 @@ function prometheus_patch_deployment() {
...
             name: secret-prometheus-tls\
+          - mountPath: /etc/proxy/htpasswd\
+            name: secret-prometheus-htpasswd\
           args:\
           - -provider=openshift\
           - -https-address=:3001\
           - -http-address=\
           - -email-domain=*\
           - -upstream=http://localhost:9090\
+          - -htpasswd-file=/etc/proxy/htpasswd/auth\
```
rcernich, line 34: Should we also add `-display-htpasswd-form=false`?  If I understand correctly, this would prevent external users from being able to use the generated password to access prometheus.

pkg/controller/servicemesh/controlplane/grafana.go
jwendell: Done

tmp/build/patch-prometheus.sh
jwendell: Done

pkg/controller/servicemesh/controlplane/grafana.go
jwendell: Done

pkg/controller/servicemesh/controlplane/grafana.go
jwendell: Just tested in an update, indeed a new password is created. Grafana->Prometheus link works as long as none of the pods are killed (which means they're using the old password) *or* if both pods are killed (they both read the new password).

jwendell: Done. Now the password is only created at the first time, thus updates work fine.

rcernich (approved):

PR #106, opened 2019-06-28
MAISTRA-553 add validating webhook for SMCP and SMMR resources

rcernich:

Validates the following:

SMCP:
* only allows a single cluster-wide instance.  all other requests are denied, including multitenant installs
* only allows a single instance per namespace

SMMR:
* only allowed if name is `default`
* only allowed if mesh is multitenent (slight hole, as a mesh can be migrated from multitenant to cluster wide after SMMR is created)
* only allowed if user creating the resource has update access on pods for all listed `members`
* only allowed if all listed members are not already listed in another SMMR


deploy/maistra-operator.yaml
```diff
@@ -61,6 +61,18 @@ spec:
 
 ---
 
+# create role that can be used to grant users permission to create smcp and smmr resources
+apiVersion: rbac.authorization.k8s.io/v1
```
brian-avery, line 65: Is there an advantage to having permission to create maistra.io resources but not Istio? I.e. is this covered by the istio-admin role?

deploy/servicemesh-operator.yaml
```diff
@@ -61,6 +61,18 @@ spec:
 
 ---
 
+# create role that can be used to grant users permission to create smcp and smmr resources
```
brian-avery, line 64: Same question here. What's the advantage of being able to create maistra resources but not Istio? Why not just use istio-admin?

brian-avery (approved):

pkg/controller/servicemesh/validation/memberroll.go
```diff
@@ -0,0 +1,126 @@
...
+
+	err := v.decoder.Decode(req, smmr)
+	if err != nil {
+		logger.Error(err, "error decoding admission request")
+		return admission.ErrorResponse(http.StatusBadRequest, err)
+	}
+
+	// do we care about this object?
+	if !watchNamespace.watching(smmr.Namespace) {
+		logger.Info(fmt.Sprintf("operator is not watching namespace '%s'", smmr.Namespace))
```
luksa, line 42: BTW: You can use `%q` instead of `'%s'` to format a string with quotes. Even works for `[]string` (quotes each element).

luksa (approved):
LGTM

brian-avery (approved):

PR #111, opened 2019-07-03
MAISTRA-563 Fixed prometheus container suffixes, added working directory

dmitri-d:


tmp/build/patch-prometheus.sh
```diff
@@ -50,10 +50,13 @@ function prometheus_patch_deployment() {
...
-    sed -i -r -e 's/image: *prometheus/image: prometheus-ubi8/' ${HELM_DIR}/istio/charts/prometheus/templates/deployment.yaml
+    sed -i -r -e 's/image:(.*)prometheus:/image:\1prometheus-ubi8:/' ${HELM_DIR}/istio/charts/prometheus/templates/deployment.yaml
   else
-    sed -i -r -e 's/image: *prometheus/image: prometheus-rhel8/' ${HELM_DIR}/istio/charts/prometheus/templates/deployment.yaml
-  fi 
+    sed -i -r -e 's/image:(.*)prometheus:/image\1prometheus-rhel8:/' ${HELM_DIR}/istio/charts/prometheus/templates/deployment.yaml
+  fi
+
+  sed -i -e '/args:/ a\
+            - --storage.tsdb.path=/prometheus' ${HELM_DIR}/istio/charts/prometheus/templates/deployment.yaml
```
rcernich, line 59: What's this, and will accessing something from the root directory cause problems with permissions?

dmitri-d: Take a look here: https://github.com/Maistra/istio-images-centos/blob/maistra-0.12/Dockerfile.prometheus#L21. Note that the group is changed to 0 too (in OpenShift uid used to run the container is a member).

dmitri-d: This is a slightly newer version of prometheus (the same one as used by OpenShift) and it required to set the working dir explicitly.

rcernich: perfect, thanks!

rcernich (approved):

PR #117, opened 2019-07-04
MAISTRA-567 Remove hub & tag from prometheus' values.yaml

luksa:
Removing the hub & tag in values.yaml to make the values.yaml file
consistent with other components' (other components don't set them in
the file).

Also moved the setting of the -ubi8/-rhel8 suffix to patch-prometheus.sh
and moved the "image: prometheus" line up (where the hub & tag were).


jwendell:
Jira issue number is wrong

luksa:
@jwendell ugh, thanks.. I've fixed it here, but the @brian-avery's PR and my minor fix for it were already merged.

jwendell (approved):

PR #120, opened 2019-07-04
[MAISTRA-574] Restore anyuid privilege for grafana

jwendell:


knrc:
We should not be restoring anyuid but rather fixing the deployment so it works without it, I've moved the issue back to in progress and we should close this PR

helm/istio/charts/grafana/templates/meshclusterrole.yaml
```diff
@@ -8,6 +8,10 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
 rules:
+- apiGroups: ["security.openshift.io"]
+  resources: ["securitycontextconstraints"]
+  resourceNames: ["anyuid"]
+  verbs: ["use"]
```
dmitri-d, line 14: Grafana image we switched to is OpenShift-specific and shoudln't need anyuid. I'm looking into what the problem is.

knrc:
Closing in favour of https://github.com/Maistra/istio-operator/pull/122

PR #122, opened 2019-07-04
Removed anyuid from grafana replicaset

dmitri-d:


knrc (approved):

jwendell:
Did you try this on a real deployment? Did grafana work without that user id?

knrc:
Yes, @dmitri-d tested it before we merged

PR #127, opened 2019-07-05
Update 3scale version to 0.7.1: MAISTRA-498

knrc:


jwendell:
Missing the Jira number?

jwendell:
Also, I wonder if it's better just to update `helm/maistra-threescale/values.yaml` with the correct tag and just change it in patch-charts.sh to handle the `v` present in the community tag.

knrc:
@jwendell Yes, I missed the JIRA number so will push it again.  I'm happy with the modification being in the patch file for now since it is with the other changes we make, we can revisit this at a later date if it proves to be a problem.

knrc:
@jwendell I've updated the commit message

jwendell (approved):

rcernich (approved):

PR #128, opened 2019-07-05
Added new SAR/delegate-urls to Jaeger's OAuth Proxy

jpkrohling:
Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

### DIFF
--- a/helm/istio/charts/tracing/templates/jaeger-production-elasticsearch.yaml
+++ b/helm/istio/charts/tracing/templates/jaeger-production-elasticsearch.yaml
@@ -63,6 +63,9 @@ spec:
     enabled: {{ .Values.ingress.enabled }}
     # XXX: should this be parameterized?
     security: oauth-proxy
+    openshift:
+      sar: '{"namespace": "{{ .Release.Namespace }}", "resource": "pods", "verb": "get"}'
+      delegate-urls: '{"/":{"namespace": "{{ .Release.Namespace }}", "resource": "pods", "verb": "get"}}'
     annotations:
       {{- range $key, $value := .Values.ingress.annotations }}
         {{ $key }}: {{ $value | quote }}


jpkrohling:
Depends on a Jaeger Operator version that incorporates jaegertracing/jaeger-operator#508

The values are based on the ones from Grafana/Prometheus:

https://github.com/Maistra/istio-operator/blob/8a5ef75b875038b28d9d2f06d467ee545313f3e9/tmp/build/patch-grafana.sh#L37-L38

cc @mwringe, @objectiser

rcernich:
Why would we want to configure this?  I assume the Jaeger operator configures the proxy using a sensible authorization check already.

That said, this looks okay to me.

rcernich:
I think this should be sensibly defaulted in the jaeger operator itself.  I can't think of a reason why we'd want to change anything related to this.

jpkrohling:
Setting the `delegate-urls` means that the service account used by the pod containing the OAuth Proxy needs a cluster role binding with the role `system:auth-delegator`. Judging by [the comments](https://github.com/openshift/oauth-proxy/pull/59) I've seen in the OpenShift OAuth Proxy issue tracker, this is something that not all users might want, so, I'd rather not have this in the Jaeger Operator as the default, at least for the moment. 

If we later determine that this is indeed the right default for most use cases, we'll certainly be open to change it.

rcernich:
Judging by the docs, it also says this `--openshift-delegate-urls` is required to allow service account access.  It looks like we're giving those service accounts permission to create TokenReviews,, which I suspect is the equivalent of system:auth-delegator.

Given this, I wonder if we should configure the maistra proxies differently, or if there's another solution to this problem.  @jwendell, @knrc, opinions?

rcernich:
If you want this to be configurable, I think it would be better to just expose a boolean flag, instead of exposing the raw options to oauth-proxy exe, something like:

```yaml
  security: oauth-proxy
  openshift:
    enableTokenReview: true
```

jpkrohling:
`get pods on ${namespace}` might not be the right permission in some installations. I'd rather keep it flexible and compatible with OAuth Proxy for this first release, until we learn what's appropriate and what users really want/need.

rcernich:
That's fair.  I would assume you would know who should have access to your service (e.g. based on the data it collects and will provide), but if this is too simplistic, I understand.

That said, it appears as if this is a security concern for us as well, so we may need to rethink the way we are using `--openshift-delegate-urls` in our own deployments.

knrc:
Apparently we do something similar for prometheus/grafana so we need to revisit this, it's too much of a security risk if the deployments are compromised.

jpkrohling:
For the next release, we should at least set the SAR field, which should not require an elevated set of permissions.

jwendell:
This is covered in https://github.com/Maistra/istio-operator/pull/173, so we are safe to close this one.
Thanks @jpkrohling

PR #131, opened 2019-07-05
MAISTRA-353 install Istio CRDs during reconciliation instead of startup

rcernich:

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -9,13 +9,11 @@ import (
 	"time"
 
 	"github.com/maistra/istio-operator/pkg/apis"
-	"github.com/maistra/istio-operator/pkg/bootstrap"
 	"github.com/maistra/istio-operator/pkg/controller"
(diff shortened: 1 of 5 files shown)


cmd/manager/main.go
```diff
@@ -104,13 +102,7 @@ func main() {
...
 
 	// Create Service object to expose the metrics port.
-	metrics.ExposeMetricsPort()
-
-	// Enusure CRDs are installed
-	if err := bootstrap.InstallCRDs(mgr); err != nil {
-		log.Error(err, "failed to install CRDs")
-		os.Exit(1)
-	}
+	//metrics.ExposeMetricsPort()
```
jwendell, line 105: Do we want to leave this commented out?

jwendell (approved):
LGTM

cmd/manager/main.go
rcernich: yes.  it doesn't work.

PR #134, opened 2019-07-08
MAISTRA-581 Make CNI pod resource requests & limits configurable

luksa:


helm/istio/charts/istio_cni/templates/daemonset.yaml
```diff
@@ -98,12 +98,16 @@ spec:
             - mountPath: /host/etc/cni/
               name: etc-cni-dir
           resources:
+{{- if .Values.resources }}
+{{ toYaml .Values.resources | indent 12 }}
+{{- else }}
             requests:
```
jwendell, line 104: I think it doesn't make sense to check for the global values as well, right?

luksa: Yes, IMHO it does _not_ make sense to use global resource requests/limits values anywhere. Elsewhere, we are using global values, but here I just hardcoded the values that work on 4.1. Maybe we should increase them to 100Mi just to leave a bit more room.

jwendell: +1 on increasing to 100

luksa: OK, I've increased it to 100Mi. We can reduce it when we review the resource requests/limits of all the components.

jwendell (approved):

PR #137, opened 2019-07-09
MAISTRA-594 Install istio-node DaemonSet before other components

luksa:


pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -98,6 +98,7 @@ func (r *ControlPlaneReconciler) Reconcile() (reconcile.Result, error) {
 	// these components have to be installed in the specified order
 	orderedComponents := []string{
 		"istio", // core istio resources
+		"istio/charts/istio_cni",
```
rcernich, line 101: does this require any cert injection from citadel?

luksa: no.

rcernich (approved):

rcernich:
@knrc, this should be good to merge.

PR #140, opened 2019-07-10
MAISTRA-605 add UID and Extra fields to SubjectAccessReview calls

rcernich:


luksa (approved):
LGTM

PR #142, opened 2019-07-11
[kiali] put all kiali CRs in the kiali-operator namespace

jmazzitelli:
The Kiali CRs are now named `kiail-<control plane namespace>` (e.g. "kiali-istio-system") and all will be stored in the `kiali-operator` namespace.

This now requires the kiali operator to be watching only the namespace named `kiali-operator`. README instructions have been changed to indicate this.

The original code worked fine; this change is solely for security purposes.

Note: We can revert this code back to how it was before once the kiali operator has a validating webhook to prevent regular users from asking kiali to access namespaces that the user has no rights to.

_This is a PoC. Do not merge yet._

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ oc create -n observability -f https://raw.githubusercontent.com/jaegertracing/ja
 To install the Kiali operator, execute the following command:
 
 ```
-bash <(curl -L https://git.io/getLatestKialiOperator) --operator-image-version v1.0.0 --operator-watch-namespace '**' --accessible-namespaces '**' --operator-install-kiali false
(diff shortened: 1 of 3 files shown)


rcernich:
@jmazzitelli, what is the rationale behind moving the kiali cr into kiali-operator?

jmazzitelli:
> @jmazzitelli, what is the rationale behind moving the kiali cr into kiali-operator?

security issue otherwise.

jmazzitelli:
This seems like a false alarm. @mwringe - we can close this unless you can think of a reason not to.

mwringe:
I think we can close. It looks like normal users can't create, edit or even view the kiali CR resource. So they can't just modify the kiali CR to try and access everything from the cluster. It looks like only the maistra operator is the only one who is going to be updating these files.

PR #148, opened 2019-07-15
MAISTRA-621 fix conflict handling when updating namespace labels in s…

rcernich:
…mmr controller

### DIFF
--- a/pkg/controller/servicemesh/memberroll/controller.go
+++ b/pkg/controller/servicemesh/memberroll/controller.go
@@ -741,7 +741,7 @@ func (r *namespaceReconciler) reconcileNamespaceInMesh(namespace string) error {
 			err = r.client.Update(context.TODO(), namespaceResource)
 			if err != nil {
 				if errors.IsConflict(err) {
-					namespaceResource := &corev1.Namespace{}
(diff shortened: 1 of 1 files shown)


pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -741,7 +741,7 @@ func (r *namespaceReconciler) reconcileNamespaceInMesh(namespace string) error {
 			err = r.client.Update(context.TODO(), namespaceResource)
 			if err != nil {
 				if errors.IsConflict(err) {
-					namespaceResource := &corev1.Namespace{}
+					namespaceResource = &corev1.Namespace{}
```
luksa, line 744: Can probably remove this line altogether.

rcernich: Probably, and that was my gut instinct, but I wasn't sure whether or not the client would clear all the fields.  (I know for SAR in member roll validator, it doesn't clear all the status fields.)

luksa: Ah, ok. Then better not remove it for now.

luksa (approved):
LGTM

PR #151, opened 2019-07-16
[MAISTRA-625] Enable Istio CNI by default

jwendell:


rcernich:
I think we should query the api server to see if it supports `NetworkAttachmentDefinition` resources and if it does, then we should use it.  This would effectively mean that we remove this setting from the smcp spec.

jwendell:
Will close and open another one

PR #154, opened 2019-07-18
MAISTRA-578 cleanup ClusterRole for istio-operator

rcernich:


rcernich:
@dmitri-d, this is a cleaned up version of the `ClusterRole` required by the operator.  I still need to run some tests through bookinfo to verify that everything is working correctly.  Note, there are still a few items that need to be removed, but those require changes to some of the components.

deploy/servicemesh-operator.yaml
```diff
@@ -127,23 +127,27 @@ rules:
...
+  - namespaces
   - persistentvolumeclaims
-  - events
-  - configmaps
+  - pods
+  - replicationcontrollers
   - secrets
   - serviceaccounts
-  - namespaces
-  - routes
```
brian-avery, line 139: is routes still needed?

deploy/maistra-operator.yaml
```diff
@@ -127,23 +127,27 @@ rules:
...
+  - namespaces
   - persistentvolumeclaims
-  - events
-  - configmaps
+  - pods
+  - replicationcontrollers
   - secrets
   - serviceaccounts
-  - namespaces
-  - routes
```
brian-avery, line 139: is routes still needed?

deploy/servicemesh-operator.yaml
rcernich: It's namespaced now, route.openshift.io/v1 instead of just v1.

rcernich:
Also removes jaeger-agent service account, unsupported charts, and most instances of use anyuid scc.

jwendell (approved):

brian-avery (approved):

PR #157, opened 2019-07-18
[MAISTRA-625] Enable CNI unconditionally on OCP 4

jwendell:
We check for the presence of the NetworkAttachmentDefinition CRD and enable CNI
for all Control Planes in the cluster if it is present.

If it's not present, then we disable CNI, likely in OCP 3 environments.

This means that we are overriding whatever the user puts in the `istio_cni.enabled`
field of the SMCP definition.


pkg/controller/controller.go
```diff
@@ -2,17 +2,26 @@ package controller
...
 
 // AddToManagerFuncs is a list of functions to add all Controllers to the Manager
 var AddToManagerFuncs []func(manager.Manager) error
 
-// AddToManager adds all Controllers to the Manager
-func AddToManager(m manager.Manager) error {
+
+// Init initializes controller
+func Init(m manager.Manager) error {
+	common.InitCNIStatus(m)
```
rcernich: I think we should leave this function named `AddToManager()` and move this `InitCNIStatus()` code into the initialization of the smcp controller's `Add()` function.

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -110,6 +110,10 @@ func (r *ControlPlaneReconciler) Reconcile() (reconcile.Result, error) {
 		"istio/charts/grafana",
 		"istio/charts/kiali",
 	}
+	if !common.IsCNIEnabled {
```
rcernich: I think a better way is to force the `spec.istio.istio_cni.enabled` flag to true/false prior to rendering the charts.  Removing this from the list of ordered components will not prevent it from being installed.

pkg/controller/common/cni.go
```diff
@@ -0,0 +1,33 @@
...
+	netAttachDef := &unstructured.Unstructured{}
+	netAttachDef.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "k8s.cni.cncf.io",
+		Version: "v1",
+		Kind:    "NetworkAttachmentDefinition",
+	})
+
+	err := m.GetClient().Get(context.TODO(), client.ObjectKey{Namespace: "unexistent", Name: "unexistent"}, netAttachDef)
+
+	IsCNIEnabled = errors.IsNotFound(err)
```
rcernich: Shouldn't this be `IsNoMatchError(err)`.  If it's a nonexistent resource, it will be not found.  If the resource type doesn't exist, there will be a match error.

Besides that, is there a simpler way to detect whether or not the resource exists?  Maybe something like:

```go
_, err := mgr.GetRESTMapper().ResourcesFor(schema.GroupVersionResource{
		Group:   "k8s.cni.cncf.io",
		Version: "v1",
		Resource:    "network-attachment-definitions",
	})
```

helm/istio/charts/istio_cni/templates/serviceaccount.yaml
```diff
@@ -1,7 +1,5 @@
-{{ if .Values.enabled }}
```
rcernich: See my comment about setting this in the controller.  @luksa is doing something similar to hard code the namespace used for the cni resources (moving them into operator namespace).

Also, removing all these diff's to the templates will save @luksa from having to resolve conflicts in his PR.

cmd/manager/main.go
```diff
@@ -96,7 +96,7 @@ func main() {
 	}
 
 	// Setup all Controllers
-	if err := controller.AddToManager(mgr); err != nil {
+	if err := controller.Init(mgr); err != nil {
```
rcernich: See my comment below about this rename.

rcernich (changes requested):

pkg/controller/servicemesh/controlplane/reconciler.go
luksa: You definitely need to set {{istio_cni.enabled}} to true, because it's also checked in some charts (even the upstream charts use the setting).

jwendell:
PR updated

rcernich (approved):

luksa (approved):

PR #158, opened 2019-07-23
Update configuration to use the new names and version 1.0.0: MAISTRA-678

knrc:


helm/istio/charts/istio_cni/templates/daemonset.yaml
```diff
@@ -24,7 +24,7 @@ spec:
     metadata:
       labels:
         heritage: {{ .Release.Service }}
-        maistra-version: 0.12.0
+        maistra-version: 1.0.0
         release: {{ .Release.Name }}
```
jwendell, line 26: this label is added automatically by patch-charts.sh, no?

knrc: It wasn't but should be, the script hadn't been updated to do that

helm/istio/charts/istio_cni/templates/daemonset.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     heritage: {{ .Release.Service }}
-    maistra-version: 0.12.0
```
brian-avery, line 12: We auto-generate the version that the rest of Istio in the build process. Would it make sense to update these there as well?

brian-avery (approved):

helm/istio/charts/istio_cni/templates/daemonset.yaml
rcernich: Yes, I think these should be removed.  See @knrc's comment below.

brian-avery: Ah ok. I misunderstood that one as talking about the helm labels.

knrc: The script does not currently modify this version, I'll push an update which handles this.

helm/istio/charts/istio_cni/templates/daemonset.yaml
knrc: @jwendell you are correct the script does add in the label, the generated version contained two maistra-version labels so the daemonset originally had the 0.12.0 and 1.0.0 versions.  I'll tidy up the overlay version to fix this.

helm/istio/charts/istio_cni/templates/daemonset.yaml
knrc: I looked deeper into this and the script does handle it, the original overlay therefore resulted in two maistra-version labels being included within the daemonset

brian-avery (approved):

jwendell (approved):

rcernich (approved):

PR #159, opened 2019-07-23
MAISTRA-682 - Fixing Kiali CR image_pull_secrets

gbaufake:
On the Kiali CR,

it doesn't need to be used name which is used on helm. Because the kiali-operator does pass the name after receiving the array.

[kiali-kiali-fixed.yaml.log](https://github.com/Maistra/istio-operator/files/3422853/kiali-kiali-fixed.yaml.log)
[kiali-kiali-wrong.yaml.log](https://github.com/Maistra/istio-operator/files/3422854/kiali-kiali-wrong.yaml.log)

### DIFF
--- a/helm/istio/charts/kiali/templates/kiali-cr.yaml
+++ b/helm/istio/charts/kiali/templates/kiali-cr.yaml
@@ -22,7 +22,7 @@ spec:
 {{- if .Values.global.imagePullSecrets }}
     image_pull_secrets: 
 {{- range .Values.global.imagePullSecrets }}
-    - name: {{ . }}
+    - {{ . }}
 {{- end }}
 {{- end }}
     image_version: "{{ .Values.tag }}"


gbaufake:
cc @brian-avery

gbaufake:
After not using the name, 
kiali-operator deploys correctly with pull secret.
[pod-kiali-5dcd79f95f-8z9xm.yaml.log](https://github.com/Maistra/istio-operator/files/3422938/pod-kiali-5dcd79f95f-8z9xm.yaml.log)

helm/istio/charts/kiali/templates/kiali-cr.yaml
```diff
@@ -22,7 +22,7 @@ spec:
 {{- if .Values.global.imagePullSecrets }}
     image_pull_secrets: 
 {{- range .Values.global.imagePullSecrets }}
-    - name: {{ . }}
+    - {{ . }}
```
brian-avery, line 25: Note that Kubernetes style (https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/) is to use name: foo, but Kiali adds this internally when it applies the configuration rather than reading it out of the CR

brian-avery (approved):

helm/istio/charts/kiali/templates/kiali-cr.yaml
gbaufake: cc @jmazzitelli

gbaufake: I think it is because of the kiali_cr is an ansible based operator and following helm/Kubernertes patterns.

gbaufake:
cc @rcernich

rcernich (approved):

PR #160, opened 2019-07-23
WIP: Populate NetworkAttachmentDefinition with cni config

dmitri-d:
This populates ```NetworkAttachmentDefinition```s in member namespaces with cni config, which will take precedence over the configuration created by the daemonset.

### DIFF
--- a/pkg/controller/servicemesh/memberroll/controller.go
+++ b/pkg/controller/servicemesh/memberroll/controller.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"encoding/json"
+
 	"github.com/go-logr/logr"
 	pkgerrors "github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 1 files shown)


pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -831,9 +848,23 @@ func (r *namespaceReconciler) addNetworkAttachmentDefinition(namespace, meshName
 	}
 
 	// TODO: update resource if its state isn't what we want
-
 	netAttachDef.SetNamespace(namespace)
 	netAttachDef.SetName(name)
+	config := &runtimeCNIConfig{
```
rcernich, line 853: Should this config match what's in the config map?  If so, maybe we should read the config out of the config map and use that instead of hard coding it here.

dmitri-d: I don't think it would work in a single daemonset case (we create it once, but need separate cni configs for each mesh/cni version).

knrc:
This has been superceded by Marko's work

PR #163, opened 2019-07-24
MAISTRA-587: Watch galley, sidecar-injector secrets & update webhook configs on changes

dgn:
This still needs other changes:
- Istio components should not try to update webhook configs anymore (in review: https://github.com/Maistra/istio/pull/36)
- ~~ClusterRoles should be updated to remove permissions to do so~~
- ~~WebhookConfigs have to be created by the operator~~

Even without the changes in https://github.com/Maistra/istio/pull/36 merged, you can test with the images supplied in the comment below.


dgn:
You can use my operator image for testing: `docker.io/dgrimm/istio-operator:maistra-587`

Galley and sidecar-injector won't start because they don't understand the `--manageWebhookConfig` flag yet. For the CP to come up without errors, you can use the images I pushed to `docker.io/dgrimm/`: (assuming your CP lives in `istio-system`)

```
kubectl patch deployment -n istio-system istio-galley --type='json' -p='[{"op": "replace", "path": "/spec/template/spec/containers/0/image", "value":"docker.io/dgrimm/galley:maistra-587"}]'
kubectl patch deployment -n istio-system istio-sidecar-injector --type='json' -p='[{"op": "replace", "path": "/spec/template/spec/containers/0/image", "value":"docker.io/dgrimm/sidecar_injector:maistra-587"}]'
```

They contain the changes I made in https://github.com/Maistra/istio/pull/36 (plus a little fix for the health checks)

pkg/controller/servicemesh/webhook/controller.go
```diff
@@ -0,0 +1,212 @@
...
+		return reconcile.Result{}, nil
+	}
+	caRoot, ok := secret.Data[rootCertKey]
+	if !ok {
+		logger.Error(errors.New("secret does not contain root certificate"), "Incomplete secret")
+		return reconcile.Result{}, nil
+	}
+
+	// is this for galley or the sidecar-injector?
+	if request.Name == secretNames[0] {
```
jwendell: Given these two blocks are almost identical, why not put those in a function?

dgn: Because they operate on two different types with no common base except for `runtime.Object` - I guess I could use `unstructured` - but that would feel weird, exchanging type safety for brevity?

dgn: I can remove a bit of the duplication though - the part that does the `Client.Update` below is pretty generic

dgn: @jwendell I reduced duplication to a minimum without dropping the types. If you have an idea how to make it even more generic let me know!

helm/istio/charts/galley/templates/validatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,126 @@
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: ValidatingWebhookConfiguration
+metadata:
```
luksa, line 3: This section contains no `ownerReferences` as the original `ValidatingWebhookConfiguration` did. Is the owner added in the code?

pkg/controller/servicemesh/webhook/controller.go
```diff
@@ -0,0 +1,205 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+var (
+	log         = logf.Log.WithName("controller_webhook")
+	secretNames = []string{
```
luksa: insert empty line here, since log and secretNames aren't related? (will also remove the bunch of spaces in the first of the two lines)

pkg/controller/servicemesh/webhook/controller.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	// Watch MutatingWebhookConfigurations
+	err = c.Watch(&source.Kind{Type: &v1beta1.MutatingWebhookConfiguration{}}, &handler.EnqueueRequestsFromMapFunc{
+		ToRequests: handler.ToRequestsFunc(func(ns handler.MapObject) []reconcile.Request {
+			requests := []reconcile.Request{}
+			if strings.HasPrefix(ns.Meta.GetName(), injectorWebhookConfigNamePrefix) {
+				requests = append(requests, reconcile.Request{
+					NamespacedName: types.NamespacedName{
+						Name:      secretNames[0],
```
luksa: Nitpick: can you introduce `sidecarInjectorServiceAccountSecretName` instead of referring to `secretNames[0]`? It would be a bit nicer to anyone reading this code.

pkg/controller/servicemesh/webhook/controller.go
```diff
@@ -0,0 +1,205 @@
...
+			return false
+		},
+	})
+	if err != nil {
+		return err
+	}
+
+	// Watch MutatingWebhookConfigurations
+	err = c.Watch(&source.Kind{Type: &v1beta1.MutatingWebhookConfiguration{}}, &handler.EnqueueRequestsFromMapFunc{
+		ToRequests: handler.ToRequestsFunc(func(ns handler.MapObject) []reconcile.Request {
```
luksa: Rename `ns` to `obj` or `mwc` or similar.

pkg/controller/servicemesh/webhook/controller.go
```diff
@@ -0,0 +1,205 @@
...
+			}
+			return requests
+		})})
+	if err != nil {
+		return err
+	}
+
+	// Watch ValidatingWebhookConfigurations
+	err = c.Watch(&source.Kind{Type: &v1beta1.ValidatingWebhookConfiguration{}}, &handler.EnqueueRequestsFromMapFunc{
+		ToRequests: handler.ToRequestsFunc(func(ns handler.MapObject) []reconcile.Request {
```
luksa: Rename `ns` variable.

pkg/controller/servicemesh/webhook/controller.go
```diff
@@ -0,0 +1,205 @@
...
+// reconciles webhook configurations
+type reconciler struct {
+	common.ResourceManager
+	scheme *runtime.Scheme
+}
+
+// Reconcile reacts to changes in secrets by updating the corresponding webhook
+// configurations
+func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	logger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
```
luksa: I know this is how all our other controllers log things, but I'd like to see `Request.Namespace` changed to `Secret.Namespace` and `Request.Name` to `Secret.Name`. 

Actually, ideally this would just be `log.WithValues("Secret", request.Namespace + "/" + request.Name)`.

pkg/controller/servicemesh/webhook/controller.go
```diff
@@ -0,0 +1,205 @@
...
+	caRoot, ok := secret.Data[rootCertKey]
+	if !ok {
+		logger.Error(errors.New("secret does not contain root certificate"), "Incomplete secret")
+		return reconcile.Result{}, nil
+	}
+	updated := false
+	var config runtime.Object
+
+	// is this for galley or the sidecar-injector?
+	if request.Name == secretNames[0] {
```
luksa: As above, I'd replace `secretNames[0]` with `sidecarInjectorSecretName`. Then you can also remove the comment, since the code would be clear enough.

luksa (changes requested):
A few minor nitpicks, but I'm also concerned about the missing `ownerReference`.

helm/istio/charts/galley/templates/validatingwebhookconfiguration.yaml
dgn: I was assuming the operator would take care of that; but it explicitly doesn't for cluster-scoped resources: https://github.com/Maistra/istio-operator/blob/878bee2d374595b41bc5fa1ed6952a55f720b543/pkg/controller/servicemesh/controlplane/manifestprocessing.go#L129

dgn:
All comments should be addressed. I'm now also setting the `ownerReferences` the same way they were set before.

luksa (approved):

helm/istio/charts/galley/templates/meshclusterrole.yaml
```diff
@@ -8,9 +8,6 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
 rules:
-- apiGroups: ["admissionregistration.k8s.io"]
-  resources: ["validatingwebhookconfigurations"]
-  verbs: ["*"]
 # required to be able to set OwnerReference with BlockOwnerDeletion=true
```
rcernich: Can we remove the following too?  I thought we were going to get rid of the owner reference on the webhook.

helm/istio/charts/sidecarInjectorWebhook/templates/meshclusterrole.yaml
```diff
@@ -9,9 +9,6 @@ metadata:
     release: {{ .Release.Name }}
     istio: sidecar-injector
 rules:
-- apiGroups: ["admissionregistration.k8s.io"]
-  resources: ["mutatingwebhookconfigurations"]
-  verbs: ["create", "get", "list", "watch", "patch"]
 # required to be able to set OwnerReference with BlockOwnerDeletion=true
```
rcernich: Same comment as above, can we remove this if we're not setting owner references?

pkg/controller/servicemesh/webhook/controller.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	caRoot, ok := secret.Data[rootCertKey]
+	if !ok {
+		logger.Error(errors.New("secret does not contain root certificate"), "Incomplete secret")
+		return reconcile.Result{}, nil
+	}
+	updated := false
+	var config runtime.Object
+
+	if request.Name == sidecarInjectorServiceAccountSecretName {
```
rcernich: Would these work better as separate reconcilers, instead of if/else?  You could have one for mutating webhook and another for validating webhook, each would sync it's cabundle to the latest, so above, you'd be sending the webhook resources through the reconcile loop instead of the secrets.

Also, because you're setting the owner reference on the webhooks, you'll need to watch them for deletion and recreate them.  This could be tricky, as you need to know whether or not it's being deleted because of garbage collection or because the control plane is being deleted.  I think it would be easiest to remove the owner reference altogether, but this would block pod creation in member namespaces.  @knrc?

luksa: I've been researching the ownerReferences stuff today, and it's still true that we're not using them correctly. Creating an ownerReference on a cluster-scoped object that points to a namespace-scoped owner is not supported. Only the inverse case is supported (see https://github.com/kubernetes/kubernetes/blob/6e5176b48df8719ab969a233a0411ad6ba35d222/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L288-L290)

Instead of the adding the ownerReference and letting the GC ensure that the webhook is deleted when the Deployment goes away, if the SMCP still exists, the Operator should re-create the Deployment anyway. Having the GC delete the webhook is unnecessary and just complicates things (the Operator should re-create the webhook in that case). 

Previously, the problem was if someone deleted the control plane namespace. But now, if that happens, the SMCP is deleted as well, so the Operator should (and probably already does) delete the webhook automatically.

helm/istio/charts/sidecarInjectorWebhook/templates/mutatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,40 @@
...
+      matchExpressions:
+      - key: maistra.io/member-of
+        operator: In
+        values:
+        - {{ .Release.Namespace }}
+      - key: maistra.io/ignore-namespace
+        operator: DoesNotExist
+      - key: istio.openshift.com/ignore-namespace
+        operator: DoesNotExist
+      matchExpressions:
```
luksa: Duplicate `matchExpressions` field. (see line 27 above)

helm/istio/charts/sidecarInjectorWebhook/templates/mutatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - key: maistra.io/ignore-namespace
+        operator: DoesNotExist
+      - key: istio.openshift.com/ignore-namespace
+        operator: DoesNotExist
```
luksa: These 5 lines are duplicated.

jwendell: Also, I think we must get rid of `istio.openshift.com` stuff? @rcernich

pkg/controller/servicemesh/webhook/controller.go
dgn: > I've been researching the ownerReferences stuff today, and it's still true that we're not using them correctly. Creating an ownerReference on a cluster-scoped object that points to a namespace-scoped owner is not supported. Only the inverse case is supported (see https://github.com/kubernetes/kubernetes/blob/6e5176b48df8719ab969a233a0411ad6ba35d222/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L288-L290)

That's what I was referring to on IRC yesterday. The docs clearly state that what we're doing is not supported, however it behaves as expected: webhooks are removed when their namespace-scoped owner (the deployment) is deleted. This 'wrong' use of ownerReferences is coming from upstream.

> Instead of the adding the ownerReference and letting the GC ensure that the webhook is deleted when the Deployment goes away, if the SMCP still exists, the Operator should re-create the Deployment anyway. Having the GC delete the webhook is unnecessary and just complicates things (the Operator should re-create the webhook in that case).
> 
> Previously, the problem was if someone deleted the control plane namespace. But now, if that happens, the SMCP is deleted as well, so the Operator should (and probably already does) delete the webhook automatically.

I'm bringing up a new cluster right now to test through these use cases, but I have created and deleted dozens of SMCPs in the last few days and never had a single leftover webhook, neither with or without the ownerReferences. That was also the reason it never occurred to me to add them. If we're running the operator we shouldn't rely on the GC for cleanup IMO. But I don't know if that's best practice.

dgn: > Would these work better as separate reconcilers, instead of if/else? You could have one for mutating webhook and another for validating webhook, each would sync it's cabundle to the latest, so above, you'd be sending the webhook resources through the reconcile loop instead of the secrets.

I think that's a matter of taste. The biggest difference I see is that I would have to create two watches on `Secret`, one for each Webhook type. Possibly that is handled by controller-runtime internally by reusing the same watch. I don't know for sure though.

luksa: An ownerRef pointing from a cluster-scoped dependent to a namespace-scoped owner is conceptually wrong, because it's not clear what namespace the owner is in. It might appear to work, but there are no guarantees that it will always work. A few months ago, someone reported the webhook disappearing after a few days. It turned out it was caused by a race condition in the GC, which only occurred when you restarted the "cluster" (more specifically, it happend during startup of the GC). One such race condition was fixed, but for the inverted ownerRef (from a namespace-scoped dependent to a cluster-scoped owner). Our use-case continues to be unsupported, so I consider it very dangerous, but I suggest we deal with this in a different PR. Your changes now keep the ownerRef as it was.

dgn:
I removed the
```
      - key: istio.openshift.com/ignore-namespace
        operator: DoesNotExist
```
namespaceSelector, @rcernich let me know in case that's still needed.

pkg/controller/servicemesh/webhook/controller.go
rcernich: Some history (it took me a long time to understand the purpose behind all this):

The ownerReference was used so that if the deployment pods disappeared, the webhook would also be removed.  This ensured that other deployments would succeed, even if the webhook pods were not available.  This was primarily and issue when we were watching all namespaces.  The side effect was that the GC would periodically delete the webhook, because it couldn't resolve the owner (no namespace on a webhook), so the deployments implementing the webhooks were modified to check for deletion and recreate them.

That said, the operator properly cleans up these resources, and, if we're deleting the webhooks when the deployments aren't there, we can simply change the failure policy on the webhooks so pod creation can continue if the webhook invocation fails.  This removes the need for ownerReference entirely, while preserving the existing behavior.  (You could also argue that since we're not watching all namespaces anymore, we could let pod creation hang in the member namespaces until the webhook service comes back.)

> >  Would these work better as separate reconcilers, instead of if/else? You could have one for mutating webhook and another for validating webhook, each would sync it's cabundle to the latest, so above, you'd be sending the webhook resources through the reconcile loop instead of the secrets.

> I think that's a matter of taste. The biggest difference I see is that I would have to create two watches on Secret, one for each Webhook type. Possibly that is handled by controller-runtime internally by reusing the same watch. I don't know for sure though.

That is correct, two reconcilers, each with its own watch.  My rationale was that we're actually reconciling the webhooks, not the secrets, so running reconcilation against the secrets inverts the logic, imo.

rcernich:
I think `istio.openshift.com/ignore-namespace` can safely be removed, as we replaced it a few releases ago.

helm/istio/charts/sidecarInjectorWebhook/templates/mutatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: admissionregistration.k8s.io/v1beta1
```
rcernich, line 1: I wonder if this change should be moved into maistra/istio?  Upstream istio provides this file and the `.tpl` file was added by maistra.  In other words, should we revert part of these changes, maistra/istio#6?

dgn: I asked @knrc about rolling back those changes and he said he was assigned an upstream issue to implement them upstream, so no. See comments here: https://github.com/Maistra/istio/pull/36

dgn:
There are two open topics:

1. @rcernich suggested splitting the code into two controllers: I don't like that because it would create a lot of code duplication. What I do agree with though is that this should be reconciling webhook configs instead of secrets; I will make that change today. If there are strong sentiments towards splitting into two controllers, I can make that change as well.

2. As to the ownerReferences: I will leave them in there for now because that's how it's done upstream and it's also the behavior before this PR. Removing ownerReferences should be a separate PR IMO. @knrc do you have an opinion on this?

knrc:
OwnerReferences were important when we had webhooks which were receiving requests from the majority of the namespaces, with the switch over the multi tenancy they are not as important and can likely be dropped.

While we are technically using them in a way which is unsupported they do work in the use case we care about and we cover the other use case by having the components watching their own webhook configuration and recreating when it disappeared.  There is still a race with this setup, having the configuration disappear just as a pod is being created, however this was a better (read recoverable) situation from having the webhook configuration take down the cluster which was the alternative.

knrc:
> 2. As to the ownerReferences: I will leave them in there for now because that's how it's done upstream and it's also the behavior before this PR. Removing ownerReferences should be a separate PR IMO. @knrc do you have an opinion on this?

Having a separate PR makes sense to me.

dgn:
> While we are technically using them in a way which is unsupported they do work in the use case we care about and we cover the other use case by having the components watching their own webhook configuration and recreating when it disappeared.

Sorry, still wrapping my head around this... In this case we might actually want to remove them in this PR - because we move management of the webhooks into the operator, and the operator does not re-create them if they are deleted by the GC. This is because they are created only as part of the SMCP reconciliation, not constantly monitored.

rcernich:
@knrc, doesn't modifying the failure policy provide the same effective behavior, thus eliminating the need for the owner reference while preserving the desired behavior?  (If the pod behind the webhook service disappears or is unavailable, the webhook succeeds, which is effectively what would happen if the webhook were removed.)

dgn:
Alright, final round of changes (hopefully):
- ownerReferences removed again,
- reconciler reconciles webhook configs instead of secrets

rcernich (approved):
lgtm

pkg/controller/servicemesh/webhook/controller.go
```diff
@@ -0,0 +1,247 @@
...
+			return strings.HasPrefix(event.MetaNew.GetName(), galleyWebhookConfigNamePrefix)
+		},
+		// deletion and generic events don't interest us
+		DeleteFunc: func(event event.DeleteEvent) bool {
+			return false
+		},
+		GenericFunc: func(event event.GenericEvent) bool {
+			return false
+		},
+	})
```
luksa: The duplication in these predicate functions and the ones for mutating webhooks above can be removed by using a single predicate.Funcs with a configurable prefix.

pkg/controller/servicemesh/webhook/controller.go
```diff
@@ -0,0 +1,247 @@
...
+			logger.Info("could not get secret: " + err.Error())
+			return reconcile.Result{}, nil
+		}
+		// update caBundle if it doesn't match what's in the secret
+		newConfig := currentConfig.DeepCopyObject().(*v1beta1.MutatingWebhookConfiguration)
+		for i := range newConfig.Webhooks {
+			updated = injectCABundle(&newConfig.Webhooks[i].ClientConfig, caRoot) || updated
+		}
+		config = newConfig
+	} else if strings.HasPrefix(request.Name, galleyWebhookConfigNamePrefix) {
```
luksa: Ugh, so this reconciler reconciles two different kinds of objects. I haven't seen this pattern in any existing controller. IMHO, this really should be two different controllers.

pkg/controller/servicemesh/webhook/controller.go
```diff
@@ -0,0 +1,247 @@
...
+	err := r.Client.Get(context.TODO(), types.NamespacedName{
+		Namespace: namespace,
+		Name:      name,
+	}, secret)
+	if err != nil {
+		return nil, err
+	}
+	caBundle, ok := secret.Data[rootCertKey]
+	if !ok {
+		return nil, errors.New("secret " + namespace + "/" + name + " does not contain root certificate under key " + rootCertKey)
```
luksa: Use string formatting instead of concatenation?

pkg/controller/servicemesh/webhook/controller.go
dgn: Alright then, let's go there.

pkg/controller/servicemesh/webhook/controller.go
dgn: Will do

knrc:
(following up here for completeness, this should be what we discussed in the call yesterday)

> @knrc, doesn't modifying the failure policy provide the same effective behavior, thus eliminating the need for the owner reference while preserving the desired behavior? (If the pod behind the webhook service disappears or is unavailable, the webhook succeeds, which is effectively what would happen if the webhook were removed.)

Technically yes however the better behaviour is to cause the deployments to fail rather than allowing them to deploy if the webhook is not available.  We would have preferred this with the previous implementation however it would have rendered the cluster unusable since the webhook also covered the cluster control plane namespaces, with the switch to multi tenancy we now only impact the service mesh namespaces.

dgn:
Another round!
- I split up the code into two separate controllers
- some common code lives in `common/webhooks.go`

rcernich (approved):
lgtm

jwendell (approved):

PR #166, opened 2019-07-26
[MAISTRA-430] Add a '-version' CLI flag and output the version on sta…

jwendell:
…rtup

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,13 +25,43 @@ COMMUNITY       ?= true
 
 export MAIN_DIR OUT_DIR MAISTRA_BRANCH MAISTRA_VERSION COMMUNITY
 
+PACKAGE_NAME := github.com/maistra/istio-operator
+LD_EXTRAFLAGS =
+
+VERSION ?= development
+LD_EXTRAFLAGS += -X ${PACKAGE_NAME}/pkg/version.buildVersion=${VERSION}
+
+GITREVISION ?= $(shell git rev-parse --verify HEAD 2> /dev/null)
+ifeq ($(GITREVISION),)
+  GITREVISION = unknown
+endif
+LD_EXTRAFLAGS += -X ${PACKAGE_NAME}/pkg/version.buildGitRevision=${GITREVISION}
+
+GITSTATUS ?= $(shell git diff-index --quiet HEAD --  2> /dev/null; echo $$?)
+ifeq ($(GITSTATUS),0)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -25,13 +25,43 @@ COMMUNITY       ?= true
...
+VERSION ?= development
+LD_EXTRAFLAGS += -X ${PACKAGE_NAME}/pkg/version.buildVersion=${VERSION}
+
+GITREVISION ?= $(shell git rev-parse --verify HEAD 2> /dev/null)
+ifeq ($(GITREVISION),)
+  GITREVISION = unknown
+endif
+LD_EXTRAFLAGS += -X ${PACKAGE_NAME}/pkg/version.buildGitRevision=${GITREVISION}
+
+GITSTATUS ?= $(shell git diff-index --quiet HEAD --  2> /dev/null; echo $$?)
```
dmitri-d, line 40: What's the purpose of this field? This is information that end-users are going to see, I think it would be best to omit things that can cause confusion.

This could be a check in the build script that would refuse to proceed in the presence of uncommitted changes Doesn't rhpkg/fedpkg do that already?.

jwendell: This is mostly informational. No checks or blockers. It's mimic'ing istio upstream output for their `version` commands.

dmitri-d: Hmm, ok. I'm not convinced this is a useful field, but if you think it's important to have it, I'm ok with it.

jwendell:
This is how the output looks like:
```
./_output/bin/istio-operator -version
version.BuildInfo{Version:"1.1.0", GitRevision:"45acc115938190dc9264b0c6418edf2b818a9815", BuildStatus:"Modified", GitTag:"maistra-0.12.0-33-g45acc11", GoVersion:"go1.12.7", GoArch:"linux/amd64", OperatorSDK:"v0.2.1"}
```

brian-avery (approved):

Makefile
jwendell: I don't have a strong opinion. It's nice to have, but if you think it will cause more harm than good I can remove it quickly.

dmitri-d: I don't think it's a big deal. We can remove it if we find it causes confusion.

brian-avery: I'll vote a yes to keeping it. Shows you what commit your on and is consistent with other software.

dmitri-d (approved):

PR #167, opened 2019-07-26
 MAISTRA-435 asynchronously check deployment state when reconciling SMCP

rcernich:

Also:
* add readiness condition to SMCP resource
* removed support for legacy ControlPlane type
* MAISTRA-557 - use standard `app.kubernetes.io` labels


deploy/maistra-operator.yaml
```diff
@@ -33,6 +15,19 @@ spec:
   subresources:
     status: {}
   version: v1
+  additionalPrinterColumns:
+  - JSONPath: .status.conditions[?(@.type=="Ready")].status
+    name: Ready
+    type: string
+  - JSONPath: .status.conditions[?(@.type=="Ready")].message
+    name: Status
+    type: string
```
luksa, line 26: This could be fairly wide, so it should only be printed when using `-o wide`. this is achieved by adding a `priority` field to the column definition.

pkg/apis/maistra/v1/status.go
```diff
@@ -47,19 +47,23 @@ type ComponentStatus struct {
...
 type ConditionType string
 
 const (
-	// ConditionTypeInitialized signifies the whether or not the controller has
-	// initialized the CR.
-	ConditionTypeInitialized ConditionType = "Initialized"
+	// ConditionTypeReconciling signifies that the controller is reconciling the CR.
+	ConditionTypeReconciling ConditionType = "Reconciling"
+	// ConditionTypeDeleting signifies that the controller is deleting the CR.
+	ConditionTypeDeleting ConditionType = "Deleting"
```
luksa: Why introduce these two conditions, when we already have `Reconciled` below?

rcernich:
This is ready to go.  Also includes some minimal event recording.

pkg/controller/common/metadata.go
```diff
@@ -27,6 +27,15 @@ const (
...
 	InternalKey = MetadataNamespace + "/internal"
+
+	// KubernetesAppNamespace is the common namespace for application information
+	KubernetesAppNamespace = "app.kubernetes.io"
+	KubernetesAppNameKey = KubernetesAppNamespace + "/name"
+	KubernetesAppInstanceKey = KubernetesAppNamespace + "/instance"
+	KubernetesAppVersionKey = KubernetesAppNamespace + "/version"
+	KubernetesAppComponentKey = KubernetesAppNamespace + "/component"
+	KubernetesAppPartOfKey = KubernetesAppNamespace + "/part-of"
+	KubernetesAppManagedByKey = KubernetesAppNamespace + "/managed-by"
```
luksa, line 38: Nitpick: I prefer to define externally visible constants as a single string, so that any users searching for the constant in the source can easily find it.

pkg/controller/servicemesh/controlplane/controller.go
```diff
@@ -53,36 +64,61 @@ func newReconciler(mgr manager.Manager, operatorNamespace string) reconcile.Reco
...
+		}
+		return []reconcile.Request{
+			reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Name: owner.Name, Namespace: ownedResource.Meta.GetNamespace(),
+				},
+			},
+		}
+	}),
+}
```
luksa: Doesn't the existing `handler.EnqueueRequestForOwner` do the same thing?

deploy/maistra-operator.yaml
rcernich: ...so add the wide property to the status fields?

pkg/apis/maistra/v1/status.go
rcernich: `Deleting` should be deleted, as it's never used.  `Reconciling` is used as a flag to allow reentry into the reconcile loop, and cleanup of the local cache of instance reconcilers when reconciliation completes.  That said, writing this out, reconcile will never complete if there's an error, which means we'd never reach a `Reconciled` state, so I guess these both could be removed.

pkg/controller/servicemesh/controlplane/controller.go
rcernich: We're mapping from a `Deployment` event to a `ServiceMeshControlPlane` event, so this is similar to what we're doing in the `ServiceMeshMemberRoll` controller for namespaces..

That said, it looks like there's an 'EnqueueRequestForOwner` that would do the trick.

deploy/maistra-operator.yaml
rcernich: added `priority` fields

pkg/apis/maistra/v1/status.go
rcernich: Removed both of these.

pkg/controller/common/metadata.go
rcernich: For me, this saves typo's.  I think I've seen it both ways.  I usually end up searching for the root of the key, then search using the variable name(s).

pkg/controller/servicemesh/controlplane/controller.go
rcernich: Updated to use `EnqueueRequestForOwner`.

rcernich:
Last push incrementally updates the condition status to give users a better idea of what's going on (e.g. "Paused until galley becomes ready", "Progressing", etc.).

rcernich:
@luksa, if you're okay with the updates, could you mark your comments as resolved?

pkg/controller/servicemesh/controlplane/controller.go
```diff
@@ -124,25 +156,15 @@ func (r *ReconcileControlPlane) Reconcile(request reconcile.Request) (reconcile.
...
+	reconciler := r.getOrCreateReconciler(instance)
+	defer r.deleteReconciler(reconciler)
 	deleted := instance.GetDeletionTimestamp() != nil
 	finalizers := instance.GetFinalizers()
 	finalizerIndex := common.IndexOf(finalizers, finalizer)
 
 	if deleted {
 		if finalizerIndex < 0 {
-			reqLogger.Info("ServiceMeshControlPlane deleted")
+			reqLogger.Info("Deletion ServiceMeshControlPlane complete")
```
luksa: missing "of"

rcernich: done

luksa (approved):

pkg/controller/common/metadata.go
brian-avery: My vote here is for the combination of the two. Each of the endpoint names here is pretty vague and could match multiple places.

brian-avery (approved):

PR #170, opened 2019-07-29
[MAISTRA-695] Roll back changes to the build system

jwendell:


knrc (approved):

brian-avery (approved):

rcernich:
did you really want to revert maistra-430?

knrc:
@rcernich I believe it is one of the issues which came after the switch to Makefile so it will need to be reopened.

PR #174, opened 2019-08-01
MAISTRA-705 Return error when SMCP finalizer can't be removed

luksa:


rcernich (approved):

pkg/controller/servicemesh/controlplane/controller.go
```diff
@@ -193,10 +193,13 @@ func (r *ReconcileControlPlane) Reconcile(request reconcile.Request) (reconcile.
...
 		if finalizerError != nil {
-			reqLogger.Error(finalizerError, "error removing finalizer")
 			r.Manager.GetRecorder(controllerName).Event(instance, "Warning", "ServiceMeshDeleted", fmt.Sprintf("Error occurred removing finalizer from service mesh: %s", finalizerError))
 		}
-		return result, err
+		if err != nil {
+			// original error is more important than finalizerError
+			return result, err
+		}
+		return result, fmt.Errorf("Error removing finalizer from ServiceMeshControlPlane: %v", finalizerError)
```
dgn: Won't this throw an error even if `err` and `finalizerError` are both `nil`?

luksa: OMG, what was I thinking... thanks for catching this.

Initially, I was just returning `result, finalizerError`, which was OK, but later I decided to wrap the error and now this is total crap. :)

luksa: Plus, I should have properly wrapped the error using `errors.Wrap()`

luksa: Nice. `errors.Wrapf()` returns nil, when you wrap a nil error, so had used it, the code would have been perfectly fine.  Thanks again for spotting this.

dgn (approved):

PR #176, opened 2019-08-01
[maistra-367] [maistra-701] SMCP templates

brian-avery:


pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -208,20 +231,116 @@ func (r *ControlPlaneReconciler) Reconcile() (reconcile.Result, error) {
...
+		if _, exists := base[key]; !exists {
+			base[key] = value
+			continue
+		}
+
+		// at this point, key exists in both input and base.
+		// If both are maps, recurse.
+		// If only input is a map, ignore it. We don't want to overrwrite base.
+		// If both are values, again, ignore it since we don't want to overrwrite base.
+		if baseKeyAsMap, baseOK := base[key].(map[string]interface{}); baseOK {
```
dmitri-d, line 237: Are values in the base config are supposed to be the default values? If so, shouldn't values in the input map take precedence?

pkg/controller/servicemesh/controlplane/controller_test.go
```diff
@@ -0,0 +1,98 @@
+package controlplane
```
dmitri-d, line 1: yay, tests!

deploy/smcp-templates/product
```diff
@@ -0,0 +1,37 @@
+apiVersion: maistra.io/v1
```
dmitri-d: nit: should the name of the template be "service-mesh", not "product"?

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -49,10 +52,30 @@ var orderedCharts = []string{
 	"istio/charts/kiali",
 }
 
+const (
+	smcpTemplatePath        = "/etc/istio-operator/templates/"
```
dmitri-d: we probably want to keep the templates in ```/usr/share/istio-operator```, as this is an immutable application data, not application config.

rcernich: Just to keep terminology consistent are these `profiles` or `templates`?

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -208,20 +231,116 @@ func (r *ControlPlaneReconciler) Reconcile() (reconcile.Result, error) {
...
+		//configmaps mounted in directories with pre-existing content
+		defaultTemplateContent, defaultErr := ioutil.ReadFile(smcpDefaultTemplatePath + name)
+		if defaultErr != nil {
+			return v1.ControlPlaneSpec{}, fmt.Errorf("template cannot be loaded from user or default directory. Error from user: %s. Error from default: %s", err, defaultErr)
+		}
+		templateContent = defaultTemplateContent
+	}
+
+	var template v1.ServiceMeshControlPlane
+	if err = yaml.Unmarshal(templateContent, &template); err != nil {
```
rcernich, line 264: Should we be using k8s decoder for this?  I believe that's the proper way to deserialize a k8s resource.

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -49,10 +52,30 @@ var orderedCharts = []string{
...
+	//render the ServiceMeshControlPlane templates
+	spec, err := r.updateSMCPWithTemplates(r.Instance.Spec)
+	if err != nil {
+		r.Log.Info(fmt.Sprintf("Warning: failed to render ServiceMeshControlPlane templates:%s", err))
+
+		//TODO: Remove these comments we've decided that templates are stable.
+		//updateReconcileStatus(&r.Instance.Status.StatusType, err)
+		//return reconcile.Result{}, err
+	} else {
+		r.Instance.Spec = spec
```
rcernich: We should not be modifying any user specified fields.  We should store the result in the status, as you have below and use that value when rendering the templates.

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -208,20 +231,116 @@ func (r *ControlPlaneReconciler) Reconcile() (reconcile.Result, error) {
...
+
+	var template v1.ServiceMeshControlPlane
+	if err = yaml.Unmarshal(templateContent, &template); err != nil {
+		return v1.ControlPlaneSpec{}, fmt.Errorf("failed to parse template %s contents: %s", name, err)
+
+	}
+	return template.Spec, nil
+}
+
+func (r *ControlPlaneReconciler) renderSMCPTemplates(smcp v1.ControlPlaneSpec, visited map[string]struct{}) (v1.ControlPlaneSpec, error) {
```
rcernich: nit, on the naming.  We're not rendering anything here.  Maybe something like, `applyTemplate()`?

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -208,20 +231,116 @@ func (r *ControlPlaneReconciler) Reconcile() (reconcile.Result, error) {
...
+	}
+
+	visited[smcp.Template] = struct{}{}
+
+	smcp.Istio = mergeValues(smcp.Istio, template.Istio)
+	smcp.ThreeScale = mergeValues(smcp.ThreeScale, template.ThreeScale)
+	return smcp, nil
+}
+
+func (r *ControlPlaneReconciler) updateSMCPWithTemplates(smcpSpec v1.ControlPlaneSpec) (v1.ControlPlaneSpec, error) {
```
rcernich: nit, naming.  maybe `applyTemplates()`.  conflicts a bit with `applyTemplate()` above, but it seems like this is applying the templates to the smcp resource, while the above is recursive, going through each template referenced by the template.

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -208,20 +231,116 @@ func (r *ControlPlaneReconciler) Reconcile() (reconcile.Result, error) {
...
 	r.Log.Info("rendering helm charts")
 	allErrors := []error{}
 	var err error
 	var threeScaleRenderings map[string][]manifest.Manifest
 
-	if globalValues, ok := r.Instance.Spec.Istio["global"].(map[string]interface{}); ok {
-		globalValues["operatorNamespace"] = r.OperatorNamespace
-	} else {
-		return fmt.Errorf("Could not set operatorNamespace value, as .Values.global is not a map[string]interface{}: %v", err)
+	if r.Instance.Spec.Istio == nil {
```
rcernich: I think we should be using `r.Instance.Status.LastAppliedConfiguration` here, as mentioned above.  If we're doing that, I don't think any of these other changes are necessary, other than to operate on the status value instead of the spec.

rcernich (changes requested):

pkg/controller/servicemesh/controlplane/reconciler.go
brian-avery: The code between here and 341 is a fix for 701. Currently, if an SMCP is created without an Istio section, the operator will crash every time it goes to process the config because there was no check for nil on the Istio field. Also, if a user is using templates, the following a perfectly valid SMCP:

`apiVersion: maistra.io/v1
kind: ServiceMeshControlPlane
metadata:
  name: basic-install
spec:
  template: maistra`

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -49,10 +52,30 @@ var orderedCharts = []string{
...
+	smcpDefaultTemplatePath = "/etc/istio-operator/default-templates/"
+	smcpDefaultTemplate     = "default"
+)
+
 func (r *ControlPlaneReconciler) Reconcile() (reconcile.Result, error) {
 	var err error
 	var ready bool
 
+	//render the ServiceMeshControlPlane templates
+	spec, err := r.updateSMCPWithTemplates(r.Instance.Spec)
```
rcernich: I think this should be moved into `renderCharts()`.

rcernich: We should probably only be doing this when we render the charts, which is not on every reconcile, as we may be coming back after pausing for a component to become available.

pkg/controller/servicemesh/controlplane/reconciler.go
dgn: I think the naming is a bit misleading, but we're calling this with user-supplied config as `base` and the template config as `input`, so this behavior is correct.

deploy/smcp-templates/base
```diff
@@ -0,0 +1,48 @@
+apiVersion: maistra.io/v1
```
luksa, line 1: How come this file is called `base` and not `base.yaml`?

brian-avery: The filenames are used as the name of the template. If the user decides to create a file named foo.yaml then the template needs to be called foo.yaml. Same with JSON. Since extension doesn't matter in Linux, I removed it to clean things up a bit.

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -49,10 +52,30 @@ var orderedCharts = []string{
 	"istio/charts/kiali",
 }
 
+const (
+	smcpTemplatePath        = "/etc/istio-operator/templates/"
+	smcpDefaultTemplatePath = "/etc/istio-operator/default-templates/"
+	smcpDefaultTemplate     = "default"
```
dgn, line 57: I was wondering if we could use this `default` template to replace our default-overrides that we currently do via `sed`? We'd have to make sure all those defaults end up in the other templates as well, of course.

brian-avery: That's my hope here. Everything except for a couple of deletes is between base, Maistra, and product. Maistra specific stuff is what was community before. Product stuff is what was product before. Base is everything else. Product and Maistra use base as their template.

brian-avery: The build process will add either Maistra or product as default.

rcernich: I like this idea.  Have you tried using `null` for the values being deleted?

dgn: awesome!

pkg/controller/servicemesh/controlplane/reconciler.go
brian-avery: I've been using the term templates since this is a template for a user to overlay their custom settings on top of rather than a profile which to me means a complete set of configuration assigned to a specific user.

brian-avery: **profile** --  the personal details, images, user statistics, social media timeline, etc., that an individual creates and associates with a username or online account

**template** -- A template is a file that serves as a starting point for a new document.

pkg/controller/servicemesh/controlplane/reconciler.go
brian-avery: Fixed

pkg/controller/servicemesh/controlplane/reconciler.go
brian-avery: Done

pkg/controller/servicemesh/controlplane/reconciler.go
brian-avery: Fixed

pkg/controller/servicemesh/controlplane/reconciler.go
brian-avery: Correct. This one is "applying" the template, or updating the SMCP with the associated template values. Whereas the other one traverses all templates before processing them.

pkg/controller/servicemesh/controlplane/reconciler.go
brian-avery: Done

pkg/controller/servicemesh/controlplane/reconciler.go
brian-avery: Done

deploy/smcp-templates/product
brian-avery: Done

pkg/controller/servicemesh/controlplane/reconciler.go
brian-avery: This relies on the scheme to create the required resources, which is nil in the controlplane. Leaving as is for now. Created maistra-720 to look into this.

pkg/controller/servicemesh/controlplane/reconciler.go
rcernich: Would something like `pattern` work better?  Template seems a bit overused.

pkg/apis/maistra/v1/servicemeshcontrolplane_types.go
```diff
@@ -44,21 +44,28 @@ type ControlPlaneStatus struct {
...
+	LastAppliedConfiguration ControlPlaneSpec `json:"lastAppliedConfiguration"`
 }
 
 // HelmValuesType is typedef for Helm .Values
 type HelmValuesType map[string]interface{}
 
 // ControlPlaneSpec represents the configuration for installing a control plane
 type ControlPlaneSpec struct {
+	//Template selects the template to use for default values
+	Template string `json:"template,omitempty"`
```
rcernich, line 58: Would it make sense to have this be a list?  For example, a user specifies something like:

```yaml
spec:
  template:
  - full
  - auth
```

brian-avery: Think we discussed this with Kevin before and decided not to for now. This would also bring up questions like in what order do things overlay each other. Instead we have a hierarchical inheritance.

tmp/build/patch-jaeger.sh
```diff
@@ -1,5 +1,12 @@
 #!/usr/bin/env bash
 
+function jaeger_patch_values() {
+	# update jaeger zipkin port name
+  sed -i -e '/service:$/,/externalPort:/ {
```
brian-avery, line 5: @rcernich No idea what this is doing. Doesn't look like it gets reflected in the final values.yaml. Reverted for now.

deploy/smcp-templates/servicemesh
```diff
@@ -0,0 +1,40 @@
...
+      tag: 1.0.0
+    sidecarInjectorWebhook:
+      image: sidecar-injector-rhel8
+    gateways:
+      istio-ingressgateway:
+        ior_image: istio-ior-rhel8
+    prometheus:
+      image: prometheus-rhel8
+  threeScale:
+    hub: registry.redhat.io/openshift-istio-tech-preview
```
philipgough: @brian-avery remove `openshift-istio-tech-preview` and replace with `openshift-service-mesh` here and elsewhere?

pkg/apis/maistra/v1/servicemeshcontrolplane_types.go
rcernich: complication here is that it would require an api change, so better to decide now, as we'll have to live with it for a while.

as for the logic, why not apply them in order, full then auth, given the example above.  nothing complicated about that.

rcernich: adding some color, this allows for composition, which avoids an expansion of templates (full-auth, minimal-auth, etc.), "prefer composition over inheritance."

pkg/controller/servicemesh/controlplane/reconciler.go
rcernich: This is nil, https://github.com/Maistra/istio-operator/blob/maistra-1.0/pkg/controller/servicemesh/controlplane/controller.go#L132?  Somehow the scheme is there, because the client uses it to deserialize the control plane object you're working on.  https://github.com/Maistra/istio-operator/blob/maistra-1.0/cmd/manager/main.go#L93

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -208,34 +216,150 @@ func (r *ControlPlaneReconciler) Reconcile() (reconcile.Result, error) {
...
+	r.Status.LastAppliedConfiguration = spec
+
+	if err := r.validateSMCPSpec(r.Status.LastAppliedConfiguration); err != nil {
+		return err
+	}
 
-	if globalValues, ok := r.Instance.Spec.Istio["global"].(map[string]interface{}); ok {
+	if globalValues, ok := r.Status.LastAppliedConfiguration.Istio["global"].(map[string]interface{}); ok {
 		globalValues["operatorNamespace"] = r.OperatorNamespace
-	} else {
```
rcernich, line 219: Why did you remove this?  This is required for the cni charts.  If we can't set this, then we can't proceed with the rendering/install.

rcernich: ..,or we should proceed as below and create a map for the `global` field.

brian-avery: @rcernich I moved validation of the global section being present into the SMCP validation (312).

brian-avery: r.Status.LastAppliedConfiguration.Istio["global"]

still has the same problem as the previous code if the above layers aren't checked. If .Istio is nil, then attempting to access the global field will crash.

And the global section not being present at all is likely an issue even outside of setting the operatornamespace. I.e. what tag do you use?

tmp/build/patch-jaeger.sh
```diff
@@ -35,9 +22,9 @@ function jaeger_patch_production() {
 function JaegerPatch() {
   echo "Patching Jaeger"
 
-  jaeger_patch_values
```
rcernich, line 38: you've removed the invocation, but not the function

rcernich: sorry, didn't see that it was moved below

rcernich (approved):

tmp/build/patch-charts.sh
```diff
@@ -78,37 +66,17 @@ function patchTemplates() {
...
-    sed -i -e 's/hub:.*$/hub: kiali/' \
-           -e 's/tag:.*$/tag: v'${KIALI_VERSION}'/' ${HELM_DIR}/istio/charts/kiali/values.yaml
-  else
-    sed -i -e 's+hub:.*$+hub: registry\.redhat\.io\/openshift-service-mesh+' \
-           -e 's/tag:.*$/tag: '${KIALI_VERSION}'/' ${HELM_DIR}/istio/charts/kiali/values.yaml
-  fi
-
-  # In Istio 1.2, this viewOnlyMode is there, but in Istio 1.1 we need to add it - it is supported by the latest Kiali
-  sed -i -e 's/grafanaURL:/viewOnlyMode: false\n  grafanaURL:/' ${HELM_DIR}/istio/charts/kiali/values.yaml
+	#CNI is handled separately
```
dmitri-d, line 74: I don't think this is needed anymore -- cni image location is now being specified in the operator deployment ENV.

brian-avery: Yeah, I'm not sure so I added it back in to be safe. That said, I haven't been able to find anywhere it's actually used.

dmitri-d (approved):
I think updating location of CNI images can be safely omitted, but it doesn't have affect on the rest of the functionality.

lgtm

PR #177, opened 2019-08-01
[MAISTRA-430] Add a -version flag and prints the version on startup

jwendell:


jwendell:
This is the same of https://github.com/Maistra/istio-operator/pull/166 but without using Makefile.

tmp/build/build.sh
```diff
@@ -14,5 +13,29 @@ mkdir -p ${BIN_DIR}
 PROJECT_NAME="istio-operator"
 REPO_PATH="github.com/maistra/istio-operator"
 BUILD_PATH="${REPO_PATH}/cmd/manager"
+
+: ${VERSION:=development}
```
rcernich, line 17: Should this be an actual version, even if for development, e.g. `${VERSION:-1.0.0-dev}

jwendell: It could be. RPM's will be overriding these variables. Default values are used only for developers while developing Operator. That said, I don't have a strong opinion. If we choose to do that, we should remember bumping it every new [even micro?] version.

rcernich: Yeah, development just seems really vague.  Let's leave it for now, dev can always override it on the build, `VERSION=1.x-dev .... ./tmp/build.sh`

rcernich (approved):

pkg/version/version.go
```diff
@@ -0,0 +1,45 @@
...
+func (b BuildInfo) String() string {
+	return fmt.Sprintf("%#v", b)
+}
+
+func init() {
+	Info = BuildInfo{
+		Version:     buildVersion,
+		GitRevision: buildGitRevision,
+		BuildStatus: buildStatus,
+		GitTag:      buildTag,
```
brian-avery, line 40: Not sure GitTag is the proper name for this since version-release won't exist as a Git tag

dmitri-d:
lgtm

pkg/version/version.go
rcernich: I think it just shows up as "unknown" then.

brian-avery: @rcernich Apologies, pulled this in from the other PR. There it's using $version-$release. So if we do 15 internal builds and a single external build, then the external build is 1.0.0 and the GitTag is 1.0.0-15 while the actual Git tag is 1.0.0

jwendell: I think we should move this discussion to the other PR :)

brian-avery: Sgtm. It shouldn't have been in this one to begin with.

brian-avery (approved):

PR #179, opened 2019-08-02
MAISTRA-577 Allow users to specify oauth-proxy hub, tag and imagePullPolicy

dgn:
Questions: 
- where should I document these new config options?
- is oauth-proxy hosted elsewhere, too?
- is v1.1.0 really the latest version?

As soon as https://github.com/Maistra/istio-operator/pull/176/files is merged, we can move this into a template.


tmp/build/patch-charts.sh
```diff
@@ -252,6 +252,13 @@ function patchTemplates() {
     sed -i -e 's/\/health/\/tmp\/health/' ${HELM_DIR}/istio/charts/sidecarInjectorWebhook/templates/deployment.yaml
   fi
 
+  # add oauth-proxy default hub, tag and imagePullPolicy
```
rcernich: I think we need to distinguish between community and product.  Sadly, it seems there are two versions for product, openshift3/oauth-proxy and openshift4/ose-oauth-proxy, along with version specific tags (e.g. 3.11, 4.0, 4.1, etc.).  There also appears to be an openshift/origin-oauth-proxy on quay.io, which is specific to 4.x.

I wonder if there is some metadata we can pull from openshift to determine some of these settings (e.g. namespace [openshift, openshift3, openshift4], name prefix ["", ose, origin], tag [3.0, 4.1, etc.]?

knrc (approved):

tmp/build/patch-charts.sh
jwendell: So, currently we are using the hard coded value openshift/oauth-proxy (always community). With this change we give the users the option to override that value, which is better than we currently have.

I agree that having better defaults is better. Since we are not officially supporting OCP 3, I think we can go with product for 4. Users still have the power to override it if they want.

tmp/build/patch-charts.sh
```diff
@@ -252,6 +252,13 @@ function patchTemplates() {
...
+  sed -i -e '/^global:/a \
+\  oauthproxy:\
+\    hub: docker.io/openshift\
+\    tag: v1.1.0\
```
jwendell: For community, go with latest, as 1.1.0 is way outdated.

jwendell (changes requested):

tmp/build/patch-charts.sh
knrc: As long as you don't actually mean use the 'latest' tag, that's definitely out.  Which version have we been testing with?

jwendell: I literally meant `latest` tag. There are only three tags available: `latest`, one month old, `v1.1.0` more than one year old and `v1.0.0`, 2 years old. I've tested my changes on `latest`.

rcernich: This was the thrust of my comment above.  Should we be using one of the images from registry.redhat.io for product?

jwendell: IMO, yes (see my comment above). It's worth to note that jaeger, for instance, is also using docker hub for product (I believe they're going to work on this).

dgn:
I'm not sure what the requested changes are, so I'll try to recap:
- @jwendell you're suggesting different defaults. Can you point me to those? I tried contacting someone who is working on oauth-proxy but haven't gotten a reply yet.
- @rcernich your proposal is different default values for community vs product. Do you have any hints on where I can find productised images of oauth-proxy?

jwendell:
@dgn 
1. I'm seconding Rob's proposal of having different values for product and community
1. For community, IMO you could stick with openshift/oauth-proxy:latest since it's what been maintained (other tags are ancient)
1. For product, that's the main question. It seems there are different repos and images. I suggested that we use one specific for OCP 4 and if users want to use one for OCP 3, they can override that on the CR. So, IMO, you should use `registry.redhat.io/openshift4/ose-oauth-proxy:4.1`

dgn:
I pushed some changes:
- using different images for product vs community now
- image name itself is also configurable now, because oauth-proxy for ocp4 has a different image name than the one for ocp3.11
- product image defaults to 4.1 image
- community uses :latest (@knrc you don't like that; however the v1.1.0 image is very old. let me know if I should revert to using v1.1.0 )

tmp/build/patch-charts.sh
```diff
@@ -250,6 +250,22 @@ function patchTemplates() {
...
   fi
 
+  # add oauth-proxy default hub, image, tag and imagePullPolicy
+  if [[ "${COMMUNITY,,}" == "true" ]]; then
+    sed -i -e '/^global:/a \
+\  oauthproxy:\
+\    hub: docker.io/openshift\
+\    image: oauth-proxy\
+\    tag: latest\
+\    imagePullPolicy: IfNotPresent\n' ${HELM_DIR}/istio/values.yaml
```
jwendell, line 271: Istio upstream has this rule: For `latest` tag, policy is `Always`; for other tags, it's `IfNotPresent`. Not sure if we want to follow them.

jwendell (approved):
LGTM, made a minor comment, but I'm happy with any of the options.

tmp/build/patch-charts.sh
knrc: We should not be using latest tags, we need to be fixing them to a specific tag or a floating tag which we expect to be compatible.

tmp/build/patch-charts.sh
```diff
@@ -250,6 +250,22 @@ function patchTemplates() {
     sed -i -e 's/\/health/\/tmp\/health/' ${HELM_DIR}/istio/charts/sidecarInjectorWebhook/templates/deployment.yaml
   fi
 
+  # add oauth-proxy default hub, image, tag and imagePullPolicy
+  if [[ "${COMMUNITY,,}" == "true" ]]; then
+    sed -i -e '/^global:/a \
+\  oauthproxy:\
+\    hub: docker.io/openshift\
+\    image: oauth-proxy\
+\    tag: latest\
```
knrc: We need to find an alternative to the latest tag

knrc (changes requested):

tmp/build/patch-charts.sh
knrc: We should not be using the latest tag in any case, I've requested changes to that part

tmp/build/patch-charts.sh
jwendell: @dgn  There is `quay.io/openshift/origin-oauth-proxy:4.1` which is open to anyone.
Not sure if it's the same bits as on docker hub.
We have not tested it. You should probably test it before using.

dgn: @jwendell testing that image now

dgn:
use of `:latest` tag removed; I'm now pointing to `quay.io/openshift/origin-oauth-proxy:4.1`, this was recommended to me by @jwendell and confirmed in #forum-apiserver on CoreOS slack

knrc (approved):

tmp/build/patch-grafana.sh
```diff
@@ -9,8 +9,8 @@ function grafana_patch_deployment() {
   sed -i -e '/      containers:/ a\
           # OAuth proxy\
         - name: grafana-proxy\
-          image: openshift/oauth-proxy:latest\
-          imagePullPolicy: IfNotPresent\
+          image: {{ .Values.global.oauthproxy.hub }}/{{ .Values.global.oauthproxy.image }}:{{ .Values.global.oauthproxy.tag }}\
+          imagePullPolicy: {{ .Values.global.oauthproxy.imagePullPolicy }}\
```
rcernich, line 13: Is there a reason we're not using `.Values.global.imagePullPolicy`?

PR #187, opened 2019-08-06
MAISTRA-411: Updated alm-examples to include ServiceMeshMemberRoll

dmitri-d:
The latest version of operator-courier was flagging this as an issue.


manifests-maistra/1.0.0/maistra.v1.0.0.clusterserviceversion.yaml
```diff
@@ -83,7 +83,19 @@ metadata:
...
+        },
+        {
+          "apiVersion": "maistra.io/v1",
+          "kind": "ServiceMeshMemberRoll",
+          "metadata": {
+            "name": "default"
+          },
+          "spec": {
+            "members": [
+              "bookinfo"
```
brian-avery: Does it make sense to include bookinfo here? Should we include some sort of a comment saying "include the proper namespaces" here?

manifests-servicemesh/1.0.0/maistra.v1.0.0.clusterserviceversion.yaml
```diff
@@ -83,7 +83,19 @@ metadata:
...
+        },
+        {
+          "apiVersion": "maistra.io/v1",
+          "kind": "ServiceMeshMemberRoll",
+          "metadata": {
+            "name": "default"
+          },
+          "spec": {
+            "members": [
+              "bookinfo"
```
brian-avery: same question

tmp/build/manifest-templates/clusterserviceversion.yaml
```diff
@@ -83,7 +83,19 @@ metadata:
...
+        },
+        {
+          "apiVersion": "maistra.io/v1",
+          "kind": "ServiceMeshMemberRoll",
+          "metadata": {
+            "name": "default"
+          },
+          "spec": {
+            "members": [
+              "bookinfo"
```
brian-avery: and here

brian-avery (approved):

manifests-maistra/1.0.0/maistra.v1.0.0.clusterserviceversion.yaml
jwendell: Is the user going to copy and paste this? If yes, perhaps should we comment the bookinfo line?

Also, I'm guessing the namespace is omitted on purpose, correct?

dmitri-d: It's an example -- I figured including bookinfo is ok (or better than leaving the list empty). I think the idea is that it should be usable?

jwendell: If the user copies and paste this, and there's already a namespace called bookinfo in the cluster, then it might be a problem. Having it commented and also a comment `#Put the namespaces here blah` should work I think.

dmitri-d: Yes re: namespace. 

Here's what the docs say re: alm-examples: 
> You can provide templates for each of your CRDs with a minimum set of configuration as an annotation named alm-examples

I feel like this should be a complete minimum config example. I feel that "bookinfo" ns is pretty safe, as it's a commonly used example in k8s and Istio. Leaving the members list empty isn't a good example, imo. Adding a comment is ok, but not a great example either.

It is possible (if unlikely), that someone has production code deployed into "bookinfo" ns, if @jwendell  and @brian-avery feel this needs to be addressed, I can add a comment along the lines of:

```
Please add projects that are part of the mesh to the "members" field.
```

brian-avery: My understanding of the templates are interpreted as a starting point in operator marketplace. I.e. you say I'm going to deploy the istio-operator. Then the user clicks the CRDs that they'd like to load and these templates are loaded into an online editor, which the user then fills in with their details. It's been a little while since I've looked at this so I could be wrong on how this presents itself, but if it's to load into an online editor, then I think that it makes sense to have a comment saying your data here instead.

jwendell: You could leave bookinfo as it is and put a comment above like `Replace bookinfo below with the actual projects you want to be part of the mesh...`

dmitri-d: @brian-avery: you are indeed correct, these examples get loaded as a starting config.

dmitri-d: Turns out there isn't a valid way to add comments to json, other than create an empty key in a hash, e.g.:
```
         "spec": {
            "": "Replace bookinfo with the names of actual projects you want to include in the mesh",      
            "members": [
              "bookinfo" 
            ]
          }
```

I'd rather not do that, as this may lead to validation failures in the operator. I also tried C++ style comments which seem to be supported in some parsers, but alas these didn't pass the validation.

jwendell (approved):
LGTM

PR #188, opened 2019-08-07
MAISTRA-653: Update protobuf descriptors for generated 3scale API

philipgough:
For context, see https://github.com/3scale/3scale-istio-adapter/pull/111


helm/maistra-threescale/values.yaml
```diff
@@ -2,7 +2,7 @@ enabled: false
 
 hub: quay.io/3scale
 image: 3scale-istio-adapter
-tag: v0.4.1
+tag: v1.0.0
```
brian-avery: Technically this should be updated in patch-charts.sh, but #176 is changing this.

knrc (changes requested):
We will merge this separately from #176, however we need a maistra issue to cover this work and the version change moved to the patch scripts.

philipgough:
@knrc https://issues.jboss.org/browse/MAISTRA-653 covers it I think

knrc:
@PhilipGough Thanks, that does indeed cover it.  Can we have the PR updated to include the update for the patch scripts?

philipgough:
@knrc I've updated the version in the patch scripts as requested. However I'm not sure that value is not redundant now also. I've rebased on top of #176 and now the tags appear to be hardcoded in https://github.com/Maistra/istio-operator/blob/maistra-1.0/deploy/smcp-templates/servicemesh#L41 and https://github.com/Maistra/istio-operator/blob/maistra-1.0/deploy/smcp-templates/maistra#L43

Wondering if I can just remove the two tag references in this PR entirely at this point?

knrc:
@PhilipGough #716 wasn't merged when I made the comment and it was still under discussion, it looks as if it was merged yesterday afternoon.  Brian has indeed moved the versions to the new profile mechanism.

knrc (approved):

brian-avery (approved):

PR #189, opened 2019-08-07
MAISTRA-670 remove webhook configuration from *-operator.yaml files

rcernich:


jwendell:
@rcernich So, the user is going to be responsible for removing it after uninstall? That sounds awkward that we intentionally leave something behind. I'm assuming the re-installation will not fail because there's already an object with the same name installed?

README.md
```diff
@@ -94,11 +94,12 @@ $ oc delete -n istio-system -f ./deploy/examples/maistra_v1_servicemeshcontrolpl
 
 ### Uninstalling the Istio Operator
 
-If you followed the instructions above for installation, the operator can be uninstalled by simply issuing a delete
-command against the same resources.  For example:
+If you followed the instructions above for installation, the operator can be uninstalled by issuing the following delete
```
brian-avery, line 97: Can you please put this in the docs as well?

rcernich:
Not really awkward, just a bit messy.  The webhook is only associated with maistra.io resources, and uninstalling the operator removes those crds, which means the webhook would never get invoked anyway.  Even if those crds were left around, one could argue that we should prevent creation of maistra.io resources if the operator is not running.

Reinstalling the operator without removing the webhook first is not an issue as the operator creates the webhook programmatically, so whatever was there will be overwritten.

In other words, there are no side effects to leaving the webhook around.  Or you could look at it as it's worse to include it in the *-operator.yaml files than it is to not have it deleted.

jwendell:
@rcernich what about setting up a signal handler (either in the operator binary, or as a kubernetes pre-stop handler) to remove those things?

rcernich:
Too complicated for something that doesn't make a difference, plus you don't know why the pod is disappearing, e.g. could just be getting evicted, or updated to a new image, etc.

rcernich:
I suppose we could add an owner reference on the webhook to the crd resources, that way if the crds are removed, so is the webhook.  I don't know if this helps or not.  We'd have to manually update the webhook, as I don't think there's anything in the controller runtime api that allows you to do this.

jwendell (approved):

brian-avery (approved):

brian-avery (approved):

PR #192, opened 2019-08-08
MAISTRA-740 Make SMCP template paths configurable

luksa:


luksa:
Submitting this now, but it doesn't need to go into rc1.

brian-avery (approved):

rcernich (approved):

cmd/manager/main.go
```diff
@@ -36,6 +36,8 @@ func main() {
 
 	// ControlPlane
 	flag.StringVar(&common.ChartPath, "chartPath", "/etc/istio-operator/1.1.0/helm", "The location of the Helm charts.  The charts will be rendered using $chartPath/istio (similar layout to istio.io/istio/install/kubernetes/helm).")
+	flag.StringVar(&common.TemplatePath, "templatePath", "/usr/share/istio-operator/templates/", "The location of the ServiceMeshControlPlane templates.")
+	flag.StringVar(&common.DefaultTemplatePath, "defaultTemplatePath", "/usr/share/istio-operator/default-templates/", "The location of the default ServiceMeshControlPlane templates.")
```
jwendell, line 41: Let's make /usr/local/share the default dir. That's what istio uses and we use /usr/local as our prefix everywhere.

luksa: We should also move the chart path there. Can you file a new jira for this? We'll need to update the RPM spec.

rcernich: Shouldn't we use linux filesystem recommendations?  I consider ChartPath part of the application configuration, which is why I chose etc.  That said, keep in mind that if you change these, you will also need to change the rpm spec.

jwendell: If charts are configuration, `{sysconfdir}` is fine. If it's data, `{datadir}` should be used.

The issue is that on our RPM's, we are using `/usr/local` as our prefix, so, `{datadir}` becomes `/usr/local/share`, `{bindir}` becomes `/usr/local/bin`. That is intentional to keep in sync with upstream (for istio at least).

So, by default our RPM's will put things under `/usr/local`. We can decide whether or not use `/usr/local` as prefix, but that should be another issue, I think.

dgn (approved):

cmd/manager/main.go
dgn: Let's move that discussion into https://issues.jboss.org/browse/MAISTRA-750 - we will touch the default paths in that ticket anyway

knrc:
Closing as we will implement an alternative approach

PR #193, opened 2019-08-08
MAISTRA-726 Handle SMCP disappearing from cache during removal of finalizer

luksa:
If the SMCP is deleted, the reconciler deletes resources owned by the
SMCP. This requeues the same SMCP, causing the reconcile to be invoked
again immediately after it has completely deleted the SMCP (i.e. removed
its finalizer). Because the watch event caused by the deletion may
arrive during the invocation of reconcile(), the SMCP object may
disappear from the cache mid-way through. The update will fail with a
conflict, and the get() on the cache will then fail with a NotFound
error. If that happens, reconcile() can safely return without any errors.

If there's any other error during get(), there's no sense in calling
get() again immediately, as that will most likely result in the same
error. This will just result in a flood of errors in the log. Instead,
reconcile() now returns an error, properly requeueing the SMCP with
backoff.

### DIFF
--- a/pkg/controller/servicemesh/controlplane/controller.go
+++ b/pkg/controller/servicemesh/controlplane/controller.go
@@ -3,10 +3,12 @@ package controlplane
 import (
 	"context"
 	"fmt"
+
+	errors2 "github.com/pkg/errors"
+
 	v1 "github.com/maistra/istio-operator/pkg/apis/maistra/v1"
 	"github.com/maistra/istio-operator/pkg/bootstrap"
 	"github.com/maistra/istio-operator/pkg/controller/common"
-	errors2 "github.com/pkg/errors"
 
 	appsv1 "k8s.io/api/apps/v1"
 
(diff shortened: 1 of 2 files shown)


luksa:
I'll squash the commits before merging. Kept them separate for easier review.

rcernich (approved):

jwendell (approved):
Confirmed it doesn't show any errors when deleting the SMCP.

rcernich (approved):

pkg/controller/servicemesh/controlplane/controller.go
```diff
@@ -174,18 +176,22 @@ func (r *ReconcileControlPlane) Reconcile(request reconcile.Request) (reconcile.
 			return reconcile.Result{}, nil
 		}
 		reqLogger.Info("Deleting ServiceMeshControlPlane")
-		result, err := reconciler.Delete()
+		err := reconciler.Delete()
 		// XXX: for now, nuke the resources, regardless of errors
```
dgn, line 180: Shouldn't we return early and requeue if deleting fails? It seems wrong to me to try to remove the finalizer if there are still resources lurking about. Or do we explicitly rely on the GC in case of error?

dgn: (note that this was not changed in this PR so this comment is just a general question)

dgn (approved):
lgtm, cannot reproduce error

pkg/controller/servicemesh/controlplane/controller.go
luksa: Yup, I was thinking the same thing. I'll have a PR shortly.

PR #194, opened 2019-08-08
MAISTRA-730 Remove istio-galley-mesh cluster role and binding

luksa:


luksa:
Don't merge into rc1. While it doesn't look like Galley actually ever reads nodes, we need more testing to see if that's really the case.

rcernich (approved):

knrc (approved):

PR #197, opened 2019-08-12
MAISTRA-753 Add validation webhook port to operator manifest YAMLs

luksa:


knrc (approved):

jwendell (approved):
How was it working without this? This is just a hint to k8s, right?

luksa:
@jwendell k8s doesn't care about container ports. Adding them to a container definition is just information for other cluster _users_. Even if a port isn't specified in the container definition, anyone can still connect to it.

PR #198, opened 2019-08-12
MAISTRA-754 Add instructions on running Operator locally during devel.

luksa:


DEVEL.md
```diff
@@ -0,0 +1,68 @@
...
+Run the operator locally with the following command-line options:
+```bash
+--discoveryCacheDir /tmp/istio-operator/.kube/cache/discovery --chartPath tmp/_output/helm/istio-releases/istio-1.1.0/ istio-operator
+```
+
+### Modifying the ValidatingWebhookConfiguration 
+
+Because the Operator installs a `ValidatingWebhookConfiguration` that points to the Operator pod running inside the cluster, all CRUD operations on `ServiceMeshControlPlane` or `ServiceMeshMemberRoll` objects will fail. To fix this, you either need to remove the `ValidatingWebhookConfiguration`:
+```bash
+oc -n istio-operator delete validatingwebhookconfigurations istio-operator.servicemesh-resources.maistra.io
```
dgn: `-n istio-operator` shouldn't be necessary; validatingwebhookconfigurations are not namespaced

DEVEL.md
```diff
@@ -0,0 +1,68 @@
+# Developing the Maistra Istio Operator
+
```
dgn, line 2: We should probably add one or two sentences on how to build the operator

dgn (approved):
looks good, just two very minor things. My cluster is currently restarting, will try this out as soon as it's up. Until now I've always built an image and deployed that to my cluster

DEVEL.md
luksa: Ah, of course. Thanks!

DEVEL.md
luksa: Added.

DEVEL.md
```diff
@@ -0,0 +1,78 @@
...
+## Running the operator locally
+
+During development, you can run Istio Operator locally on your dev machine and have it reconcile objects in a local or remote OpenShift cluster. You can run the Operator from within your IDE and even use a debugger.
+
+To run the Operator locally, you need to set the following environment variables:
+
+- `KUBECONFIG=<path to your kubectl config file>`
+- `POD_NAME=istio-operator`
+- `POD_NAMESPACE=istio-operator`
```
dgn, line 23: This is ignored by the operator; instead it reads it from a file:
`/var/run/secrets/kubernetes.io/serviceaccount/namespace`

dgn:
We're also still missing information on how to setup the templates

DEVEL.md
luksa: Ah, I've been using an extracted default token secret in `/var/run/secrets/...`, but when writing the doc, I realized that I can probably just set `KUBECONFIG` instead. I thought I removed the `/var/run/secrets/...` directory, but apparently not. 

Anyway, I've now submitted https://github.com/Maistra/istio-operator/pull/200 to allow us to run it without the namespace file, so `POD_NAMESPACE` now has to be set when running locally.

luksa:
@dgn For the templates, I'm waiting for https://github.com/Maistra/istio-operator/pull/192 to be merged. Until then, I just copied the templates to `/usr/share/...`, where the operator expects them. I'll update the docs when we finalize #192

dgn:
> @dgn For the templates, I'm waiting for #192 to be merged. Until then, I just copied the templates to `/usr/share/...`, where the operator expects them. I'll update the docs when we finalize #192

Yes, makes sense

knrc (approved):

DEVEL.md
```diff
@@ -0,0 +1,78 @@
+# Developing the Maistra Istio Operator
+
+## Building the binary and the container image
+
+To build the Operator container image, run:
+```bash
+IMAGE=docker.io/maistra/istio-operator:latest tmp/build/docker_build.sh
```
jwendell: Can we use other image name? Like `docker.io/<your-username>/istio-operator:latest`

DEVEL.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+This builds the binary, downloads and patches the Istio Helm charts, and finally builds the container image. The build script doesn't push the image to the container registry, so you'll need to push it manually:
+```bash
+docker push docker.io/maistra/istio-operator:latest
```
jwendell: Same here

jwendell (changes requested):

DEVEL.md
luksa: I didn't want to use `<your username>`, but couldn't think of a good example username (best I could think of was `johndoe`), so I went with `maistra` - I thought it would be obvious people should replace this with their own usernames... but now I agree `<your username>` is perfectly fine.

Anyway, I've fixed it now.

luksa:
I've now also added a reference to DEVEL.md to README.md

DEVEL.md
```diff
@@ -0,0 +1,78 @@
...
+- `ISTIO_CNI_IMAGE_PULL_SECRET=my-pull-secret` (required if the registry hosting the CNI image is not public; secret must be in operator's namespace as defined above)
+
+Make sure you run `tmp/build/docker_build.sh` before running the Operator to generate the Helm charts in `tmp/_output/helm`.
+
+Before running the Operator, don't forget to stop the Operator running in the cluster:
+```bash
+oc -n istio-operator scale deployment istio-operator --replicas 0
+```  
+
+Run the operator locally with the following command-line options:
```
rcernich, line 34: Does `operator-sdk up local ...` work?  Is that a better approach?

luksa: I don't know about that one, but running the operator binary directly seems simpler. Plus, I usually just run it inside my IDE (in debug mode actually). Not sure how that would work through operator-sdk.

rcernich (approved):

DEVEL.md
```diff
@@ -0,0 +1,78 @@
...
+```bash
+--resourceDir tmp/_output/resources/ --discoveryCacheDir /tmp/istio-operator/.kube/cache/discovery istio-operator
+```
+
+### Modifying the ValidatingWebhookConfiguration 
+
+Because the Operator installs a `ValidatingWebhookConfiguration` that points to the Operator pod running inside the cluster, all CRUD operations on `ServiceMeshControlPlane` or `ServiceMeshMemberRoll` objects will fail. To fix this, you either need to remove the `ValidatingWebhookConfiguration`:
+```bash
+oc delete validatingwebhookconfigurations istio-operator.servicemesh-resources.maistra.io
```
jwendell, line 43: It gets recreated by the operator (via code).

So we get errors like
```
ERROR	kubebuilder.controller	controller/controller.go:209	Reconciler error	{"Controller": "servicemeshmemberroll-controller", "Request": "i1/default", "error": "Internal error occurred: failed calling webhook \"smmr.validation.maistra.io\": Post https://admission-controller.istio-operator.svc:443/validate-smmr?timeout=30s: no endpoints available for service \"admission-controller\""}
```

luksa: Yes, I think I mention that you need to delete it every time you run the operator.

jwendell: Ah ok, gotcha. I missed an **after** in the sentence ... every time you restart the Operator...

DEVEL.md
```diff
@@ -0,0 +1,78 @@
...
+
+To run the Operator locally, you need to set the following environment variables:
+
+- `KUBECONFIG=<path to your kubectl config file>`
+- `POD_NAME=istio-operator`
+- `POD_NAMESPACE=istio-operator`
+- `ISTIO_CNI_IMAGE=some-registry/istio-cni-image:latest`
+- `ISTIO_CNI_IMAGE_PULL_SECRET=my-pull-secret` (required if the registry hosting the CNI image is not public; secret must be in operator's namespace as defined above)
+
+Make sure you run `tmp/build/docker_build.sh` before running the Operator to generate the Helm charts in `tmp/_output/helm`.
```
jwendell: What about

Make sure you run `tmp/build/docker_build.sh` before running the Operator to generate the required resources in `tmp/_output/resources`.

DEVEL.md
```diff
@@ -0,0 +1,78 @@
...
+
+Before running the Operator, don't forget to stop the Operator running in the cluster:
+```bash
+oc -n istio-operator scale deployment istio-operator --replicas 0
+```  
+
+Run the operator locally with the following command-line options:
+```bash
+--resourceDir tmp/_output/resources/ --discoveryCacheDir /tmp/istio-operator/.kube/cache/discovery istio-operator
```
jwendell: The last argument is not necessary.

DEVEL.md
```diff
@@ -0,0 +1,78 @@
...
+or point it to the Operator process running on your local machine by patching the `Service` and `Endpoints` objects:
+```bash
+oc -n istio-operator patch svc admission-controller --type=json -p='[{"op": "remove", "path": "/spec/selector"}]'
+oc -n istio-operator patch ep admission-controller --type=json -p='[{"op": "add","path": "/subsets","value": [{"addresses": [{"ip": "192.168.1.100"}],"ports": [{"port": 11999}]}]}]'
+```
+
+Replace `192.168.1.100` with the IP of your dev machine, where the Operator is running. You also need to ensure that the OpenShift master node can connect to your dev machine on port 11999. Note: this is trivial if running OpenShift through DIND.
+
+ 
+NOTE: you'll need to remove the `ValidatingWebhookConfiguration` or patch the `Service` and `Endpoints` every time you restart the Operator, as it resets them during startup.
```
jwendell: What about putting the word `after` e.g.:

NOTE: You need to remove the `ValidatingWebhookConfiguration` or patch the `Service` and `Endpoints` every time after you restart the Operator, as it is created automatically during startup.

DEVEL.md
luksa: done.

DEVEL.md
luksa: oh, thanks. I've had this in since forever. I think it was necessary before.

DEVEL.md
luksa: done.

jwendell (approved):

PR #199, opened 2019-08-12
MAISTRA-746 Set maxUnavailable to 0 for grafana, prometheus deployments

dgn:
Istio 1.1 is still using the extensions/v1beta1 API definition of deployments, which defaults maxUnavailable to 1. This can lead to the operator assuming the deployments are available, when in fact they are not, e.g. due to the oauth-proxy not being able to start. This commit injects a maxUnavailable value of 0 into grafana & prometheus deployments.

Note that all other deployments already specify `maxUnavailable: 0`


tmp/build/patch-grafana.sh
```diff
@@ -73,6 +73,14 @@ function grafana_patch_deployment() {
   ${HELM_DIR}/istio/charts/grafana/templates/deployment.yaml
 
   sed -i -e '/securityContext/,/fsGroup/d' ${HELM_DIR}/istio/charts/grafana/templates/deployment.yaml
+
+  # Fix for MAISTRA-746, can be removed when we move to Istio-1.2
+  grep -q 'maxUnavailable' ${HELM_DIR}/istio/charts/grafana/templates/deployment.yaml || \
```
jwendell: So, can we get rid of this check as discussed?

tmp/build/patch-prometheus.sh
```diff
@@ -62,6 +62,14 @@ function prometheus_patch_deployment() {
 {{- end }}/' ${HELM_DIR}/istio/charts/prometheus/templates/deployment.yaml
 
 	sed -i "/storage.tsdb.retention.*/a\ \ \ \ \ \ \ \ \ \ \ \ - \'--storage.tsdb.path=/prometheus\'" ${HELM_DIR}/istio/charts/prometheus/templates/deployment.yaml
+
+  # Fix for MAISTRA-746, can be removed when we move to Istio-1.2
+  grep -q 'maxUnavailable' ${HELM_DIR}/istio/charts/grafana/templates/deployment.yaml || \
```
jwendell: So, can we get rid of this check as discussed?

jwendell (changes requested):

tmp/build/patch-grafana.sh
```diff
@@ -73,6 +73,14 @@ function grafana_patch_deployment() {
...
+  sed -i -e '/^spec:/ a\
+  strategy:\
+    rollingUpdate:\
+      maxSurge: 1\
+      maxUnavailable: 0' ${HELM_DIR}/istio/charts/grafana/templates/deployment.yaml
```
luksa: If someone scales the deployment up to e.g. 3, then maxUnvailable being set to 0 is no longer sensible. I wanted to suggest using a percentage instead, but the absolute value is rounded _down_ when it gets calculated from the percentage, so it could again be 0. This makes me think if we should be fixing this elsewhere.

dgn: Not sure I understand, because I don't think it's problematic for the deployment to be considered not ready if some of those replicas don't come up. Either way, AFAIK neither prometheus nor grafana are configured in a way to support that scenario. They're single-instance only.

rcernich (changes requested):
This needs to be done for gateways as well.

tmp/build/patch-grafana.sh
rcernich: Given what @luksa said, `maxUnavailable: 1%` would be fine, as this would mark the deployment as available so long as 1/100 replicas are ready.  For what we want, we just need to ensure that at least one pod is ready before moving on.

The other option is to check the replica count instead of the `Available` condition on the Deployment resources.

tmp/build/patch-grafana.sh
dgn: if you really want to, sure

dgn:
> This needs to be done for gateways as well.

You're right; I thought this was in there already, must have been looking in the wrong place. Thanks!

Just noticed it's also missing for IOR. Will add there, too.

tmp/build/patch-grafana.sh
dgn: I'm still not sure about this. As long as we're running a single replica, we definitely want `maxUnavailable` to be `0` at all times, there's nothing bad about it because that means there's no downtime during rolling upgrades. For `replicas >= 2`, it's okay to be `1`. I agree however that we should be using percentages instead of absolutes because it slows down upgrades for large replicasets (in this current config, kubernetes can only ever create one new replica, wait until ready, then shoot down an old one).

dgn: How about we just go with today's defaults in kubernetes?
```
      maxSurge: 25%
      maxUnavailable: 25%
```

for `replicas: 1` this will actually be the same as above, but for larger replicaSets it will adjust.

luksa: Yeah, the second part of my comment didn't make sense. We actually _want_ maxUnavailable to be zero, so rounding down is exactly what we want. As for the fact that these deployments are single-instance - that doesn't really matter, as we need to figure this out for any type of deployment - even horizontally scalable.

Setting maxUnavailable to say 25% would mean that maxUnavailable=0 for deployments with less than 4 replicas. If we focus only for replicas=1, which is currently the case for these deployments, maxUnavailable of 25% would be zero, which means the operator will treat the deployment as ready only when the replica is ready - and that's exactly what we want.

For replicas=4, the deployment would be considered ready when 3 replicas are ready. That's ok, but probably not really necessary. A single ready replica would probably suffice? But then again, maxUnavailable will eventually be configurable by the mesh admin, as it will be different for every environment. It makes sense for the operator to take maxUnavailable into account when determining the ready status of a deployment instead of just requiring 1 replica to be ready (1 ready replica maybe isn't enough in some clusters). 

So, I think we should keep the current method of checking the `Available` condition, but ensure that for replicas=1, maxUnavailable becomes zero. And we do that by setting maxUnavailable to a percentage.

luksa: +1 on using k8s defaults.

rcernich (approved):

jwendell (approved):

luksa (approved):

brian-avery (approved):

PR #200, opened 2019-08-12
MAISTRA-755 Obtain operator namespace from POD_NAMESPACE env var

luksa:


dgn (approved):

jwendell (commented):
Two questions:
1. Does it make sense to also update YAML files to include that variable by default?
1. Does this need to get into RC1?

luksa:
@jwendell No and No. This is just for development, so it doesn't need to be in RC1 and also doesn't need to be in YAMLs, as it's only used when running it locally (where you don't even use YAML to run it).

jwendell (approved):
Thanks for this.

rcernich (approved):

PR #201, opened 2019-08-12
MAISTRA-706 Remove default values for jaeger/elasticsearch hub & tag

dgn:
The Jaeger operator should decide which image to deploy and manage. Users can still override it by setting hub/tag values, but we default to not setting them and if they're not set, we will not specify an image in the Jaeger CR.


brian-avery:
If we're going to do this for Jaeger, we should also do it for Kiali

brian-avery:
Are we coupled to the wrong part of Jaeger? Is it a problem if, say, the Jaeger version is significantly older? If it is, should we check the operator version?

JStickler:
How is this making it easier for the user to install?  How are they supposed to know which versions are compatible with each other?  If there is a default, the user can still change it, but if there is NO value at all, you've now forced them to go do some homework before they can install.

dgn:
> If we're going to do this for Jaeger, we should also do it for Kiali

Probably, but this was specifically requested for Jaeger, see the issue: https://issues.jboss.org/projects/MAISTRA/issues/MAISTRA-706

dgn:
> How is this making it easier for the user to install? How are they supposed to know which versions are compatible with each other? If there is a default, the user can still change it, but if there is NO value at all, you've now forced them to go do some homework before they can install.

We're basically moving the decision about which version to use into the Jaeger operator, and the operator will choose a version that it can manage. We're specifying version 1.13 of the operator in our installation requirements, so that will also be the image version that the operator installs (1.13.x at least). Yes, if you installed the wrong operator version or the community operator instead of product, you might end up with a different image. But then, using the wrong operator for the right image might also cause problems.

dgn:
I'm actually not sure how the dependency resolution would work when using OLM to install the dependencies, @dmitri-d maybe you can comment on how this would work in an OLM context?

brian-avery (approved):

helm/istio/charts/tracing/templates/jaeger-production-elasticsearch.yaml
```diff
@@ -12,7 +12,9 @@ spec:
   strategy: production
 
   query:
+    {{- if and .Values.jaeger.hub .Values.jaeger.tag }}
     image: {{ .Values.jaeger.hub }}/jaeger-query:{{ .Values.jaeger.tag }}
```
knrc, line 16: The image name should be jaeger-query-rhel7, @objectiser can you please confirm?

helm/istio/charts/tracing/templates/jaeger-production-elasticsearch.yaml
```diff
@@ -22,7 +24,9 @@ spec:
       {{- end }}
 
   collector:
+    {{- if and .Values.jaeger.hub .Values.jaeger.tag }}
     image: {{ .Values.jaeger.hub }}/jaeger-collector:{{ .Values.jaeger.tag }}
```
knrc, line 28: The image name should be jaeger-collector-rhel7, @objectiser can you please confirm?

helm/istio/charts/tracing/templates/jaeger-all-in-one.yaml
```diff
@@ -12,7 +12,9 @@ spec:
   strategy: allInOne
 
   allInOne:
+    {{- if and .Values.jaeger.hub .Values.jaeger.tag }}
     image: {{ .Values.jaeger.hub }}/all-in-one:{{ .Values.jaeger.tag }}
```
knrc, line 16: The image name should be jaeger-all-in-one-rhel7, @objectiser can you please confirm?

knrc (changes requested):

helm/istio/charts/tracing/templates/jaeger-all-in-one.yaml
knrc: I'm moving this to another issue

knrc (approved):

PR #202, opened 2019-08-12
[MAISTRA-750] Unify templates and charts dir

jwendell:
And make /usr/local/share/istio-operator the default location

### DIFF
--- a/tmp/build/Dockerfile
+++ b/tmp/build/Dockerfile
@@ -3,6 +3,6 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
 ARG build_type=maistra
 
 ADD tmp/_output/bin/istio-operator /usr/local/bin/istio-operator
-ADD tmp/_output/templates/ /usr/share/istio-operator/default-templates
+ADD tmp/_output/resources/default-templates/ /usr/local/share/istio-operator/default-templates
 ADD manifests-${build_type} /manifests
-COPY tmp/_output/helm/istio-releases/istio-1.1.0 /etc/istio-operator/1.1.0/helm
+COPY tmp/_output/helm/istio-releases/istio-1.1.0 /usr/local/share/istio-operator/helm/1.1.0
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -33,9 +33,7 @@ var discoveryCacheDir string
 
 func main() {
 	flag.StringVar(&discoveryCacheDir, "discoveryCacheDir", "/home/istio-operator/.kube/cache/discovery", "The location where cached discovery information used by the REST client is stored.")
-
-	// ControlPlane
-	flag.StringVar(&common.ChartPath, "chartPath", "/etc/istio-operator/1.1.0/helm", "The location of the Helm charts.  The charts will be rendered using $chartPath/istio (similar layout to istio.io/istio/install/kubernetes/helm).")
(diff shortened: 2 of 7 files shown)


tmp/build/download-charts.sh
```diff
@@ -37,13 +37,11 @@ function retrieveIstioRelease() {
...
-      echo "downloading Istio Release: ${ISTIO_URL}"
-      cd "${RELEASES_DIR}"
-      curl -LO "${ISTIO_URL}"
-    )
-  fi
+  (
+    echo "downloading Istio Release: ${ISTIO_URL}"
+    cd "${RELEASES_DIR}"
+    curl -LO "${ISTIO_URL}"
+  )
```
knrc: It looks like this change snuck into the PR, it's not related to the task at hand

knrc (changes requested):

tmp/build/Dockerfile
```diff
@@ -3,6 +3,6 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
 ARG build_type=maistra
 
 ADD tmp/_output/bin/istio-operator /usr/local/bin/istio-operator
-ADD tmp/_output/templates/ /usr/share/istio-operator/default-templates
+ADD tmp/_output/templates/ /usr/local/share/istio-operator/default-templates
 ADD manifests-${build_type} /manifests
-COPY tmp/_output/helm/istio-releases/istio-1.1.0 /etc/istio-operator/1.1.0/helm
+COPY tmp/_output/helm/istio-releases/istio-1.1.0 /usr/local/share/istio-operator/helm
```
luksa: I think we need to keep the version in the directory name, as we'll eventually have multiple versions in there.

luksa (commented):
Can we also reorganize the paths in `tmp/_output` so that it contains a `resources` directory with all required subdirectories (`templates`, `default-templates` and `helm`). This will allow us to run the operator locally with `--resourceDir tmp/_output/resources/`.

tmp/build/Dockerfile
jwendell: @luksa  Do you think we should do this now or can we leave that for the future, when we proper implement it?

tmp/build/download-charts.sh
jwendell: heh, I had this change locally and got caught by git commit -a :/
Fixed, thanks.

jwendell:
> Can we also reorganize the paths in `tmp/_output` so that it contains a `resources` directory with all required subdirectories (`templates`, `default-templates` and `helm`). This will allow us to run the operator locally with `--resourceDir tmp/_output/resources/`.

Done. I wanted to make this as a not intrusive change (not touching download-charts), so, I'm just copying the files to the right location after patch-charts is run.

tmp/build/Dockerfile
luksa: Since we had the version in so far, I'd say it's best to keep it.

jwendell:
@luksa ok, I went with the version being a subdir of helm dir:

```
resources
resources/default-templates
resources/helm
resources/helm/1.1.0
```

Right now the version 1.1.0 is hard coded (that has not changed from what we have now) everywhere until MAISTRA-766 is implemented.

jwendell:
I've also updated https://github.com/Maistra/rpms/pull/65 to reflect this change

knrc (approved):

luksa:
@jwendell Don't we need to version the templates too? So they should be under 1.1.0, right?

luksa (approved):

PR #204, opened 2019-08-13
MAISTRA-531: Add multitenancy support to prometheus

knrc:


knrc:
linked to https://github.com/Maistra/prometheus/pull/2

brian-avery (approved):

dgn (approved):

rcernich:
Do we also need to remove the clusterrole.yaml and clusterrolebinding.yaml currently used by prometheus?  Does clusterrolebinding need to be converted to simple rolebinding?

https://github.com/Maistra/istio/blob/maistra-1.0/install/kubernetes/helm/istio/charts/prometheus/templates/clusterrole.yaml

knrc:
@rcernich The role binding is currently converted by the script, we could certainly update it and just make it a plain role binding without needing the script.  We can tidy both up in another JIRA.

knrc:
@rcernich on second thoughts those are taken from the distribution so the script should be the correct way to alter the content

rcernich:
@knrc, sorry I missed that.  I don't know what I was looking at.  You are correct, it is a RoleBinding and it only has perms for services, endpoints, pods, and configmaps.  Ignore my previous comment.

rcernich (approved):

PR #207, opened 2019-08-13
MAISTRA-742 do not spam api server with status updates

rcernich:

### DIFF
--- a/pkg/controller/servicemesh/controlplane/manifestprocessing.go
+++ b/pkg/controller/servicemesh/controlplane/manifestprocessing.go
@@ -1,6 +1,8 @@
 package controlplane
 
 import (
+	"fmt"
+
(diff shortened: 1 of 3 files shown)


rcernich:
I can't guarantee that this "fixes" the issues with the status update errors, but I haven't seen any after applying these changes.  I think the main issue was caused by updating the status when reconciling components that didn't have any manifests, which caused status updates with essentially no pause between them.

I also updated the smmr controller to ignore smcp events if the smcp resource is not in a "clean" state.  This takes some of the load off the smmr controller loop and removes a lot of those, "waiting for initial ServiceMeshControlPlane install to complete," messages.

brian-avery (approved):

jwendell (approved):

PR #212, opened 2019-08-14
MAISTRA-765: Update kiali version to 1.0.5

knrc:


brian-avery (approved):

helm/istio/charts/kiali/templates/kiali-cr.yaml
```diff
@@ -8,29 +8,47 @@ spec:
   istio_namespace: "{{ .Release.Namespace }}"
 
   auth:
+{{- if .Values.dashboard }}
```
brian-avery, line 11: Does helm let us do:
{{- if and (.Values.dashboard (not (and (.Values.dashboard.user) (.Values.dashboard.passphrase)))) }}

In other words, do we need else strategy: login else strategy: login?

jwendell (approved):

PR #217, opened 2019-08-19
MAISTRA-537 Remove access to node objects from istio-cni cluster role

luksa:


jwendell (approved):

brian-avery (approved):

rcernich (approved):
This is a low priority, as the cni service account already has access to elevated SCC.

PR #218, opened 2019-08-19
MAISTRA-792 Fix error handling in UpdateReadiness / PostStatus

luksa:


pkg/controller/servicemesh/controlplane/readiness.go
```diff
@@ -1,39 +1,36 @@
...
 	r.Log.Info("Updating ServiceMeshControlPlane readiness state")
 	notReadyState, err := r.calculateNotReadyState()
 	if err != nil {
-		condition := v1.Condition{
-			Type:    v1.ConditionTypeReady,
-			Status:  v1.ConditionStatusUnknown,
-			Message: fmt.Sprintf("error collecting ready state: %s", err),
-		}
-		r.Status.SetCondition(condition)
-		r.Manager.GetRecorder(controllerName).Event(r.Instance, corev1.EventTypeWarning, eventReasonNotReady, condition.Message)
```
rcernich, line 36: Shouldn't we be updating the state and recording an event for this case?  These are the only ways a regular user has to know something went wrong.

luksa: Yeah, I was wondering about this. But nowhere else do we record an event when an actual error happens - not even where we call the exact same function as here (`calculateNotReadyState`). The function returns an error only if it's unable to read from the cache, so it's unlikely to ever happen. 

Having said that, we should probably record an event any time there's an actual error during reconciliation so that users know why it's not progressing. We could probably handle that in a single place (e.g. in `defer`). We can do that in a follow-up PR. However, the majority of those errors will be caused by the operator not being able to reach the API server. Recording an event in that case wouldn't even work, so it doesn't make much sense to record it. I've checked the code of the core k8s controllers, and none of them actually record events in those cases. The only time they record a warning event is when the cause of the error is somewhere else (e.g. pod container crashing).

rcernich: So there will be error text coming out of `postReconciliationStatus()`, including an event.  My concern here is that this code is run on periodic reconcile events (i.e. those not related to an actual resource change) and is intended to provide continuous status of the "ready" state.  For example, if you delete the galley pod, you'll see readiness go to false until a new pod spins up.

luksa: Those events and state changes are still recorded properly. The only event & status change I'm removing is when there's an actual internal error in `calculateNotReadyState()`, which can only originate from `FetchOwnedResources()`, more specifically from `kubeclient.List()`.

rcernich: Right, which means we can't determine the readiness state, hence `Unknown`.

luksa: OK, I'm convinced. I've reverted this part of the change.

rcernich (approved):

luksa:
@rcernich FYI I've added another small fix: https://github.com/Maistra/istio-operator/pull/218/commits/0a3b69be9cf8748113e05307e78bc3e6d6983767#diff-343aae33bd11703b4416cbb2d9342f3aR78-R83

rcernich (approved):

luksa:
@rcernich @brian-avery Conflict resolved.

brian-avery (approved):

PR #219, opened 2019-08-19
MAISTRA-794 Fix inconsistencies in operator event messages

luksa:


pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -222,11 +222,11 @@ func (r *ControlPlaneReconciler) Reconcile() (result reconcile.Result, err error
 
 	// delete unseen components
 	reconciliationMessage = "Pruning obsolete resources"
-	r.Manager.GetRecorder(controllerName).Event(r.Instance, corev1.EventTypeNormal, eventReasonPruningObsoleteResources, reconciliationMessage)
+	r.Manager.GetRecorder(controllerName).Event(r.Instance, corev1.EventTypeNormal, eventReasonPruning, reconciliationMessage)
 	r.Log.Info(reconciliationMessage)
 	err = r.prune(r.Instance.GetGeneration())
 	if err != nil {
-		reconciliationMessage = "unexpected error pruning obsolete resources"
+		reconciliationMessage = "Unexpected error pruning obsolete resources"
```
brian-avery: A couple of comments here:
* can we remove the word unexpected? Not sure if we have expected errors, but the unexpected are certainly more common. 
* according to Go style, error messages should not begin with a capital or end with a period.

luksa: I'll remove the word, but since this is a k8s event message (not an error message) it should start with a capital letter.

rcernich (approved):

rcernich (approved):

pkg/controller/servicemesh/controlplane/reconciler.go
luksa: I've removed "unexpected" from all error and event messages. Capitalization is actually a bit tricky, as some messages are used both in events and as logged error messages. I feel event messages are more visible than error messages, so having them capitalized seems like the better option (short of having duplicated messages).

brian-avery: Sounds good to me. I agree. Better to have good capitalization in the events than logs. I think this was more designed with the idea that errors will become part of other error messages as they filter out.

brian-avery (approved):

dgn (approved):

PR #230, opened 2019-08-22
MAISTRA-819 Fix SMCP status conditions

luksa:
To see how the events & conditions appear after applying these changes, see comment https://issues.jboss.org/browse/MAISTRA-819?focusedCommentId=13774056&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13774056


pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -465,8 +481,10 @@ func (r *ControlPlaneReconciler) postReconciliationStatus(reconciliationMessage
 	}
 	reconciledCondition := r.Status.GetCondition(v1.ConditionTypeReconciled)
 	if processingErr == nil {
+		reconciledCondition.Reason = reconciliationReason
```
brian-avery: Since this is true regardless of which condition we hit, should it be external to the if?

brian-avery (approved):

pkg/controller/servicemesh/controlplane/reconciler.go
luksa: Of course. Thanks! Fixed.

pkg/controller/servicemesh/controlplane/controller.go
```diff
@@ -185,9 +185,36 @@ func (r *ReconcileControlPlane) Reconcile(request reconcile.Request) (reconcile.
...
+
 		reqLogger.Info("Deleting ServiceMeshControlPlane")
 		err := reconciler.Delete()
 		if err != nil {
+			reconciler.Status.SetCondition(v1.Condition{
+				Type:    v1.ConditionTypeReconciled,
+				Status:  v1.ConditionStatusFalse,
+				Reason:  v1.ConditionReasonDeletionError,
+				Message: fmt.Sprintf("Error deleting service mesh: %s", err),
+			})
```
dgn, line 218: Isn't this missing a `reconciler.PostStatus()`?

luksa: Great catch! I've fixed it now. And tested it by manually inserting an error. Thanks!

dgn (approved):

rcernich (approved):

PR #234, opened 2019-08-22
MAISTRA-836 enabled tracing in alm-examples

dmitri-d:


manifests-maistra/1.0.0/maistraoperator.v1.0.0.clusterserviceversion.yaml
```diff
@@ -69,15 +69,15 @@ metadata:
...
               },
               "grafana": {
-                "enabled": false
+                "enabled": true
               },
               "tracing": {
                 "enabled": true,
                 "jaeger": {
-                  "template": "all-in-one"
+                  "template": "production-elasticsearch"
```
brian-avery: Is this the appropriate value for Maistra? Especially without an elastic search version?

knrc: This would match the profile defaults since it is set in base.  It is probably better to have the all-in-one profile for maistra and production-elasticsearch for servicemesh but this would also need a change to the profiles.

tmp/build/manifest-templates/clusterserviceversion.yaml
```diff
@@ -69,15 +69,15 @@ metadata:
...
               "grafana": {
-                "enabled": false
+                "enabled": true
               },
               "tracing": {
                 "enabled": true,
                 "jaeger": {
-                  "template": "all-in-one"
+                  "template": "production-elasticsearch"
                 }
```
knrc, line 81: This seems to be a build artifact and should not be checked into git, can we also clean this up and set up .gitignore?

knrc (changes requested):

tmp/build/manifest-templates/clusterserviceversion.yaml
knrc: Never mind, I had assumed this was being created from the ones above but it looks as if this is a template for build and the other ones are ignored

manifests-maistra/1.0.0/maistraoperator.v1.0.0.clusterserviceversion.yaml
knrc: I'll leave my review as requesting changes until we get an answer to this one

dmitri-d: jaeger.template in the maistra csv is now being set to "all-in-one"

dmitri-d:
* rebased
* contents of jaeger.template is now being set to "production-elasticsearch" in the servicemesh csv and "all-in-one" for the maistra one.

knrc (approved):

jwendell (approved):

PR #237, opened 2019-08-23
MAISTRA-842 only update smcp status when condition status changes

rcernich:

### DIFF
--- a/pkg/controller/servicemesh/controlplane/reconciler.go
+++ b/pkg/controller/servicemesh/controlplane/reconciler.go
@@ -72,6 +72,12 @@ const (
 )
 
 func (r *ControlPlaneReconciler) Reconcile() (result reconcile.Result, err error) {
+	if r.Status.GetCondition(v1.ConditionTypeReconciled).Status != v1.ConditionStatusFalse { 
+		r.initializeReconcileStatus()
+		err := r.PostStatus()
+		return reconcile.Result{}, err // ensure that the new reconcile status is posted immediately. Reconciliation will resume when the status update comes back into the operator
+	}
+
 	var ready bool
 	// make sure status gets updated on exit
 	reconciledCondition := r.Status.GetCondition(v1.ConditionTypeReconciled)
@@ -79,6 +85,7 @@ func (r *ControlPlaneReconciler) Reconcile() (result reconcile.Result, err error
 	reconciliationReason := reconciledCondition.Reason
 	reconciliationComplete := false
 	defer func() {
+		// this ensures we're updating status (if necessary) and recording events on exit
 		if statusErr := r.postReconciliationStatus(reconciliationReason, reconciliationMessage, err); statusErr != nil {
 			if err == nil {
 				err = statusErr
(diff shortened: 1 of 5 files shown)


pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -490,13 +502,30 @@ func (r *ControlPlaneReconciler) postReconciliationStatus(reconciliationReason v
...
 	} else {
-		reconciledCondition.Message = fmt.Sprintf("%s: error: %s", reconciliationMessage, processingErr)
-		r.Manager.GetRecorder(controllerName).Event(r.Instance, corev1.EventTypeWarning, reason, reconciliationMessage)
+		// grab the cause, as it's likely the error includes the reconciliation message
+		reconciledCondition.Message = fmt.Sprintf("%s: error: %s", reconciliationMessage, errors.Cause(processingErr))
+		r.Manager.GetRecorder(controllerName).Event(r.Instance, corev1.EventTypeWarning, reason, reconciledCondition.Message)
 	}
 	r.Status.SetCondition(reconciledCondition)
+
+	// we should only post status updates if condition status as changed
```
luksa: as -> has

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -72,13 +72,20 @@ const (
 )
 
 func (r *ControlPlaneReconciler) Reconcile() (result reconcile.Result, err error) {
+	if r.Status.GetCondition(v1.ConditionTypeReconciled).Status != v1.ConditionStatusFalse { 
+		r.initializeReconcileStatus()
+		r.PostStatus()
```
luksa: unhandled error

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -490,13 +502,30 @@ func (r *ControlPlaneReconciler) postReconciliationStatus(reconciliationReason v
...
+	if r.skipStatusUpdate() {
+		return nil
+	}
+
 	return r.PostStatus()
 }
 
+func (r *ControlPlaneReconciler) skipStatusUpdate() bool {
+	for _, conditionType := range []v1.ConditionType{v1.ConditionTypeInstalled, v1.ConditionTypeReconciled, v1.ConditionTypeReady} {
+		// TODO: what if the status is the same, but the reason is different? - we should also compare the reason
```
luksa: I think only checking the status is correct. The reason only applies when the status actually transitions from one state to the other. We should never update the status if the state doesn't change, regardless if the underlying reason changes.

luksa (changes requested):

pkg/controller/servicemesh/controlplane/reconciler.go
rcernich: removed

pkg/controller/servicemesh/controlplane/reconciler.go
rcernich: returning the error now

pkg/controller/servicemesh/controlplane/reconciler.go
rcernich: updated

luksa (approved):
LGTM. Events and conditions look perfect now.

knrc (approved):

PR #238, opened 2019-08-23
[MAISTRA-840] Remove the login strategy from Kiali auth CR

jwendell:
The only supported authentication strategy is now "openshift".
This aligns with what we have for other components already, like
Prometheus, Grafana and Jaeger.


deploy/examples/maistra_v1_servicemeshcontrolplane_cr_auth.yaml
```diff
@@ -7,6 +7,9 @@ spec:
   # element too.  An empty element is interpreted as null and will override all
   # default values (i.e. no values will be specified for that element, not even
   # the defaults baked into the chart values.yaml).
+
+  # For a complete list of all options, see the docs at https://maistra.io/docs/getting_started/custom-install/
```
brian-avery, line 11: Can you please update docs as well?

brian-avery: Other than that, lgtm

jwendell: It will be part of maistra-787

brian-avery (approved):
Content lgtm. This should be merged alongside the PR for maistra-787

helm/istio/charts/kiali/templates/kiali-cr.yaml
```diff
@@ -8,14 +8,10 @@ spec:
   istio_namespace: "{{ .Release.Namespace }}"
 
   auth:
-{{- if .Values.dashboard }}
```
rcernich, line 11: The only change here should be adding `.Values.dashboard.secretName` to the `and` statement below.  If not all of those are set, we should be using oauth (`openshift`) by default.

rcernich (changes requested):

rcernich (approved):

knrc (approved):

PR #243, opened 2019-08-27
MAISTRA-854: Revert "MAISTRA-764: Update elasticsearch resources for running in restricted…"

rcernich:
Reverts Maistra/istio-operator#241

### DIFF
--- a/helm/istio/charts/tracing/templates/jaeger-production-elasticsearch.yaml
+++ b/helm/istio/charts/tracing/templates/jaeger-production-elasticsearch.yaml
@@ -63,7 +63,7 @@ spec:
 {{- if .Values.jaeger.elasticsearch.nodeCount }}
       nodeCount: {{ toYaml .Values.jaeger.elasticsearch.nodeCount }}
 {{- else }}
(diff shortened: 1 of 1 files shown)


brian-avery (approved):

jwendell (approved):

knrc:
MAISTRA-764 should not have been reverted, we now have a configuration which *does not* install.

brian-avery:
Hey Kev,

Just saw your comment on this. This decision was made in the program call
after a discussion with everyone. The resulting decision was that it was
better to have a config that doesn't install in a default OpenShift install
that is production ready than one that doesn't install that is not
production ready. The elasticsearch defaults set before this PR was
reverted do not start.

knrc:
@brian-avery The default values used previously was starting on the default openshift installation, albeit after about 15 seconds or so, at least when I tested it.  Also the elastic search documentation for minimum configuration is https://docs.openshift.com/container-platform/4.1/logging/config/efk-logging-elasticsearch.html#efk-logging-elasticsearch-limits_efk-logging-elasticsearch

PR #246, opened 2019-09-03
[MAISTRA-885] Fix the links to the installation options

jwendell:


brian-avery:
Outside the scope of this PR, but I wonder if these files should be removed. Since we're not supporting install outside of OLM (which provides its own example) and the documentation should provide enough documentation, I wonder if it makes sense to start moving away from this deployment examples directory. Something to think about.

brian-avery (approved):

rcernich:
OLM just governs the operator installation, not control plane deployment.

brian-avery:
OLM includes examples for the control plane as do both product and community documentation. I was just asking if we still feel like it's worth it to have a third source of examples. That being said, we could also script both maistra.io and OLM to pull from one of these examples, so this could become authoritative. This directory is hidden (not listed) for users though.

rcernich:
OLM only provides support for a single example for each CRD owned by the operator.  I don't believe you can have multiple examples.  That said, I don't know how OLM uses these examples, just that it documents that there should be one per CRD.

knrc (approved):

PR #247, opened 2019-09-04
MAISTRA-889: switch default jaeger configuration over to all-in-one

knrc:
This is the PR representing the operator changes, rather than rebuild the RPMs we will be updating the container build to override the files within the container.

### DIFF
--- a/deploy/smcp-templates/servicemesh
+++ b/deploy/smcp-templates/servicemesh
@@ -16,7 +16,7 @@ spec:
       image: grafana-rhel8
     tracing:
       jaeger:
-        template: production-elasticsearch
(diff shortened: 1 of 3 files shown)


brian-avery (approved):

deploy/smcp-templates/servicemesh
```diff
@@ -16,7 +16,7 @@ spec:
       image: grafana-rhel8
     tracing:
       jaeger:
-        template: production-elasticsearch
+        template: all-in-one
```
brian-avery, line 19: @knrc Did you want resource changes in here as well?

dmitri-d (approved):

PR #249, opened 2019-09-17
MAISTRA-788 update operator-sdk to v0.10

rcernich:
The vendor updates are in a separate commit for easier reviewing.


pkg/controller/common/util.go
```diff
@@ -114,24 +111,13 @@ func SetAnnotation(resource metav1.Object, annotation, value string) {
 var initOperatorNamespace sync.Once
 var operatorNamespace string
 
-// XXX: remove after updating to newer version of operator-sdk
-// from newer version of operator-sdk
+// GetOperatorNamespace initializes and caches this operator's namespace; panics on error
 func GetOperatorNamespace() string {
 	initOperatorNamespace.Do(func() {
-		if ns := os.Getenv("POD_NAMESPACE"); ns != "" {
```
dgn, line 121: Maybe we should leave this in; we added this to run the operator locally as if it was deployed to `POD_NAMESPACE` without having to write to `/var/run/secrets/kubernetes.io/serviceaccount/namespace` locally.

rcernich: Added it back in, if `ErrNoNamespace` is returned.

tools.go
```diff
@@ -0,0 +1,7 @@
+// +build tools
```
dgn, line 1: what is this used for?

cmd/manager/main.go
```diff
@@ -99,25 +115,63 @@ func main() {
...
+// It serves those metrics on "http://metricsHost:operatorMetricsPort".
+func serveCRMetrics(cfg *rest.Config) error {
+	// Below function returns filtered operator/CustomResource specific GVKs.
+	// For more control override the below GVK list with your own custom logic.
+	filteredGVK, err := k8sutil.GetGVKsFromAddToScheme(maistrav1.SchemeBuilder.AddToScheme)
+	if err != nil {
+		return err
+	}
+	// Get the namespace the operator is currently deployed in.
+	operatorNs, err := k8sutil.GetOperatorNamespace()
```
dgn: I think we should use our `util.GetOperatorNamespace()` here for consistency

luksa (approved):

dgn (approved):
looks good, just one (very) minor comment and a question

ah btw we have to update `DEVEL.md` to remove `--discoveryCacheDir`

cmd/manager/main.go
rcernich: good catch, i'll update that.

tools.go
rcernich: I'm not 100% certain, but I believe this is required to use the k8s.io/code-generator tooling with go modules.  The file is generated by operator-sdk.

rcernich:
@luksa, let me know if the controller-runtime changes should be in here or not.  (I changed this PR back to WIP.)

PR #250, opened 2019-09-30
MAISTRA-980 Introduce the ServiceMeshMember CRD and controller

luksa:


pkg/apis/maistra/v1/servicemeshmember_types.go
```diff
@@ -0,0 +1,51 @@
...
+type ServiceMeshControlPlaneRef struct {
+	Name      string `json:"name"`
+	Namespace string `json:"namespace"`
+}
+
+// ServiceMeshMemberStatus contains the state last used to reconcile the list
+type ServiceMeshMemberStatus struct {
+	ObservedGeneration    int64    `json:"observedGeneration,omitempty"`
+	ServiceMeshGeneration int64    `json:"meshGeneration,omitempty"`
+	ConfiguredMembers     []string `json:"configuredMembers,omitempty"`
```
rcernich: Do we want this on the member resources?

pkg/controller/servicemesh/member/controller.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	if !contains(member.Namespace, memberRoll.Spec.Members) {
+		memberRoll.Spec.Members = append(memberRoll.Spec.Members, member.Namespace)
+
+		if isNewMemberRoll {
+			err = r.Client.Create(context.TODO(), memberRoll)
+			if err != nil {
+				if errors.IsNotFound(err) {
+					reqLogger.Info("Could not create ServiceMeshMemberRoll, because the referenced namespace doesn't exist", "namespace", memberRoll.Namespace)
```
rcernich: The member roll controller was listening for namespace events, as opposed to requeuing the resource.

pkg/apis/maistra/v1/servicemeshmember_types.go
luksa: I removed this a while ago. I just pushed the commits instead of also commenting here. PTAL.

pkg/controller/servicemesh/member/controller.go
luksa: Also fixed in one of the new commits. PTAL

luksa:
I've now squashed the three commits and rebased to current maistra-1.1 branch. Please take a look.

deploy/examples/use-mesh-rbac-example.yaml
```diff
@@ -0,0 +1,31 @@
...
+  namespace: istio-system
+rules:
+  - apiGroups:
+    - maistra.io
+    resources:
+    - servicemeshcontrolplanes
+    verbs:
+    - use
+    resourceNames:
+    - multitenant-install
```
rcernich: Should this be something like minimal-install?

Also, not to expand the scope, but should we look at constraining the name of the smcp to be the namespace within which it's created?  Or maybe a fixed name, like default?  Seems like this might simplify usage of the member resource.  Something for a follow on discussion.

deploy/examples/use-mesh-rbac-example.yaml
```diff
@@ -0,0 +1,31 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: mesh-users
+  namespace: istio-system
+roleRef:
+  kind: Role
+  name: mesh-user
+subjects:
+- kind: User
```
rcernich, line 25: Should we remove the users and just keep the group?  Maybe add a comment about how to add users using cli?

pkg/apis/maistra/v1/servicemeshmember_types.go
```diff
@@ -0,0 +1,48 @@
...
+
+// ServiceMeshControlPlaneRef is a reference to a ServiceMeshControlPlane object
+type ServiceMeshControlPlaneRef struct {
+	Name      string `json:"name"`
+	Namespace string `json:"namespace"`
+}
+
+// ServiceMeshMemberStatus contains the state last used to reconcile the list
+type ServiceMeshMemberStatus struct {
+}
```
rcernich, line 123: Should we have something here?  Maybe something as simple as `joinTimestamp` and/or `reconcileTimestamp`?  Basically, something that lets the user know whether or not they are a fully configured member.

Perhaps this could be added by the smmr controller when it updates the namespace containing the smm resource?

pkg/controller/servicemesh/member/controller.go
```diff
@@ -0,0 +1,242 @@
...
+
+	err = mgr.GetFieldIndexer().IndexField(&v1.ServiceMeshMember{}, "spec.controlPlaneRef.namespace", func(obj runtime.Object) []string {
+		roll := obj.(*v1.ServiceMeshMember)
+		return []string{roll.Spec.ControlPlaneRef.Namespace}
+	})
+	if err != nil {
+		return err
+	}
+	// watch namespaces and trigger reconcile requests as those that match a member come and go
+	err = c.Watch(&source.Kind{Type: &corev1.Namespace{}}, &handler.EnqueueRequestsFromMapFunc{
```
rcernich: Is this necessary?  If the smcp resource doesn't exist, then creation of the smm resource should fail, so it seems like there shouldn't be a need to watch namespaces at all.

pkg/controller/servicemesh/member/controller.go
```diff
@@ -0,0 +1,242 @@
...
+	memberRoll := &v1.ServiceMeshMemberRoll{}
+	isNewMemberRoll := false
+	err = r.Client.Get(context.TODO(), memberRollKey, memberRoll)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return reconcile.Result{}, err
+		}
+		// MemberRoll doesn't exist, let's create it
+		isNewMemberRoll = true
+		memberRoll = &v1.ServiceMeshMemberRoll{
```
rcernich: Should we add an owner reference back to the smm resource?  Would this facilitate automatic deletion of the smmr resource, once the last smm is deleted?

Do we prevent deletion of the referenced smcp resource?  Should we add an owner reference to smcp too?

pkg/controller/servicemesh/validation/member.go
```diff
@@ -0,0 +1,103 @@
...
+		err := v.decoder.DecodeRaw(req.AdmissionRequest.OldObject, oldSmm)
+		if err != nil {
+			logger.Error(err, "error decoding old object in admission request")
+			return admission.ErrorResponse(http.StatusBadRequest, err)
+		}
+
+		if smm.Spec.ControlPlaneRef.Name != oldSmm.Spec.ControlPlaneRef.Name ||
+			smm.Spec.ControlPlaneRef.Namespace != oldSmm.Spec.ControlPlaneRef.Namespace {
+			logger.Info("Client tried to mutate ServiceMeshMember.spec.controlPlaneRef")
+			return admission.ErrorResponse(http.StatusBadRequest, fmt.Errorf("Mutation of .spec.controlPlaneRef isn't allowed"))
```
dgn, line 61: I understand that it simplifies the controller logic, but this could be frustrating - is this standard behaviour when editing refs?

pkg/controller/servicemesh/validation/member.go
```diff
@@ -0,0 +1,103 @@
...
+type memberValidator struct {
+	client  client.Client
+	decoder atypes.Decoder
+}
+
+var _ admission.Handler = (*memberValidator)(nil)
+var _ inject.Client = (*memberValidator)(nil)
+var _ inject.Decoder = (*memberValidator)(nil)
+
+func (v *memberValidator) Handle(ctx context.Context, req atypes.Request) atypes.Response {
```
dgn, line 29: I think we should check if the namespace is already a member of another mesh, otherwise the MemberController and MemberRollValidator will throw errors back and forth if that's the case

pkg/controller/servicemesh/member/controller.go
dgn: I agree, don't believe we need to watch namespaces; possibly SMCPs though? In case they're deleted and re-created

pkg/apis/maistra/v1/servicemeshmember_types.go
dgn: +1, we should also have ways to report errors, if any occur (e.g. control plane disappeared)

deploy/examples/use-mesh-rbac-example.yaml
dgn: +1 for enforcing 'default' and only referencing a namespace in the Member resource. But that's out of scope here I guess

dgn (commented):
Good work here. Please excuse my wall of text.

TL;DR this will work fine if the user does not make mistakes but I'm afraid they're going to.

The first issue I see is that it's not completely idempotent in that we rely on capturing every event. If we e.g. miss the deletion of a Member object for whatever reason, we could try and add it to another MemberRoll when it's re-created because we don't explicitly check if it is already a member of any mesh. That would then cause errors in the MemberRollValidator. For the user, it looks as if she created a valid ServiceMeshMember object (validation passed) but then nothing happens (except for lots of errors in the logs).

Then there are manual edits of MemberRolls - that's less a problem of implementation but rather of having multiple APIs. Examples:
- If the namespace was manually added to a MemberRoll before, we will just go and try to add it to another one (can be fixed the same way as above)
- Namespaces could be removed from MemberRolls after the fact, so we have dangling Member resources hanging around
- SMCPs can be removed, also leading to dangling Members (potentially solved by ownerRefs?)

So I'm starting to think: should we maybe only support either the MemberRoll or Member API for a given control plane, to avoid these problems?

And we have to make sure to report any errors back to the user in a way that is visible and helps them find a way to recover.

These are all edge cases of course. I think the code as it is will work fine 99% of the time.

deploy/examples/use-mesh-rbac-example.yaml
luksa: Actually, I feel like requiring a specific resource name is an anti-pattern. Also, the usual structure of a reference (e.g. `secretRef`) in core k8s objects is: name + namespace.

pkg/apis/maistra/v1/servicemeshmember_types.go
luksa: I was planning to add this later - when the rest of the team starts using this and it becomes clear what users need. The only thing I'd add right now is some kind of status.

pkg/controller/servicemesh/member/controller.go
luksa: It shouldn't fail. As elsewhere in k8s, users should be able to create resources out of order. Preventing the creation of the `Member` object before the `SMCP` is created would break scenarios where you deploy all resources at once (e.g. `SMCP`, `Member` and app itself).

pkg/controller/servicemesh/member/controller.go
luksa: No. The Member controller takes care of deleting the `SMMR`. This allows it to determine if the SMMR should even be deleted. Currently, it deletes it if it was the one who created it. But I'll need to change this, since it would also delete the SMMR in the following scenario (when it really shouldn't):
- user creates Member object
- controller creates SMMR
- user manually adds another namespace to the SMMR
- user deletes Member object

The controller shouldn't delete the SMMR in this case.

pkg/controller/servicemesh/validation/member.go
luksa: Moving a namespace between meshes is not a thing that most people will want to do. Forcing them to delete the Member and create a new one therefore isn't a big deal. We can always extend the API and allow them to mutate the refs, but it makes sense to make the field immutable at first, and then open it up instead of the other way around. Allowing mutation of this field is problematic, because the controller can't just use the old value of the ref to figure out which SMMR to remove the member from. It would have to list all SMMRs to find the right one. Not a big deal. I can do that in a follow-up PR.

dgn: Agreed. Moving between meshes should IMO be an epic in itself that we should strive to make easier, because it allows for blue/green deployments of control planes. Anyway, separate discussion.

luksa:
> The first issue I see is that it's not completely idempotent in that we rely on capturing every event. If we e.g. miss the deletion of a Member object for whatever reason, we could try and add it to another MemberRoll when it's re-created because we don't explicitly check if it is already a member of any mesh. 

We can't miss the deletion because of the finalizer. The controller removes the finalizer only after removing the namespace from the SMMR's member list. 

> That would then cause errors in the MemberRollValidator. For the user, it looks as if she created a valid ServiceMeshMember object (validation passed) but then nothing happens (except for lots of errors in the logs).

I'll add the status. 

> Then there are manual edits of MemberRolls - that's less a problem of implementation but rather of having multiple APIs. Examples:
> 
> * If the namespace was manually added to a MemberRoll before, we will just go and try to add it to another one (can be fixed the same way as above)

The MemberRoll validator prevents this. 

> * Namespaces could be removed from MemberRolls after the fact, so we have dangling Member resources hanging around

The periodic resync should take care of this, but I'll replace the namespace watch with a SMMR watch and this should then add the member to the SMMR immediately after someone manually removes it.

> * SMCPs can be removed, also leading to dangling Members (potentially solved by ownerRefs?)

There's nothing wrong with a dangling Member. Users should always be able to create resources in any order they want. This implies that they should also be able to delete a referenced resource and then re-create it, without having to delete every other object that refers to it. Also, k8s shouldn't delete resources that a users creates manually. When a user creates a resource, they own it.   

> So I'm starting to think: should we maybe only support either the MemberRoll or Member API for a given control plane, to avoid these problems?

No need. We can make everything work even if we allow both at once. Allowing only one would mean that all existing deployments would have to use just the MemberRoll. This would effectively prevent users from switching to the Member mechanism. 

> And we have to make sure to report any errors back to the user in a way that is visible and helps them find a way to recover.

Absolutely. I never intended the Member to have an empty status. I just wanted to let others from the team to try this out and see what users really need.

deploy/examples/use-mesh-rbac-example.yaml
luksa: I've changed this to minimal-install.

deploy/examples/use-mesh-rbac-example.yaml
luksa: Since this is an example, it makes sense to show both users and groups are possible (although someone familiar with RBAC should already know this).

pkg/controller/servicemesh/validation/member.go
luksa: If the MemberRollValidator prevents the controller from adding the namespace to the SMMR, the returned validation message will be written to the Member's status. I don't see how the error would be bouncing back and forth.

luksa:
I've remembered why the controller needs to watch namespaces. If you create a Member that references a SMCP in a nonexistent namespace, the controller needs to create the SMMR as soon as the namespace is created. 

On that note, I have now also added a watch on SMMRs, to ensure the controller reverts any incompatible changes to the SMMR made by a user (e.g. a user removes a namespace from the SMMR's member list, but the ns still has a Member object).

pkg/controller/servicemesh/validation/member.go
dgn: It looked like we're just wrapping the error in "Could not update ServiceMeshMemberRoll" and return it, which IIRC triggers another reconciliation? Which will then again throw an error, ... Haven't tested this though

rcernich:
Regarding SMMR/SMM interaction, I think we should enforce a single usage pattern: you either use SMMR or you use SMM, but not both.  We could enforce this through the validators, e.g. if the user creates SMMR, then creation of SMM should be rejected, and vice versa.  We could do this in a follow on PR.

rcernich:
Regarding namespace watch, won't the validation fail, because the referenced smcp resource is non existent, hence the user creating the smm would be able to `use` it?  (Although, I guess cluster-admin could create it regardless, so maybe just leave it to be safe.)

luksa:
> Regarding SMMR/SMM interaction, I think we should enforce a single usage pattern: you either use SMMR or you use SMM, but not both. We could enforce this through the validators, e.g. if the user creates SMMR, then creation of SMM should be rejected, and vice versa. We could do this in a follow on PR.

But that kind of validation would just be complicating things. Right now, the Member controller is just a helper that adds or removes members to/from a SMMR instead of the user doing that manually. The whole member system was designed to be an additive change instead of replacing the MemberRoll - specifically to reduce the chance of breaking things.

luksa:
> Regarding namespace watch, won't the validation fail, because the referenced smcp resource is non existent, hence the user creating the smm would be able to `use` it? (Although, I guess cluster-admin could create it regardless, so maybe just leave it to be safe.)

The SMCP doesn't need to exist for a user to have the privilege to use it. This ensures you can create objects out of order. RBAC rules are always completely separate from the (existence of) objects.

pkg/controller/servicemesh/validation/member.go
JStickler: "Moving a namespace between meshes is not a thing that most people will want to do."  And yet there's already a JIRA to document that very scenario (https://issues.jboss.org/browse/MAISTRA-859).  Just an FYI.

pkg/controller/servicemesh/validation/member.go
dgn: I just verified this; it will throw errors forever in the operator log. The scenario I'm testing is:
- namespace a is part of mesh cp (manually added to cp's MemberRoll)
- create a Member resource in namespace a pointing to mesh cp2
- the validating webhook for Member resources passes *
- the Member controller tries to add namespace a to cp's MemberRoll
- validating webhook for MemberRoll resources fails: already member of cp
- Member controller will throw errors forever (exponential backoff means they become rarer over time, but they're there)

I think the most consistent way would be to not accept a Member resource that tries to re-assign a namespace ie checking whether it is already member of another namespace *in the validating webhook* (see *) - this is the same behaviour as with MemberRolls

luksa: It wouldn't throw the error forever - only until it can finally successfully add the member to cp2 (when the user manually removes it from cp). That's not really wrong - that's how all k8s controllers should work - they should retry indefinitely. 

What is wrong is that the operator logs this as an error. Instead of doing that, it should only emit an event and/or update the Member's status to notify the user of why it can't add the member to the mesh. 

While adding a validating webhook would help notify the user immediately, it would also force users to first remove the member from the old cp and only then create the Member object. This would effectively prevent users from doing a `kubectl apply -f *.yaml`, where one of the yaml removes the member from the SMMR and the other one creates the Member. That's why I'm not a big fan of these types of validating webhooks. The beauty of Kubernetes is that it doesn't force you to do things in a specific order (initially this was true for virtually everything; now some operations do demand performing them in a specific order - e.g. create CRD first, then do other stuff - but we should do our best to not introduce these types of ordering requirements). 

There's one other reason against introducing a validating webhook like this: if the webhook uses a local cache (as it should and as it usually does), it could reject a perfectly valid request simply because its cache is stale (e.g. it still contains a SMMR that contains the member, even though the user has just removed it). To work around this, the webhook would have to bypass the cache and obtain a list of all SMMRs from the API server every time it is invoked. List operations aren't cheap, so this would cause additional strain on the API server.  

In any case, we can add the validating webhook later.

dgn: > What is wrong is that the operator logs this as an error. Instead of doing that, it should only emit an event and/or update the Member's status to notify the user of why it can't add the member to the mesh.

Agreed. I can live with moving the error message to the status.

pkg/controller/servicemesh/member/controller.go
```diff
@@ -341,12 +364,58 @@ func (r *MemberReconciler) updateStatus(member *maistra.ServiceMeshMember, recon
...
+			freshMember := unstructured.Unstructured{}
+			freshMember.SetAPIVersion(maistra.SchemeGroupVersion.String())
+			freshMember.SetKind("ServiceMeshMember")
+
+			err = r.Client.Get(context.TODO(), types.NamespacedName{member.Namespace, member.Name}, &freshMember)
+			if err != nil {
+				break
+			}
+
+			unstructuredStatus, err := converter.ToUnstructured(member.Status)
```
rcernich, line 399: Is `ToUnstructured()` part of this commit, or does it already exist?

luksa: It's part of apimachinery.

deploy/examples/use-mesh-rbac-example.yaml
rcernich: Should we move the role into the smcp yaml?  We can do that as follow on work.

pkg/apis/maistra/v1/servicemeshmember_types.go
```diff
@@ -0,0 +1,123 @@
...
+// Condition represents a specific condition on a resource
+type ServiceMeshMemberCondition struct {
+	Type               ServiceMeshMemberConditionType   `json:"type,omitempty"`
+	Status             core.ConditionStatus             `json:"status,omitempty"`
+	LastTransitionTime metav1.Time                      `json:"lastTransitionTime,omitempty"`
+	Reason             ServiceMeshMemberConditionReason `json:"reason,omitempty"`
+	Message            string                           `json:"message,omitempty"`
+}
+
+// GetCondition removes a condition for the list of conditions
```
brian-avery, line 87: Retrieves? This isn't removing anything.

brian-avery (approved):

rcernich (approved):
I'm okay with this, although I think we should have all the member logic implemented in a single controller.  I think that would ease the synchronization issues between the two controllers.  I would recommend using the existing member roll logic, but build the required members list off the union between the member roll and the list of member objects.

dgn (approved):

dgn:
> I'm okay with this, although I think we should have all the member logic implemented in a single controller. I think that would ease the synchronization issues between the two controllers. I would recommend using the existing member roll logic, but build the required members list off the union between the member roll and the list of member objects.

I can also imagine that we might want to merge them in the future, to do things such as create/delete Member resources when a MemberRoll is updated- then the two APIs would be fully aligned. But we'll have to see if we need that.

luksa:
> > I'm okay with this, although I think we should have all the member logic implemented in a single controller. I think that would ease the synchronization issues between the two controllers. I would recommend using the existing member roll logic, but build the required members list off the union between the member roll and the list of member objects.
> 
> I can also imagine that we might want to merge them in the future, to do things such as create/delete Member resources when a MemberRoll is updated- then the two APIs would be fully aligned. But we'll have to see if we need that.

Yes, merging the two controller may make sense, although I haven't yet seen any other controller operate on two separate CRs. We can try doing that later. The biggest reason for making the Member controller separate was to reduce the impact on existing functionality - the member controller is just an add-on, which doesn't require any changes to the existing SMCP and SMMR controller, ensuring it can't impact existing users at all.

rcernich:
I guess it depends on how you view the controller: does it manage a specific resource or does it manage a feature?

I was also thinking about this last night and it was wondering if there might be any performance implications doing a list across all namespaces for each reconcile.  I was wondering if we could tune the logic by using custom events (e.g. NamespaceCreated, NamespaceDeleted, MemberCreated, MemberRollUpdated, etc.).  Anyway, food for thought.

PR #261, opened 2019-10-22
MAISTRA-1005 Set Status.ServiceMeshReconciledVersion after reconciling SMMR

dgn:
This field was introduced as part of the work enabling operator upgrades, but even though we read it, we never set it, leading the operator to always reconcile the SMMR causing issues downstream.

Note that the endless reconciliation loop was triggered by the changing order of configuredMembers: with every `Status.Update()`, we would trigger `Reconcile()` again. As the if-clause was always true (because we did not set `ServiceMeshReconciledVersion`), this would change the order of configuredMembers, which would then trigger `Reconcile()` again, and so on.


rcernich (approved):

rcernich:
@dgn, thanks for cleaning this up.  I can't believe I missed this.

knrc (approved):

PR #264, opened 2019-10-28
MAISTRA-1021: Fix the replaces value to allow upgrade

knrc:


knrc:
@dmitri-d Could you please cast you eye over this and make sure I've not forgotten anything else?

rcernich (approved):

rcernich:
Is this a list field?  If so, should we list 1.0.0 and 1.0.1?  Also, these changes need to be made to maistra (community) version as well.

knrc:
> Is this a list field? If so, should we list 1.0.0 and 1.0.1? Also, these changes need to be made to maistra (community) version as well.

There is a separate issue in this sprint for the community (maistra) config.

dmitri-d (approved):
lgtm

knrc:
Note from the operator docs it looks as if replaces should just be a single version, forming a DAG through which older versions are discovered.

brian-avery (approved):

PR #273, opened 2019-12-13
MAISTRA-1062 add helm charts for maistra-1.1 to source control

rcernich:
Also includes MAISTRA-691, which adds a Makefile for easier builds.


dgn (commented):
Should we wait with adding the 1.1 charts until the rebase is done? Right now the 1.1 charts are older than the 1.0 charts (because they don't have the z-stream updates)

rcernich:
The Maakefile has an update-1.1-charts target, which will copy the charts down and patch them.  We can run that whenever, understanding that we might also have to tweak the patch-charts.sh script and the overlays.  For now, consider it a placeholder based on what's in the 1.1 branch.

rcernich:
Moving back to wip, as I need to update the controller tests.

rcernich:
Fixed SMMR unit tests and added some to verify operator update behavior to ensure behavior when upgrading operator from v1.0 to v1.1 (i.e. make sure new fields are properly defaulted for v1.0).

pkg/controller/common/helm.go
```diff
@@ -25,7 +25,8 @@ func init() {
 }
 
 const (
-	istioVersion = "v1.1"
+	DefaultMaistraVersion = "v1.1"
+	LegacyMaistraVersion  = "v1.0"
```
shamsher31, line 29: Curious to know if we can keep different supported versions in ENV. which might be useful in [other places as well](https://github.com/Maistra/istio-operator/pull/273/commits/a7c45eadc23d14dff34b4d17db515685700e72db#diff-a1c55792d1e2085253441f0b5a6d895aR26).

Makefile
```diff
@@ -35,21 +42,71 @@ endif
...
+################################################################################
+.PHONY: update-remote-maistra-1.0
+update-remote-maistra-1.0:
+ifeq "${OFFLINE_BUILD}" "false"
+	git fetch ${GIT_UPSTREAM_REMOTE} maistra-1.0:maistra-1.0
+endif
+
+.PHONY: update-1.0-charts
+update-1.0-charts: update-remote-maistra-1.0
+	git checkout ${GIT_UPSTREAM_REMOTE}/maistra-1.0 -- ${SOURCE_DIR}/resources/helm/v1.0
```
dgn, line 70: This is nice. Didn't know you could do that.

Makefile
```diff
@@ -0,0 +1,95 @@
...
+
+################################################################################
+# update-generated-code target regenerates k8s api related code
+################################################################################
+.PHONY: update-generated-code
+update-generated-code:
+	${SOURCE_DIR}/build/codegen/update-generated.sh
+
+.PHONY: build
+build: compile update-generated-code generate-charts
```
luksa: `make build` fails for me:
- `vendor/k8s.io/code-generator/generate-groups.sh` isn't executable
- `git checkout upstream/maistra-1.0` fails, since there's no `resources` dir in the 1.0 branch (yet?)

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -336,7 +337,7 @@ func mergeValues(base map[string]interface{}, input map[string]interface{}) map[
 	return base
 }
 
-func (r *ControlPlaneReconciler) getSMCPTemplate(name string) (v1.ControlPlaneSpec, error) {
+func (r *ControlPlaneReconciler) getSMCPTemplate(name string, maistraVersion string) (v1.ControlPlaneSpec, error) {
```
dgn, line 340: In addition to finding the right charts, we will also have to change the default `tag` in the `base` template to the current version (e.g. `1.1.0`) - maybe we should start doing that programmatically, ie by merging it with a code-defined root template at reconcile time that just sets the `global.tag` to the current version, taken from `pkg/version`

pkg/controller/common/helm.go
rcernich: These are just constants to avoid using '"v1.0"' or '"v1.1"' throughout the code.  The "other places" are used to configure the correct 'NetworkAttachmentDefinition' on the mesh members, depending on which version of the mesh they are using.

'DefaultMaistraVersion' aligns with the latest version supported by the operator.  This is the version that will be used if the user does not specify a version in their SMCP resource.

'LegacyMaistraVersion' aligns with SMCP resources that do not have a version specified because they were managed by an older version of the operator.  This allows the new operator to initialize the new field with the correct value.  (SMCP resources managed by the new operator will have the field properly initialized.)

Makefile
rcernich: I updated the PR to fix the permissions issue.

The 'git checkout' issue will remain until #274 is merged.  I can look at using a variable in the Makefile so users can specify a repo/branch.  Let me know if you think I should add this (you'd then have to point to my branch in #274 to get it working).

pkg/controller/servicemesh/controlplane/controller.go
```diff
@@ -255,6 +255,15 @@ func (r *ReconcileControlPlane) Reconcile(request reconcile.Request) (reconcile.
 		reqLogger.V(1).Info("Adding finalizer", "finalizer", common.FinalizerName)
 		finalizers.Insert(common.FinalizerName)
 		instance.SetFinalizers(finalizers.List())
+		// set default version if necessary
+		if len(instance.Spec.Version) == 0 {
+			if len(instance.Status.AppliedVersion) == 0 {
+				instance.Spec.Version = common.DefaultMaistraVersion
```
dgn: Should we really write to the `Spec`? I think this should go into `Status`

dgn (commented):
I'm not sure about the new Version field, maybe I don't quite get it yet- but shouldn't we be able to pull that info from `Status.ReconciledVersion`?

I'd prefer to move all the Maistra-version-related `const` definitions and mappings to a central location, close to the current operator version, e.g. `pkg/version`. That would make adding support for new versions easier in the future, as we'd only have to touch one file. That's just a matter of taste though, they can keep living in `common` as well.

In general it's relatively hard to review this PR, even when going through it commit by commit it is still very large. Will keep going

pkg/controller/servicemesh/controlplane/controller.go
rcernich: Good question.  My rationale was that version selection is an explicit action and if the user doesn't specify the version, we should initialize it with the default, to help prevent unwanted upgrades in the future.  If the field is empty, it is assumed that the version is 'v1.0', as the field did not exist in the previous version of the operator.

dgn: We should probably converge on a single place to specify the version. Maybe you're right and it should be `Spec.Version`, in that case I'd suggest we make that the full version though (`1.1.0` instead of `v1.1`) and derive the short form from that (we know that 1.1.0 needs v1.1 charts) - but also give users the possibility to specify `1.1` and we will select the latest `1.1.x` release for them. We can then populate the `global.tag` field based on that choice

dgn: But I still think we should only write to `Status` ;) Maybe @luksa can enlighten us

rcernich:
Some general comments on the design:

Charts are intended to be stored in ./resources/helm/<version>, where we're currently supporting versions v1.0 and v1.1.  Charts from older versions are sourced from the maintenance branches for those versions, e.g. ./resources/helm/v1.0 is sourced from istio-operator/maistra-1.0/resources/helm/v1.0.  This allows 'patch-charts.sh' to be maintained independently for each specific branch.

Regarding versioning, we are only providing support for major.minor versions in the operator.  This means that the code in this PR supports versions 1.0.4 and 1.1.0.  Moving forward, ..z releases will require the use of a specific operator.  It also means that the current version of the operator will need to be released when any supported version is updated, e.g. operator v1.1.1 might support v1.0.5 and v1.1.0, while operator v1.1.2 might support v1.0.5 and v1.1.1.  In other words, folks using v1.0.4 of Maistra who want to upgrade to v1.0.5 would need to update their operator from v1.0.4 to v1.1.1 in this example.

Because we are only supporting major.minor versions in the operator, the version field in the SMCP resource only allows specification of major and minor version, which is currently the set [ v1.0, v1.1 ].  I've included the 'v' prefix, but am happy to remove that.

pkg/controller/servicemesh/controlplane/reconciler.go
rcernich: See my comment below on versioning, but we're only supporting the specification of major.minor within the SMCP spec.  ..z versions would require the use of a specific operator.  All we're doing with our multi-version support is adding support for the previous latest major.minor release, along with the current latest major.minor release (i.e. 1.0.4 and 1.1.0).

Maybe 'maistraVersion' is too specific, implying a full version, when it really only identifies a major.minor release.

Having said all of that, things like 'global.tag' are still defaulted in the default template/profile, just as they were previously.

pkg/controller/servicemesh/controlplane/controller.go
rcernich: I like the idea of using the full version, however there could be chart differences between .z releases which would mean a downgrade of the operator might get you 1.1.0 charts with a 1.0.2 tag.  I'm not sure if that's an issue or not, just saying.

luksa: I actually made a comment on this yesteday, but it seems to have disappeared. Weird.

Yes, I'm not a fan of controllers updating the object's spec. It always causes problems. Users should be the only ones modifying it. For example, what if the user applies a modified SMCP after the controller has modified the version. Does the version get reset to "" again?

Defaulting, which is what the code here is doing, is usually implemented in admission webhooks. If the user leaves the version empty, it should be defaulted to the current version. But any existing SMCPs without an empty version should be assumed to be v1.0. The controller should handle this without modifying the SMCP. 

Actually, I'm not sure this code here even does what it should. The object's status is always empty when this code is invoked (when the finalizer isn't there yet), so it will always set the default maistra version. Plus, this is never invoked if the finalizer is already set (e.g. on existing SMCPs).

resources/helm/v1.1/istio_cni/templates/daemonset.yaml
```diff
@@ -88,7 +88,51 @@ spec:
...
+            - name: MOUNTED_CNI_NET_DIR
+              value: "/host/etc/cni/multus/net.d"
+            # Name of the CNI config file to create.
+            - name: CNI_CONF_NAME
+              value: "istio-cni.conf"
+            # Name of the kubeconfig file used by CNI agent
+            - name: KUBECFG_FILE_NAME
+              value: "istio-cni.kubeconfig"
+            - name: CNI_BINARIES_PREFIX
+              value: ""
```
luksa: Duplicate `CNI_BINARIES_PREFIX` entry

resources/helm/v1.1/istio_cni/templates/daemonset.yaml
```diff
@@ -88,7 +88,51 @@ spec:
...
+              value: "v1.1-"
+            # Directory where the CNI config file should be created in
+            - name: CNI_NET_DIR
+              value: "/etc/cni/multus/net.d"
+            # Directory where the CNI config file should be created in (host path mounted in the container)
+            - name: MOUNTED_CNI_NET_DIR
+              value: "/host/etc/cni/multus/net.d"
+            # Name of the CNI config file to create.
+            - name: CNI_CONF_NAME
+              value: "istio-cni.conf"
```
luksa: Same filename as for v1.0 above. Should be a separate file.

pkg/controller/servicemesh/memberroll/controller_test.go
```diff
@@ -190,13 +190,77 @@ func TestReconcileFailsIfListingNamespacesFails(t *testing.T) {
...
+	assert.DeepEquals(updatedRoll.Status.ConfiguredMembers, []string{appNamespace}, "Unexpected Status.ConfiguredMembers in SMMR", t)
+	assert.Equals(updatedRoll.Status.ServiceMeshGeneration, controlPlane.Status.ObservedGeneration, "Unexpected Status.ServiceMeshGeneration in SMMR", t)
+	assert.Equals(updatedRoll.Status.ServiceMeshReconciledVersion, controlPlane.Status.GetReconciledVersion(), "Unexpected Status.ServiceMeshReconciledVersion in SMMR", t)
+	assert.Equals(updatedRoll.Status.MeshVersion, common.LegacyMaistraVersion, "MemberRoll.Status.MeshVersion should have been updated to v1.0 (from empty string)", t)
+
+	meshNetAttachDefName := "istio-cni"
+	assertNamespaceReconciled(t, cl, appNamespace, controlPlaneNamespace, meshNetAttachDefName, []rbac.RoleBinding{*meshRoleBinding})
+}
+
+func TestReconcileReconcilesAddedMemberV1_0(t *testing.T) {
```
luksa: Since this test is almost identical to `TestReconcileReconcilesAddedMember`, it would make sense to use subtests instead of a separate test.

pkg/controller/servicemesh/controlplane/controller.go
rcernich: I'll fix this.

pkg/controller/servicemesh/memberroll/controller_test.go
rcernich: Will convert.

pkg/controller/servicemesh/controlplane/controller.go
```diff
@@ -255,6 +255,11 @@ func (r *ReconcileControlPlane) Reconcile(request reconcile.Request) (reconcile.
 		reqLogger.V(1).Info("Adding finalizer", "finalizer", common.FinalizerName)
 		finalizers.Insert(common.FinalizerName)
 		instance.SetFinalizers(finalizers.List())
+		// set default version if necessary
+		if len(instance.Spec.Version) == 0 {
+			instance.Status.AppliedVersion = common.DefaultMaistraVersion
```
rcernich, line 260: Now we're only initializing `Status.AppliedVersion` if the `Spec.Version` is unset.

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -416,6 +417,17 @@ func (r *ControlPlaneReconciler) validateSMCPSpec(spec v1.ControlPlaneSpec) erro
 func (r *ControlPlaneReconciler) renderCharts() error {
 	//Generate the spec
 	r.Status.LastAppliedConfiguration = r.Instance.Spec
+	if len(r.Status.LastAppliedConfiguration.Version) == 0 {
+		// initialize version
+		if len(r.Status.AppliedVersion) == 0 {
+			// this must be from a 1.0 operator
+			r.Status.LastAppliedConfiguration.Version = common.LegacyMaistraVersion
```
rcernich, line 424: ...And we're only using the legacy (`v1.0`) version when `Status.AppliedVersion` is not set.

resources/helm/v1.1/istio_cni/templates/daemonset.yaml
```diff
@@ -0,0 +1,155 @@
...
+          env:
+            # Directory where the CNI config file should be created in
+            - name: CNI_NET_DIR
+              value: "/etc/cni/multus/net.d"
+            # Directory where the CNI config file should be created in (host path mounted in the container)
+            - name: MOUNTED_CNI_NET_DIR
+              value: "/host/etc/cni/multus/net.d"
+            # Name of the CNI config file to create.
+            - name: CNI_CONF_NAME
+              value: "v1.1-istio-cni.conf"
```
rcernich, line 120: @luksa, is this correct?  I wasn't sure if it needed to match the prefixed binary name or not.

dgn (commented):
I will attempt to test-drive this on Monday using an actual OCP cluster and an OSSM control plane. @rcernich have you tried that? I'm guessing I'll have to jump through some hoops with (CNI) images and such

rcernich:
@dgn, I have not.  I plan to start testing this today.  (Just need to find some 1.1 images ;))

resources/helm/v1.1/istio_cni/templates/daemonset.yaml
```diff
@@ -62,8 +62,8 @@ spec:
       containers:
         # This container installs the Istio CNI binaries
         # and CNI network config file on each node.
-        - name: install-cni
-          image: "{{ .Values.image }}"
+        - name: install-cni-v1_0
```
luksa: Creation of this DaemonSet fails because the container name can't contain underscores.

rcernich: i should know better :/

rcernich: fixed

resources/helm/v1.1/istio_cni/templates/daemonset.yaml
luksa: The name of the NetworkAttachmentDefinition object needs to match the name of the CNI config file. IIRC, the binary name can be configured in the conf file, but defaults to the name of the config file/NAD object name.

rcernich: We're using `v1.1-istio-cni` for the NAD name, so it sounds like we're okay.

https://github.com/Maistra/istio-operator/pull/273/files#diff-a1c55792d1e2085253441f0b5a6d895aR27

rcernich:
Assuming nothing further goes in upstream, this is good to merge as soon as #274 is merged.

PR #274, opened 2019-12-13
MAISTRA-1061 add helm charts for maistra-1.0 to source control

rcernich:


rcernich:
This has been updated to the latest 1.0.4 charts.

rcernich:
I'll set the okay to merge label once we're clear of 1.0.4

resources/helm/v1.0/istio/charts/grafana/values.yaml
```diff
@@ -0,0 +1,93 @@
+#
+# addon grafana configuration
+#
+enabled: false
+replicaCount: 1
+image:
+  repository: grafana/grafana
```
dgn, line 7: This should have been changed by the grafana patch scripts. Did the scripts run against this?

dgn (commented):
Should the patch-scripts have been executed against these files already? We might have to do some path tweaking to get them to work properly

resources/helm/v1.0/istio/charts/grafana/values.yaml
rcernich: They should have been.  These are pulled in from the maistra-1.0 branch and should match what we're using in 1.0.4.  I'll have to verify, but this might be overridden in the templates (profiles).

rcernich: Verified.  This isn't use and is overridden by a string value in the templates, e.g. https://github.com/Maistra/istio-operator/blob/maistra-1.1/deploy/smcp-templates/servicemesh#L16

PR #278, opened 2019-12-23
MAISTRA-1074 Extract common finalizer handling code

luksa:


luksa:
Note: this PR extracts the finalizer code from the Member controller only. I'll remove the code from the other two controllers in follow-up PRs, because I also need to make additional changes to ensure removing the finalizer works properly in case of conflicts (the watch precondition code needs to ensure reconciliation is triggered during periodic resyncs and resource status changes).

rcernich:
Hey @luksa, these changes look fine, but if we're refactoring, maybe we should look at this conceptually.  For example, setting up the finalizer is part of the initialization process (or check?) when entering the loop.  For some of the multi-version support work, I was looking at initializing some of the spec (version attribute) and was wedging that into where the finalizer was being added (i.e. treating initialization of the finalizer as a more general initialization step).  Another option may be to treat this like a factory, that returns the appropriate reconciler for the given state, e.g. InitializeResource (deletionTimestamp == nil, finalizer not set), ReconcileResource (deletionTimestamp == nil, finalizer set), FinalizeResource (deletionTimestamp != nil, finalizerSet), ResourceDeleted (noop, deletionTimestamp !=n, finalizer not set).  Maybe this could be some follow on work.

rcernich (approved):

dgn (approved):
lgtm, should we squash before merging?

PR #279, opened 2019-12-23
MAISTRA-984 Add Member and MemberRoll controller unit tests

luksa:


rcernich (approved):

pkg/controller/servicemesh/member/controller_test.go
```diff
@@ -0,0 +1,382 @@
...
+
+	updatedMember := test.GetUpdatedObject(cl, member.ObjectMeta, &maistra.ServiceMeshMember{}).(*maistra.ServiceMeshMember)
+
+	assert.DeepEquals(updatedMember.GetFinalizers(), []string{common.FinalizerName}, "Invalid finalizers in SMM", t)
+
+	// TODO: retrieve actions performed on client and check if no unexpected actions were performed (in all other tests, too!)
+	actions := tracker.Actions()
+	test.AssertNumberOfActions(t, actions, 3)
+	test.AssertGetAction(t, actions[0], member.ObjectMeta, member)
+	//test.AssertUpdateAction(t, actions[1], member.ObjectMeta, member)
```
dgn: I guess we can remove this line if we're not going to implement the function?

pkg/controller/servicemesh/member/controller_test.go
```diff
@@ -0,0 +1,382 @@
...
+func TestReconcileWorksIfMembersRollDoesNotExistWhenRemovingMember(t *testing.T) {
+	member := newMember()
+	member.DeletionTimestamp = &oneMinuteAgo
+
+	_, _, r := createClientAndReconciler(t, member)
+
+	assertReconcileSucceeds(r, t)
+}
+
+// TODO: what if SMMR is created by the controller, but then an additional namespace is added manually.. should the SMMR be deleted when you delete the Member?
```
dgn: Let's file a ticket to tackle this later.

Idea: also add a checksum of the SMMR spec to the annotations so we can identify whether it was always only edited by the controller, and if not, we remove the annotations - ie everytime there's a memberroll change and the annotation checksum does not match its spec, we detected a human change to the memberroll and deactivate automatic deletion (ie remove created-by annotation). We could log that as INFO. This would have to go into the docs as well ("if you manually edit an SMMR, it will still be updated by the controller, but it will no longer be removed automatically when all member resources referencing it have been deleted")

pkg/controller/servicemesh/member/controller_test.go
```diff
@@ -0,0 +1,382 @@
...
+
+	cl, tracker, r := createClientAndReconciler(t, member)
+
+	assertReconcileSucceeds(r, t)
+
+	updatedMember := test.GetUpdatedObject(cl, member.ObjectMeta, &maistra.ServiceMeshMember{}).(*maistra.ServiceMeshMember)
+
+	assert.DeepEquals(updatedMember.GetFinalizers(), []string{common.FinalizerName}, "Invalid finalizers in SMM", t)
+
+	// TODO: retrieve actions performed on client and check if no unexpected actions were performed (in all other tests, too!)
```
dgn: Thinking about it, I'm not sure we should do this. It seems a bit brittle to be checking every action; we should be fine with checking just ins and outs, not how it's done exactly, otherwise we might end up constantly editing these tests.

I'm fine with having a test that checks explicit actions if that is used to verify correct behaviour; but for most of the tests it doesn't matter what actions are executed and in what order, as long as the results are as expected.

go.mod
```diff
@@ -47,7 +47,7 @@ replace (
 )
 
 // maistra forks
-replace sigs.k8s.io/controller-runtime => github.com/maistra/controller-runtime v0.1.13-0.20190926110539-8c6fdfaa0ad1
+replace sigs.k8s.io/controller-runtime => github.com/maistra/controller-runtime v0.1.13-0.20191029133215-623f07679978
```
dgn: hmm. I had to change the date/time to `20191029124451` to make this work. Otherwise I'd get:
`go: github.com/maistra/controller-runtime@v0.1.13-0.20191029133215-623f07679978: invalid pseudo-version: does not match version-control timestamp (2019-10-29T12:44:51Z)`

pkg/controller/servicemesh/member/controller_test.go
```diff
@@ -0,0 +1,382 @@
...
+func assertReconcileSucceeds(r *MemberReconciler, t *testing.T) {
+	res, err := r.Reconcile(request)
+	if err != nil {
+		t.Fatalf("Reconcile failed: %v", err)
+	}
+	if res.Requeue {
+		t.Error("Reconcile requeued the request, but it shouldn't have")
+	}
+
+	// TODO: check if Reconcile mutated the object in the local cache (it should never do that)
```
dgn: doesn't controller-runtime always write to the API server? not sure we'd have to test this

dgn (approved):
Thank you, this is very good. Should we wait for the Makefile PR to merge and then add a `test` target in this one?

go.mod
luksa: That's odd. The commit's timestamp is 13:32:15:
```
commit 623f07679978e887bb5bf42e4c647dd5173924c1 (HEAD -> maistra-1.1, origin/maistra-1.1, origin/HEAD)
Author: Marko Lukša <marko.luksa@gmail.com>
Date:   Tue Oct 29 13:32:15 2019 +0100
```

I wonder where it's getting the 12:44:51 timestamp. What go version are you using?

luksa: Oh, I see: 
```
$ git log --pretty=fuller
commit 623f07679978e887bb5bf42e4c647dd5173924c1 (HEAD -> maistra-1.1, origin/maistra-1.1, origin/HEAD)
Author:     Marko Lukša <marko.luksa@gmail.com>
AuthorDate: Tue Oct 29 13:32:15 2019 +0100
Commit:     Marko Lukša <marko.luksa@gmail.com>
CommitDate: Tue Oct 29 13:44:51 2019 +0100
```

Definitely caused by a different go version.

pkg/controller/servicemesh/member/controller_test.go
luksa: Removed.

pkg/controller/servicemesh/member/controller_test.go
luksa: I'm a bit torn on this. Even if the end state is what we want, but the controller performs many unnecessary actions, I'd want the test to signal that. I have seen controller do what they should, but they also performed way to many actions to do it. 

I've now changed the code so it only checks the _number_ of _write_ actions. This should be enough.

go.mod
dgn: Interesting. I'm on 1.13.3

pkg/controller/servicemesh/member/controller_test.go
luksa: For now, I've added another test that checks if the controller leaves the SMMR alone when someone adds another namespace to it.

The controller already handles such scenarios properly - it leaves the SMMR alone, but deletes it later (after the user removes the manually-added namespace from it). I don't really think we need to complicate things further. Most users will only ever use one of the two options - either create and manage the SMMR manually, or use ServiceMeshMember objects instead of SMMR.

go.mod
luksa: I was using 1.12, but have since tried 1.13.5 and 1.13.3 just now. Neither of them complained about the incorrect pseudo version, so I'm not sure what's going on.

pkg/controller/servicemesh/member/controller_test.go
luksa: When using the cache directly, you're supposed to clone the object before mutating it, since cached objects are shared by all the controllers running in the same process. Luckily, the controller-runtime client itself already clones the object, so even if we mutate it, we're not doing anything wrong. I've removed the TODO now.

pkg/controller/servicemesh/member/controller_test.go
dgn: Makes sense, I'm happy to keep it as is

pkg/controller/servicemesh/member/controller_test.go
dgn: Good to know, thanks! :+1:

go.mod
dgn: Hmm. They introduced version validation in 1.13, awkward you didn't see the same behaviour: https://golang.org/doc/go1.13#version-validation

luksa: Looks like it doesn't validate it any more, since I already have it cached locally. I've now changed it to the correct timestamp. Thanks!

PR #280, opened 2019-12-23
MAISTRA-1075 A few minor fixes

luksa:


pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -298,8 +298,6 @@ func (r *ReconcileMemberList) Reconcile(request reconcile.Request) (reconcile.Re
 	deletedMembers := configuredMembers.Difference(allNamespaces)
 	unconfiguredMembers := allNamespaces.Intersection(requiredMembers.Difference(configuredMembers))
 
-	// always include the mesh namespace in the configured list
-	configuredMembers.Insert(instance.Namespace)
```
rcernich, line 300: I'm not sure if the other components (e.g. galley, kiali) rely on the control plane namespace being in the configured members list or not.

luksa: The `configuredMembers` set isn't used anywhere from this point on, so adding the control plane namespace (or not adding it) doesn't affect anything.

dgn (approved):

jwendell (approved):

PR #282, opened 2020-01-07
MAISTRA-1080: update operator support to maistra 1.0.4

knrc:


rcernich (approved):
lgtm.

we should consider adding skip versions to the csv files so users can upgrade from older versions without having to incrementally update; something for the future.

knrc:
Merging by hand for the release

PR #283, opened 2020-01-08
Configure mergify to squash commits unless "don't squash" label is present

luksa:


dgn (commented):
In general I like squashing, but I'm not sure this will preserve the `MAISTRA-XXX Description` format we're adhering to.

luksa:
I believe it will - the title of the PR becomes the first line of the commit message and the included commit messages are added as additional lines to the merged commit message.

So as long as we include `MAISTRA-XXX Description` in the PR title, we should be fine. And I think most of us do that already.

rcernich (approved):

dgn (approved):
let's do it; I think this will need manual merging though

PR #284, opened 2020-01-08
MAISTRA-984 Additional MemberRoll controller tests

luksa:


pkg/controller/servicemesh/memberroll/namespace_reconciler_test.go
```diff
@@ -0,0 +1,248 @@
...
+	// check that role binding was removed
+	roleBinding := &rbac.RoleBinding{}
+	err = cl.Get(context.TODO(), types.NamespacedName{Namespace: appNamespace, Name: meshRoleBinding.Name}, roleBinding)
+	assertNotFound(err, "Expected RoleBinding to be deleted, but it is still present", t)
+
+	assert.DeepEquals(fakeNetworkStrategy.removedNamespaces, []string{appNamespace}, "Expected removeNamespaceFromMesh to invoke the networkStrategy with only the appNamespace, but it didn't", t)
+}
+
+func TestReconcileUpdatesModifiedRoleBindings(t *testing.T) {
+	t.Skip("Not implemented yet")
```
dgn, line 111: Should we create a MAISTRA issue to add the RoleBinding update functionality and link it here? I think it's a case we should support

dgn (approved):
Good work here, thank you Marko!

The only thing missing in this pkg are now the networkStrategy implementations, and I think those would be better tested through e2e tests - not sure if we already have those in jenkins.

luksa:
I've already written the networkStrategy tests. I'll submit them in a different PR (I need to submit a different PR in between these and the networkStrategy tests).

jwendell (approved):
Thanks. We need somehow to run `make test` for every PR now.

dgn:
> Thanks. We need somehow to run `make test` for every PR now.

Ideally we'll soon have a CI system doing that for us ;) As soon as the Brno machine thing is resolved I'll setup a prow instance for us

jwendell:
@mergify refresh

dgn:
Meh, mergify seems to be broken again. It shouldn't be merging things into this branch

luksa:
@Mergifyio refresh

luksa:
@mergify refresh

PR #285, opened 2020-01-08
WIP: use appropriate types for all resources instead of unstructured.Unstructured

rcernich:

### DIFF
--- a/pkg/controller/common/manifestprocessing.go
+++ b/pkg/controller/common/manifestprocessing.go
@@ -7,8 +7,9 @@ import (
 	"github.com/ghodss/yaml"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	v1 "github.com/maistra/istio-operator/pkg/apis/maistra/v1"
@@ -23,13 +24,13 @@ import (
 
 type ManifestProcessor struct {
 	ResourceManager
-	preprocessObject func(obj *unstructured.Unstructured) error
-	processNewObject func(obj *unstructured.Unstructured) error
+	preprocessObject func(obj runtime.Object) error
+	processNewObject func(obj runtime.Object) error
 
 	appInstance, appVersion, owner string
 }
 
-func NewManifestProcessor(resourceManager ResourceManager, appInstance, appVersion, owner string, preprocessObjectFunc, postProcessObjectFunc func(obj *unstructured.Unstructured) error) *ManifestProcessor {
+func NewManifestProcessor(resourceManager ResourceManager, appInstance, appVersion, owner string, preprocessObjectFunc, postProcessObjectFunc func(obj runtime.Object) error) *ManifestProcessor {
 	return &ManifestProcessor{
 		ResourceManager:  resourceManager,
 		preprocessObject: preprocessObjectFunc,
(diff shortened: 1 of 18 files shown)


rcernich:
Currently, untested...

luksa:
@rcernich I've been working on getting this to run. My fixes are in https://github.com/luksa/istio-operator/commits/pr285-fixes

Some of these fixes are quick hacks/workarounds (especially the commit where I add the `ServiceMeshPolicy` type into the code instead of importing it from github.com/Maistra/istio - I'm not even sure why the type is declared there instead of in istio/api).

With my fixes, the operator now appears to successfully deploy the control plane (there are no errors in the log. I'm trying bookinfo now).

dgn:
> Some of these fixes are quick hacks/workarounds (especially the commit where I add the `ServiceMeshPolicy` type into the code instead of importing it from github.com/Maistra/istio - I'm not even sure why the type is declared there instead of in istio/api).

That's because we introduced it, it's a replacement for the upstream `MeshPolicy` which is cluster-scoped.

luksa:
> That's because we introduced it, it's a replacement for the upstream `MeshPolicy` which is cluster-scoped.

I know. But I'd expect `MeshPolicy` to be in istio/api instead of istio/istio and then `ServiceMeshPolicy` in our own fork of istio/api.

dgn:
> I know. But I'd expect `MeshPolicy` to be in istio/api instead of istio/istio and then `ServiceMeshPolicy` in our own fork of istio/api.

I agree, that would probably make more sense. I'm hoping though that in the near future our API won't diverge anymore and this will be unnecessary

rcernich:
Thanks @luksa for taking this the last mile.  Does it improve performance during reconcile at all, specifically prune?

Also, don't worry about the template fixes, as those are handled in MAISTRA-1062

pkg/controller/common/manifestprocessing.go
```diff
@@ -61,8 +62,7 @@ func (p *ManifestProcessor) ProcessManifests(manifests []manifest.Manifest, comp
 				allErrors = append(allErrors, err)
 				continue
 			}
-			obj := &unstructured.Unstructured{}
-			_, _, err = unstructured.UnstructuredJSONScheme.Decode(rawJSON, nil, obj)
+			obj, err := runtime.Decode(p.JSONSerializer, rawJSON)
```
rcernich, line 65: This should be changed to use 'unstructured.JSONFallbackEncoder'.  That should fix the issues with unregistered resources in the scheme.

rcernich: Whoops, scratch that, that's only for Encoding.

pkg/controller/common/patch.go
```diff
@@ -103,7 +107,7 @@ func (p *PatchFactory) CreatePatch(current, new runtime.Object) (Patch, error) {
 		if err != nil {
 			return nil, err
 		}
-		newObj, err := runtime.Decode(unstructured.UnstructuredJSONScheme, newBytes)
+		newObj, err := runtime.Decode(p.serializer, newBytes)
```
rcernich, line 110: This needs to fallback to unstructured if the type is unknown.

pkg/controller/common/patch.go
```diff
@@ -130,7 +134,7 @@ func (p *PatchFactory) CreatePatch(current, new runtime.Object) (Patch, error) {
...
```
rcernich, line 137: Ditto, fallback to unstructured...

pkg/controller/common/patch.go
rcernich: Line 83 should use our `scheme`, not the generic `runtime.Scheme`.

luksa:
A quick performance test shows that these changes speed up the operator by 3 to 4 times. 

Time spent in controller's Reconcile() function:

**Unstructured**
1st create: 	68s (prune 14s)   
1st delete: 	32s (prune 30s)
2nd create:	57s 
2nd delete:	32s 

**Strongly-typed**
1st create: 	25s (prune 4s)
1st delete: 	10s (prune 9s)
2nd create:	13s
2nd delete:	10s

rcernich:
FYI @ostromart

rcernich:
We might be able to eek out a bit more by integrating the client-go library for istio.  At that point, I think there will only be one or two resource types that still use unstructured.

dgn:
Should we open a new PR from @luksa's branch?

rcernich:
That sounds good to me, although I suspect there will be some issues with the rebase, given all the recent refactoring. :(

jwendell:
@rcernich What's the status of this? Obsolete and safe to close? It targets 1.1.

luksa:
Obsolete. We did most (if not all) of this in other PRs. Closing.

PR #286, opened 2020-01-10
MAISTRA-1057: Add IOR back

jwendell:


rcernich:
This will need to be updated after #273 is merged, as the overlays have moved to resources/helm/overlays.  You'll then need to update the v1.1 charts (`make update-1.1-charts`).

manifests-servicemesh/1.0.0/servicemeshoperator.v1.0.0.clusterserviceversion.yaml
```diff
@@ -43,7 +43,8 @@ metadata:
                   "autoscaleEnabled": false
                 },
                 "istio-ingressgateway": {
-                  "autoscaleEnabled": false
+                  "autoscaleEnabled": false,
+                  "ior_enabled": false
```
dgn: This still has to be removed

manifests-maistra/1.0.0/maistraoperator.v1.0.0.clusterserviceversion.yaml
```diff
@@ -43,7 +43,8 @@ metadata:
                   "autoscaleEnabled": false
                 },
                 "istio-ingressgateway": {
-                  "autoscaleEnabled": false
+                  "autoscaleEnabled": false,
+                  "ior_enabled": false
```
dgn: This still has to be removed

deploy/smcp-templates/servicemesh
```diff
@@ -27,6 +27,9 @@ spec:
       image: citadel-rhel8
     sidecarInjectorWebhook:
       image: sidecar-injector-rhel8
+    gateways:
+      istio-ingressgateway:
+        ior_image: istio-ior-rhel8
```
dgn: I know this is how it used be handled, but is that really something we should put into `istio-ingressgateway`? Maybe we should rethink this while we're at it.

I'd much prefer if we introduced ior as a component just like other Istio components:
```yaml
spec:
  istio:
    ior:
      enabled: false
      image: istio-ior-rhel8
```

dgn (commented):
I think we want to move the manifests into a separate repo- I suppose for that we'll extract them from the `maistra-1.0` branch, so you can likely ignore my comments about the CSVs.

In general I'd prefer it if we would not inject our options into `ingress-gateway` anymore, see my comment below. I think a separate `ior` component block is more in line with the rest of Istio. It's an easy change that only affects the YAMLs, too.

brian-avery (approved):

jwendell:
> This will need to be updated after #273 is merged, as the overlays have moved to resources/helm/overlays. You'll then need to update the v1.1 charts (`make update-1.1-charts`).

@rcernich PR updated

rcernich (approved):

PR #287, opened 2020-01-14
MAISTRA-1081 Allow users to configure the number of worker threads us…

luksa:
…ed by controllers


dgn (approved):
lgtm

pkg/controller/servicemesh/controlplane/controller.go
```diff
@@ -272,12 +272,16 @@ func (r *ReconcileControlPlane) Reconcile(request reconcile.Request) (reconcile.
 }
 
 var reconcilers = map[string]*ControlPlaneReconciler{}
+var mu sync.Mutex
```
rcernich: We only create one `ReconcileControlPlane`, but shouldn't this mutex be contained within that struct?

luksa: The mutex is to protect the `reconcilers` map, that's why it's declared here. But you're right. The map itself should be contained in the struct.

rcernich: Sorry I missed that.

luksa:
@rcernich can I get a whoop whoop? (or at least an approval) :)

rcernich (approved):

PR #288, opened 2020-01-14
MAISTRA-1008 List namespaces instead of relying on SMMR.spec.members

luksa:


pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -379,6 +382,13 @@ func (r *ReconcileMemberList) Reconcile(request reconcile.Request) (reconcile.Re
 	return reconcile.Result{}, kialiErr
 }
 
+func (r *ReconcileMemberList) findConfiguredNamespaces(meshNamespace string) (corev1.NamespaceList, error) {
+	list := corev1.NamespaceList{}
+	labelSelector := map[string]string{common.MemberOfKey: meshNamespace}
```
dgn, line 395: Why use the label and not e.g. the Status.ConfiguredMembers?

dgn (commented):
Good catch, I think looking at the `spec.members` to know which namespaces have to be removed from the mesh is wrong, but wouldn't looking at `status.configuredMembers` be sufficient? I'm probably missing something here though..

I can't help but think that it might be possible to make the operator perform actions on unrelated namespaces just by adding a label to them, which feels wrong

pkg/controller/servicemesh/memberroll/controller.go
luksa: Before this PR, the operator determined configured members in two ways (in two different places):
- lists namespaces with the label
- gets namespaces from `spec.members`.

This PR ensures that the same method is used in both places. 

Using `status.configuredMembers` might result in the operator not cleaning up a namespace if the SMMR object is stale and is missing one of the namespaces. But you've now got me thinking. Even the namespaces themselves can be stale, so you can actually run into the same problem.

rcernich:
Is the thinking here that we have a reconciler for namespaces, which is triggered based on updates to SMMR/SMR/SMCP resources?  Seems cleaner, but I don't know how you would handle removals (i.e. a namespace removed from the mesh, but not itself deleted).

pkg/controller/servicemesh/memberroll/controller.go
dgn: > This PR ensures that the same method is used in both places.

^^ I think that alone makes sense. We can see if we want to change general behaviour later on.

dgn (approved):

brian-avery (approved):

luksa:
@mergify refresh

PR #293, opened 2020-01-17
MAISTRA-1095 Don't rely on execute permission for vendored file

dgn:
When I ran `go mod vendor` it would overwrite the permissions for the executed file, breaking our build. I think this approach is safer.


luksa (approved):
LGTM. `go mod vendor` does the same thing on my machine.

rcernich (approved):

dgn:
@Mergifyio refresh

dgn:
Wtf.. Mergify made a mess here

PR #298, opened 2020-01-17
Remove the jq argument '--indentless': it does not exist

jwendell:


dgn:
It was introduced in 2.8.0: https://yq.readthedocs.io/en/latest/changelog.html?highlight=indentless#changes-for-v2-8-0-2019-10-25

jwendell:
hahaha sorry, I need coffee! I was thinking about Jq !!

PR #301, opened 2020-01-20
MAISTRA-1101: Remove egress from NetworkPolicy

knrc:


rcernich (approved):
LGTM.  However, should we consider doing something similar to what we do for ingress, like MIASTRA-964?

knrc:
@rcernich I don't believe so, the egress behaviour is controlled by the service mesh configuration and this can force all communication through the egress gateway if necessary.  It's more important to force ingress via ingressgateway since otherwise the incoming requests miss the client side features.

brian-avery (approved):

PR #304, opened 2020-01-21
MAISTRA-1112 Update 1.1 operator to support Maistra 1.0.4

luksa:


jwendell (approved):

dgn (approved):
I said it before: I think we should initialize this value from the code, so that we can change it in a single place when we release a new z-stream update. Will look at that if I have time later today

rcernich (approved):

brian-avery (approved):

PR #305, opened 2020-01-21
MAISTRA-1102: Add support for discovering Calico networking

knrc:


rcernich (approved):

jwendell (approved):

jwendell:
Should this be cherry-picked to 1.1 as well?

rcernich:
Yes, we need to make sure all of these get into 1.1 too.

knrc:
Perhaps not cherry picked, that depends on the work Rob/Marko have already done on that branch, but it should certainly have an equivalent change made into 1.1.  That goes for all the 1.0.5 commits related to this work.

PR #309, opened 2020-01-23
feat: sets up up basic CI build

bartoszmajsak:
This PR brings setup of a build running tests on every PR and merge to master using CirleCI free tier. Additionally, it fixes a little glitch in `Makefile` which assumes that there's always fork and upstream remote defined for the cloned git repository, which is not the case when running as the CircleCI job.

You can see the build on master in my fork: https://github.com/bartoszmajsak/istio-operator/commits/maistra-1.1

and the PR with the failing tests - https://github.com/bartoszmajsak/istio-operator/pull/1

CircleCI is enabled for Maistra organization on GitHub so after merging this PR, the repository needs to be enabled on CircleCI side (which is one click resulting in webhook config for that repo). Happy to do that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,9 @@ OFFLINE_BUILD       ?= false
 GIT_UPSTREAM_REMOTE ?= $(shell git remote -v |grep --color=never ':Maistra/istio-operator\.git.*(fetch)' |grep --color=never -o '^[^[:space:]]*')
 
 ifeq "${GIT_UPSTREAM_REMOTE}" ""
-$(error Could not find git remote for Maistra/istio-operator)
+$(warning Could not find git remote for Maistra/istio-operator. Adding default one)
+$(shell git remote add upstream git@github.com:Maistra/istio-operator.git)
+GIT_UPSTREAM_REMOTE = upstream
 endif
 
 ifeq "${COMMUNITY}" "true"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,16 @@
+version: 2.1
+jobs:
+
+  build:
+    working_directory: /go/src/github.com/maistra/istio-operator
+    docker:
+      - image: circleci/golang:1.13.6
+    steps:
+      - checkout
+      - run: make clean compile test
+
+workflows:
+  version: 2.1
+  circleci_build:
+    jobs:
+      - build


bartoszmajsak:
Another tool which works pretty well for us ([istio-workspace](https://github.com/Maistra/istio-workspace)) and I would be happy to contribute here too is golangci - configurable linter for Go projects (see our config [here](https://github.com/Maistra/istio-workspace/blob/d3146177f0d05c8bac8eee5a677b35cad9fd40e4/.golangci.yml)). 

It not only fails the build when certain smells are detected but comments on suspicious lines and suggests changes. This sets a reviewer free from looking for such smells and let them focus o the important things.

Makefile
```diff
@@ -31,7 +31,9 @@ OFFLINE_BUILD       ?= false
 GIT_UPSTREAM_REMOTE ?= $(shell git remote -v |grep --color=never ':Maistra/istio-operator\.git.*(fetch)' |grep --color=never -o '^[^[:space:]]*')
 
 ifeq "${GIT_UPSTREAM_REMOTE}" ""
-$(error Could not find git remote for Maistra/istio-operator)
+$(warning Could not find git remote for Maistra/istio-operator. Adding default one)
+$(shell git remote add upstream git@github.com:Maistra/istio-operator.git)
```
rcernich, line 35: Should this check to make sure `upstream` doesn't exist already?

That said, this is primarily used to update the legacy charts (`git checkout ${GIT_UPSTREAM_REMOTE}/maistra-1.0 -- resources/helm/v1.0`), so if all we want to do is compile/test, there is an `OFFLINE_BUILD` option, which maybe we could use to prevent git operations.

bartoszmajsak: Good catch. We could check, or assign another name for that - e.g. `circle_ci_upstream`.

Another option would be to move that logic to targets that rely on this value and not making it a global pre-requisite. WDYT?

rcernich: All of those sound good.

In addition, I think the update-1.0-remote target should fail in offline mode, which means the upstream remote logic could just be moved into that target.  This probably makes the most sense, i.e. don't error out just because it's not set, but only error out the target's that actually use it.

FWIW, I implemented that as a global so folks could override it via command line.  That said, I don't know if it's necessary, given the auto-setting.

bartoszmajsak: Do you want me to do that in this PR or open a new one?

luksa (approved):

bartoszmajsak:
I guess this can be closed now since `prow` is in place.

PR #310, opened 2020-01-24
MAISTRA-1124 Add ControlPlaneReconciler unit tests

luksa:


rcernich (approved):

pkg/controller/servicemesh/controlplane/controller_test.go
```diff
@@ -1,104 +1,200 @@
...
+func (r *fakeInstanceReconciler) Delete() error {
+	r.deleteInvoked = true
+	return nil
+}
+
+func (r *fakeInstanceReconciler) SetInstance(instance *maistrav1.ServiceMeshControlPlane) {
+}
+
+func (r *fakeInstanceReconciler) IsFinished() bool {
+	return r.finished
```
dgn, line 166: Do we ever set this to `true` in the fake reconciler? It's not super important here, but the `defer` calls that are supposed to clean up probably don't do much

dgn (approved):
Thanks, I appreciate the refactoring. Looks cleaner now

pkg/controller/servicemesh/controlplane/controller_test.go
luksa: Good catch. I added the `finished` field because I planned to also test `getOrCreateReconciler()` and `deleteReconcilerIfFinished()`. Then I apparently forgot to add those tests. I'll merge this PR now and add those tests when I get back on Friday.

dgn: Sounds good. Enjoy your time off!

PR #311, opened 2020-01-24
MAISTRA-1127 Fix inconsistent controller names in logs and events

luksa:


pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -31,7 +31,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-const controllerName = "controller_servicemeshmemberroll"
+const controllerName = "servicemeshmemberroll-controller"
```
luksa, line 34: Should we use acronyms (e.g. "smmr-controller") here? If we use the full name, the log lines are very long even before you get to the actual log message.

luksa: Oh, but the same name is also used in events. I'm not a big fan of using acronyms there. But maybe that's not a bad idea. Users working with the Maistra API will be familiar with the acronyms, since they'll probably use them in oc/kubectl commands.

pkg/controller/servicemesh/validatingwebhook/controller.go
```diff
@@ -24,8 +24,10 @@ import (
 	"github.com/pkg/errors"
 )
 
+const controllerName = "validatingwebhook-controller"
```
rcernich, line 27: I think `galley-webhook-controller` might be more descriptive.

pkg/controller/servicemesh/mutatingwebhook/controller.go
```diff
@@ -24,8 +24,10 @@ import (
 	"github.com/pkg/errors"
 )
 
+const controllerName = "mutatingwebhook-controller"
```
rcernich, line 27: I think something like `injector-webhook-controller` might be more descriptive.

rcernich (approved):

pkg/controller/servicemesh/memberroll/controller.go
dgn: Let's stick with the long names for now; otherwise we create inconsistencies with e.g. `validatingwebhook-controller`

dgn (approved):

PR #314, opened 2020-01-27
create fake implementations for manager.Manager related interfaces

rcernich:


rcernich:
This seems to be a good start of a "fake" controller manager.

I've also refactored @luksa's `EnhancedTracker` to use the existing client-go/testing `Fake` implementation, which slightly changed the signatures used with `Reactors`.

In addition to that, I've reworked `FakeClient` to also use the `Fake` interface, which allows for more customized reactions (e.g. the reactors used with the `FakeManager` simulate actual create, delete, and update reactions in the api server by initializing certain fields and preventing deletion when finalizers are present.

I'll focus on scripting out some actual manager tests next.

Lastly, the idea around this is to create integration tests that simulate actual installs, etc.

rcernich:
moving this back to WIP, as there are still a couple of bugs in the underlying watch code (e.g. controller creates unstructured types, but watches for specific types, like `v1.Deployment`).  I'd also like to see if I can get the event recorder to use the same `ObjectTracker` that's used by everything else.  (The core `fake.ClientSet` creates its own `ObjectTracker` and `Fake` action handler.  It would be great if this matched what's used everywhere else, which would allow us to include recorded events in verification/scripting.)

rcernich:
This should be good to go now.  Only a couple of things to add (eventually):
* Assert controller queue is empty upon test completion
* Assert no events go through object tracker after last expected ActionVerification
* Investigate using object conversion when adding resources to the ObjectTracker (e.g. v1beta1.Deployment -> v1.Deployment)

rcernich:
Look at [integration_test.go](https://github.com/Maistra/istio-operator/pull/314/files#diff-a80818cf582e1fa78b0a413b4d7b0e0e) for an example of a controller test case using the new framework.

rcernich:
This PR is ready to go.  I don't plan on making any further changes to the framework.  Please give it a look and let me know what you think.

rcernich:
Oh, out of my list of future work above...

> Assert controller queue is empty upon test completion

This will be necessary to ensure the controller is idle before injecting a new event into the system (e.g. when the SMCP controller is paused, inject a status update for the deployment that's holding things up).

> Assert no events go through object tracker after last expected ActionVerification

This has been addressed.  The test runner waits for `Reconcile()` to complete before processing assertions and will generate an assertion if an action is processed by the framework after the last ActionVerification was processed.  This means the last verifier should verify the last action expected by the reconciile (e.g. update status).  This ensures the last expected action did occur _and_ no other actions were seen after that action.

> Investigate using object conversion when adding resources to the ObjectTracker (e.g. v1beta1.Deployment -> v1.Deployment)

This will be required to ensure the watches are getting updates for resources correctly, e.g. Istio 1.1 uses v1beta1.extensions.Deployment in a lot of places, but SMCP controller watches v1.apps.Deployment.

cmd/manager/main.go
```diff
@@ -49,12 +49,15 @@ func main() {
...
+	pflag.IntVar(&common.Options.ControlPlaneReconcilers, "controlPlaneReconcilers", 1, "The number of concurrent reconcilers for ServiceMeshControlPlane resources")
+	pflag.IntVar(&common.Options.MemberRollReconcilers, "memberRollReconcilers", 1, "The number of concurrent reconcilers for ServiceMeshMemberRoll resources")
+	pflag.IntVar(&common.Options.MemberReconcilers, "memberReconcilers", 1, "The number of concurrent reconcilers for ServiceMeshMember resources")
 
 	// custom flags for istio operator
-	pflag.StringVar(&common.ResourceDir, "resourceDir", "/usr/local/share/istio-operator", "The location of the resources - helm charts, templates, etc.")
+	pflag.StringVar(&common.Options.ResourceDir, "resourceDir", "/usr/local/share/istio-operator", "The location of the resources - helm charts, templates, etc.")
+	pflag.StringVar(&common.Options.ChartsDir, "chartsDir", "", "The root location of the helm charts.")
+	pflag.StringVar(&common.Options.DefaultTemplatesDir, "defaultTemplatesDir", "", "The root location of the default templates.")
+	pflag.StringVar(&common.Options.UserTemplatesDir, "userTemplatesDir", "", "The root location of the user supplied templates.")
```
dgn, line 60: Maybe we should add "defaults to ${resourceDir}/templates if unset" to this and the other two variables?

luksa (approved):
LGTM, although I'm wondering if we should run actual integration tests against a real API server.

rcernich:
Possibly, although this requires less infrastructure and can be done easily by dev's, as it doesn't require an api server.  We should also be able to rewrite the existing tests to use this framework, as it allows you to verify everything that comes out of the reconciler, with the exception of recorded events and the results.  It also provides the option of testing interactions between the reconcilers (e.g. member roll and control plane).  I think of it as more black box testing than integration testing, as it only simulates the api server enough to get events flowing properly.

dgn (approved):

cmd/manager/main.go
rcernich: Yeah, that's actually handled here: https://github.com/Maistra/istio-operator/pull/314/commits/c91efd378bb07baa43795d1af885df9686f70916#diff-cfd100c62a1053ce78f239a61be8ab6aR65

dgn: Yup, I just meant we should maybe document that behavior

PR #316, opened 2020-01-27
MAISTRA-1129 reconcileNamespaces should return all configuredNamespaces

dgn:
The way we're using the return value of reconcileNamespaces, we will have to pass in all to-be-reconciled namespaces, always. Note that this introduces some overhead - it will reconcile all namespaces always, even if some don't have to be reconciled - but we can fix that in a later refactoring.

### DIFF
--- a/pkg/controller/servicemesh/memberroll/controller.go
+++ b/pkg/controller/servicemesh/memberroll/controller.go
@@ -351,7 +351,7 @@ func (r *MemberRollReconciler) Reconcile(request reconcile.Request) (reconcile.R
 	} else if len(unconfiguredMembers) > 0 { // required namespace that was missing has been created
 		reqLogger.Info("Reconciling newly created namespaces associated with this ServiceMeshMemberRoll")
 
-		newConfiguredMembers, err, nsErrors = r.reconcileNamespaces(unconfiguredMembers, nil, instance.Namespace, meshVersion, reqLogger)
+		newConfiguredMembers, err, nsErrors = r.reconcileNamespaces(requiredMembers, nil, instance.Namespace, meshVersion, reqLogger)
 		if err != nil {
 			return reconcile.Result{}, err
 		}
@@ -404,6 +404,10 @@ func (r *MemberRollReconciler) findConfiguredNamespaces(meshNamespace string) (c
 }
 
 func (r *MemberRollReconciler) reconcileNamespaces(namespacesToReconcile, namespacesToRemove sets.String, controlPlaneNamespace string, controlPlaneVersion string, reqLogger logr.Logger) (configuredMembers []string, err error, nsErrors []error) {
+	// current configuredNamespaces are namespacesToRemove minus control plane namespace
+	configured := sets.NewString(namespacesToRemove.List()...)
+	configured.Delete(controlPlaneNamespace)
+
 	// create reconciler
 	reconciler, err := r.namespaceReconcilerFactory(r.Client, reqLogger, controlPlaneNamespace, controlPlaneVersion, common.IsCNIEnabled)
 	if err != nil {
@@ -418,6 +422,8 @@ func (r *MemberRollReconciler) reconcileNamespaces(namespacesToReconcile, namesp
 		err = reconciler.removeNamespaceFromMesh(ns)
 		if err != nil {
 			nsErrors = append(nsErrors, err)
+		} else {
+			configured.Delete(ns)
 		}
 	}
 	for ns := range namespacesToReconcile {
@@ -434,9 +440,10 @@ func (r *MemberRollReconciler) reconcileNamespaces(namespacesToReconcile, namesp
 				nsErrors = append(nsErrors, err)
 			}
 		} else {
-			configuredMembers = append(configuredMembers, ns)
+			configured.Insert(ns)
 		}
 	}
+	configuredMembers = configured.List()
 	return configuredMembers, nil, nsErrors
 }
 
--- a/pkg/controller/servicemesh/memberroll/controller_test.go
+++ b/pkg/controller/servicemesh/memberroll/controller_test.go
@@ -219,7 +219,7 @@ func TestReconcileReconcilesAfterOperatorUpgradeFromV1_0(t *testing.T) {
 	roll.Spec.Members = []string{appNamespace}
 	roll.Status.ConfiguredMembers = []string{appNamespace}
 	controlPlane := markControlPlaneReconciled(newControlPlane(meshVersion1_0), meshVersion1_0, operatorVersionDefault)
-	namespace := newAppNamespace()
+	namespace := newNamespace(appNamespace)
 	common.SetLabel(namespace, common.MemberOfKey, controlPlaneNamespace)
 	meshRoleBinding := newMeshRoleBinding()
 	appRoleBinding := newMeshRoleBinding()
@@ -306,7 +306,7 @@ func TestReconcileReconcilesAddedMember(t *testing.T) {
 				}
 			}
 			roll.Status.ServiceMeshGeneration = controlPlane.Status.ObservedGeneration
-			namespace := newAppNamespace()
+			namespace := newNamespace(appNamespace)
 			meshRoleBinding := newMeshRoleBinding()
 
 			cl, _, r, nsReconciler, kialiReconciler := createClientAndReconciler(t, roll, controlPlane, namespace)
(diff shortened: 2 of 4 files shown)


rcernich:
Would it make sense to have a single controller for managing members?  I think the member roll controller could be modified to watch for ServiceMeshMember resources, triggering a reconcile of the roll.  The list of required members would be the union of the SMMR.spec.members and the list of SMM resources referencing the control plane.  We could register a custom indexer to make retrieving the list of SMM resources easier (i.e. list member resources with a specific control plane reference).

dgn:
> Would it make sense to have a single controller for managing members? I think the member roll controller could be modified to watch for ServiceMeshMember resources, triggering a reconcile of the roll. The list of required members would be the union of the SMMR.spec.members and the list of SMM resources referencing the control plane. We could register a custom indexer to make retrieving the list of SMM resources easier (i.e. list member resources with a specific control plane reference).

Yeah, I agree that might make sense- but maybe we do that for 1.2? We should have a discussion post-1.1 on how we want to proceed with the operator, architecturally.

For now, this really just fixes the bug I found

luksa:
+1 on doing this in 1.2. I intentionally implemented the member controller separately in order to reduce the chance of breaking the existing memberroll controller (and before it was clear whether using the Member resource will be the preferred way of adding namespaces to the mesh). That's why I agree we should leave it as is in 1.1 - to see if there are any issues with the Member resource itself - and merge the two controllers in 1.2.

pkg/controller/servicemesh/memberroll/controller_test.go
```diff
@@ -750,10 +777,10 @@ func markControlPlaneReconciled(controlPlane *maistra.ServiceMeshControlPlane, m
 	return controlPlane
 }
 
-func newAppNamespace() *core.Namespace {
+func newAppNamespace(name string) *core.Namespace {
```
luksa: This can now be renamed to `newNamespace`.

pkg/controller/servicemesh/memberroll/controller_test.go
```diff
@@ -464,6 +464,33 @@ func TestReconcileNamespacesIgnoresControlPlaneNamespace(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("reconcileNamespaces failed: %v", err)
+	}
+	if len(nsErrors) > 0 {
+		t.Fatalf("reconcileNamespaces returned unexpected nsErrors: %v", nsErrors)
+	}
+
+	assertNamespaceReconcilerInvoked(t, nsReconciler, appNamespace, appNamespace2) // NOTE: no controlPlaneNamespace
+	assert.DeepEquals(configuredMembers, []string{appNamespace, appNamespace2}, "reconcileNamespaces returned an unexpected configuredMembers list", t)
+}
```
luksa: If I revert the changes in controller.go, this test doesn't fail. The key change seems to be in `Reconcile()`, so I'd expect the test to call it instead of `reconcileNamespaces`. What was the exact cause of the infinite loop? I can't figure it out just by looking at the log/code.

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -436,9 +440,10 @@ func (r *MemberRollReconciler) reconcileNamespaces(namespacesToReconcile, namesp
 				nsErrors = append(nsErrors, err)
 			}
 		} else {
-			configuredMembers = append(configuredMembers, ns)
+			configured.Insert(ns)
 		}
 	}
+	configuredMembers = configured.List()
```
luksa, line 446: I don't think I understand these changes. The `reconcileNamespaces` function must return a list of namespaces that are configured after the function has done its job, as this list is then written to `SMMR.status.configuredMembers`. With these changes, the function also adds `namespacesToRemove` to the list. 

Oh, did you forget to add `configured.Delete(ns)` after the successful invocation of `removeNamespaceFromMesh()`?

dgn: Oh. Yes, I had that in a previous version. I squashed commits before pushing and something must have gone wrong

pkg/controller/servicemesh/memberroll/controller_test.go
dgn: You're right. I pushed a weird mix of two different approaches to fixing the problem; first I thought  I could fix it just in this function, but later I realized that the way we're using the return value from it is also wrong. I pushed the correct test

pkg/controller/servicemesh/memberroll/controller_test.go
dgn: Done

dgn:
> @dgn: The following test **failed**, say `/retest` to rerun all failed tests:
> unittests 	[7233f96](https://github.com/Maistra/istio-operator/commit/7233f96ab18985a2ec00e997e2bcc951e18d6d6d) 	[link](https://78.47.176.69/log?job=unittests&id=1222110191462387712) 	`/test unittests`

Note that the tests will fail until #317 is merged

pkg/controller/servicemesh/memberroll/controller_test.go
```diff
@@ -464,6 +464,33 @@ func TestReconcileNamespacesIgnoresControlPlaneNamespace(t *testing.T) {
 	assert.DeepEquals(configuredMembers, []string{appNamespace}, "reconcileNamespaces returned an unexpected configuredMembers list", t)
 }
 
+func TestReconcileWorksWithMultipleNamespaces(t *testing.T) {
+	roll := newDefaultMemberRoll()
+	addOwnerReference(roll)
+	roll.Spec.Members = []string{appNamespace, appNamespace2}
+	roll.ObjectMeta.Generation = 2
+	roll.Status.ObservedGeneration = 1
+	controlPlane := markControlPlaneReconciled(newControlPlane(""), meshVersionDefault, operatorVersionDefault)
```
luksa: Nit: moving this line to the top would make this setup code a bit cleaner (create the controlePlane, then the roll).

pkg/controller/servicemesh/memberroll/controller_test.go
```diff
@@ -464,6 +464,33 @@ func TestReconcileNamespacesIgnoresControlPlaneNamespace(t *testing.T) {
...
+	roll.Status.ServiceMeshGeneration = controlPlane.Status.ObservedGeneration
+	roll.Status.MeshVersion = controlPlane.Status.AppliedVersion
+	namespace := &core.Namespace{
+		ObjectMeta: meta.ObjectMeta{
+			Name: controlPlaneNamespace,
+		},
+	}
```
luksa: Could use `newNamespace()`, but you may not need to create this namespace at all.

pkg/controller/servicemesh/memberroll/controller_test.go
```diff
@@ -464,6 +464,33 @@ func TestReconcileNamespacesIgnoresControlPlaneNamespace(t *testing.T) {
...
+
+	cl, _, r, _, kialiReconciler := createClientAndReconciler(t, roll, controlPlane, namespace, newNamespace(appNamespace))
+	assertReconcileSucceeds(r, t)
+	cl.Create(context.TODO(), newNamespace(appNamespace2))
```
luksa: Unhandled error. While the error itself is probably never actually returned because we're using the fake client, IMHO it's always good to handle errors. If nothing else, the static analysis tools won't complain about unhandled errors.

I have another PR I'm about to submit, where I introduce `test.SuccessOrPanic()`, which, in the above case, would be used like this: 
```
test.SuccessOrPanic(cl.Create(context.TODO(), newNamespace(appNamespace2))
```

luksa (approved):
A few minor comments on polishing up the test code, but otherwise LGTM.

pkg/controller/servicemesh/memberroll/controller_test.go
dgn: Removed

pkg/controller/servicemesh/memberroll/controller_test.go
dgn: Done

pkg/controller/servicemesh/memberroll/controller_test.go
dgn: I'm using `test.PanicOnError()` now

luksa (approved):

rcernich (approved):

PR #321, opened 2020-01-31
MAISTRA-1126 Remove global log vars to ensure correct logger is used

luksa:


rcernich (approved):
lgtm.  I wonder if we should push a similar change up to code templates in the operator-sdk, to help promote better encapsulation from the start.

brian-avery (approved):

luksa:
> lgtm. I wonder if we should push a similar change up to code templates in the operator-sdk, to help promote better encapsulation from the start.

I'm working on another set of changes, which improve things even further. Maybe we push things upstream when I'm done.

PR #324, opened 2020-02-04
MAISTRA-1143 Propagate context properly throughout the call stack

luksa:


dgn (approved):

pkg/controller/common/manifestprocessing.go
```diff
@@ -40,35 +40,36 @@ func NewManifestProcessor(controllerResources ControllerResources, appInstance,
...
 }
 
-func (p *ManifestProcessor) ProcessManifests(manifests []manifest.Manifest, component string) error {
+func (p *ManifestProcessor) ProcessManifests(ctx context.Context, manifests []manifest.Manifest, component string) error {
 	allErrors := []error{}
 
-	origLogger := p.Log
-	defer func() { p.Log = origLogger }()
+	origCtx := ctx
+	origLogger := LogFromContext(ctx)
```
rcernich: I don't think this is necessary, if we redeclare ctx in the loop below.

pkg/controller/common/manifestprocessing.go
```diff
@@ -40,35 +40,36 @@ func NewManifestProcessor(controllerResources ControllerResources, appInstance,
...
 	for _, man := range manifests {
-		p.Log = origLogger.WithValues("manifest", man.Name)
+		log := origLogger.WithValues("manifest", man.Name)
+		ctx = NewContextWithLog(origCtx, log)
```
rcernich: `ctx := NewContextWithLog(ctx, LogFromContext(ctx).WithValues("manifest", man.Name))` should do the trick.

And, generally speaking, for all cases where we create a subcontext, should this be followed by, `defer func() { ctx.Done() }()`?  (except in this case, where we're in a loop, which is a bit trickier)

pkg/controller/common/manifestprocessing.go
```diff
@@ -78,22 +79,22 @@ func (p *ManifestProcessor) ProcessManifests(manifests []manifest.Manifest, comp
...
-func (p *ManifestProcessor) processObject(obj *unstructured.Unstructured, component string) error {
-	origLogger := p.Log
-	defer func() { p.Log = origLogger }()
+func (p *ManifestProcessor) processObject(ctx context.Context, obj *unstructured.Unstructured, component string) error {
+	origLogger := LogFromContext(ctx)
 
 	key := v1.NewResourceKey(obj, obj)
-	p.Log = origLogger.WithValues("Resource", key)
+	log := origLogger.WithValues("Resource", key)
+	ctx = NewContextWithLog(ctx, log)
```
rcernich: Can we simplify this to `ctx = NewContextWithLog(ctx, LogFromContext(ctx).WithValues("Resource", key)`?

pkg/controller/common/manifestprocessing.go
luksa: No, the local log variable needs to include the "manifest" value, too - there are log statements in this function that need to use the same log as the one we're injecting into the Context. If I use your trick, then I'd have to get the logger from the new context again.

We don't need to call `ctx.Done()`. as it only returns a channel that is used to cancel long-running operations - the Client calls it when performing external calls so that they can be cancelled if someone cancels the context. There is no other function to mark the context "done".

rcernich (approved):

pkg/controller/common/manifestprocessing.go
luksa: It wasn't possible directly, but I've now refactored the code.

pkg/controller/common/manifestprocessing.go
luksa: I had to extract a method to get rid of this. IMHO, it makes sense to extract a function every time a section of code has its own logger and context. Interesting how logging can drive the structuring of code.

rcernich (approved):

PR #325, opened 2020-02-04
MAISTRA-1108 Operator fails to create 1.1 CRDs on OpenShift 3.11

luksa:


luksa:
Note that this PR also contains #324. Please review only the MAISTRA-1108 commit.

pkg/bootstrap/crds.go
```diff
@@ -129,6 +130,7 @@ func processCRDFile(ctx context.Context, cl client.Client, fileName string) erro
 			if errors.IsNotFound(err) {
 				log.Info("creating CRD", "file", fileName, "index", index, "CRD", obj.GetName())
 				err = cl.Create(ctx, obj)
+				err = hacks.WorkAroundTypeObjectProblemInCRDSchemas(ctx, err, cl, obj)
```
rcernich, line 133: Could we be proactive, by reading the underlying k8s version and doing the right thing?

luksa: Yeah, not sure about that due to a few reasons:
- how do we know which (future) version of OpenShift will handle this properly (OpenShift could be patched specifically to fix this problem, so the underlying k8s version might not matter)
- theoretically, the API server can be upgraded without restarting the operator; we couldn't just check the version at operator startup
- it may also be possible that the operator hits a different version of the API server on every request (e.g. during an upgrade of an H/A installation of the API Server, some instances may run an older version, while others run a new one; the load balancer in front of the API server thus routes the operator's requests to different API server versions).

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -79,11 +82,11 @@ func NewControlPlaneInstanceReconciler(controllerResources common.ControllerReso
 	}
 }
 
-func (r *controlPlaneInstanceReconciler) Reconcile() (result reconcile.Result, err error) {
+func (r *controlPlaneInstanceReconciler) Reconcile(ctx context.Context) (result reconcile.Result, err error) {
```
rcernich: Should we be creating a new child context here?

pkg/controller/common/manifestprocessing.go
```diff
@@ -43,28 +43,29 @@ func NewManifestProcessor(controllerResources ControllerResources, appInstance,
 func (p *ManifestProcessor) ProcessManifests(ctx context.Context, manifests []manifest.Manifest, component string) error {
 	allErrors := []error{}
 
-	origLogger := p.Log
-	defer func() { p.Log = origLogger }()
+	origCtx := ctx
+	origLogger := LogFromContext(ctx)
```
rcernich: Would it be better to create a new child context, which includes the new logger?

pkg/bootstrap/crds.go
rcernich: Understood.  Maybe we could set a flag on the first error, so we don't have to do the error processing on every CRD?

dgn: I think it's fine as it is - we will only do the comparison if `cl.Create()` returned an error, which should be rare enough

pkg/controller/hacks/hacks.go
```diff
@@ -2,8 +2,12 @@ package hacks
```
dgn, line 2: I'm a big fan of this package

pkg/controller/hacks/hacks.go
```diff
@@ -19,3 +23,37 @@ func ReduceLikelihoodOfRepeatedReconciliation(ctx context.Context) {
...
 }
+
+// WorkAroundTypeObjectProblemInCRDSchemas works around the problem where OpenShift 3.11 doesn't like "type: object"
+// in CRD OpenAPI schemas. This function removes all occurrences from the schema.
+func WorkAroundTypeObjectProblemInCRDSchemas(ctx context.Context, err error, cl client.Client, crd *unstructured.Unstructured) error {
+	log := common.LogFromContext(ctx)
+	if err != nil && strings.Contains(err.Error(), "must only have \"properties\", \"required\" or \"description\" at the root if the status subresource is enabled") {
+		log.Info("The API server rejected the CRD. Removing type:object fields from the CRD schema and trying again.")
+
+		schema, found, err := unstructured.NestedFieldNoCopy(crd.UnstructuredContent(), "spec", "validation", "openAPIV3Schema")
```
dgn, line 34: Should we be handling the `!found && err==nil` case here? IIUC that should never happen, but if it does for whatever reason, we'll not output an error but also not create the CRD

dgn (approved):
not sure we should be working on 3.11 clusters, but I'm OK with adding this

pkg/controller/servicemesh/controlplane/reconciler.go
luksa: I don't see a reason to create one. Currently, we only create a new Context when we add fields the logger. I can imagine we'll want to create new Contexts when we want to be able to cancel them. But that's not the case yet.

pkg/controller/common/manifestprocessing.go
luksa: Fixed in #324

pkg/controller/hacks/hacks.go
luksa: In that case, the original error is returned and is logged.

I've already changed this code in a different PR I'm about to submit. I've had to move the `cl.Create` call out of this function, as I need to call `cl.Update` in one different case.

dgn: Oh, you're right, I misread that

luksa:
@rcernich can you approve this? I need it to get merged before I submit the next PR.

rcernich (approved):

PR #330, opened 2020-02-06
MAISTRA-1156 Add Make targets to check whether generated files have been checked in

dgn:
This is a first step towards some more PR safety. It checks whether you checked in all changes to charts and profiles.

Note that this currently does not
- regenerate manifests and check changes: they change the `createdAt` date everytime. I tried filtering it with `git diff -G"^crazyRegex$"` but no luck so far
- update the 1.0 charts: I don't want checks to fail just because the `maistra-1.0` branch has changed. We'll need a strategy for that, but until we have it, we just do `update-1.1-charts`


jwendell (approved):

dgn:
I will also add a prow job for this, however I will have to update the build container first, it seems we require `unzip` for the `update-charts` target which is not currently installed

luksa (approved):

PR #341, opened 2020-02-19
MAISTRA-1180: Update network policy to fix intermittent ingress gatew…

knrc:
…ay routing errors

### DIFF
--- a/helm/istio/charts/gateways/templates/networkpolicy.yaml
+++ b/helm/istio/charts/gateways/templates/networkpolicy.yaml
@@ -16,7 +16,6 @@ spec:
     matchLabels:
       istio: ingressgateway
   ingress:
-  - from:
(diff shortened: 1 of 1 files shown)


rcernich (approved):
lgtm, although I don't see how this is any different from the original.

jwendell (approved):
I think this is the rationale: https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-allow-all-ingress-traffic

brian-avery (approved):

knrc:
@jwendell No, that's not the rationale.  This aligns the definition with other NPs such as the kiali one but it also allows access to the ingress gateway from router traffic coming from other nodes if the router has been installed with hostNetwork.  Without this change you would need the load balancer to hit the router on same node as the ingress gateway pod.

PR #342, opened 2020-02-19
Additional improvements for SMMR with large number of members

luksa:


rcernich (approved):

rcernich (approved):

pkg/controller/common/helm.go
```diff
@@ -43,6 +43,12 @@ type options struct {
 	ControlPlaneReconcilers int
 	MemberRollReconcilers   int
 	MemberReconcilers       int
+
+	// The number of API requests the operator can make before throttling
+	Burst int
```
dgn, line 48: I'm not sure I like that we're using this file to dump globals.. But we can fix that in 1.2

pkg/controller/servicemesh/webhooks/validation/memberroll.go
```diff
@@ -107,32 +117,103 @@ func (v *MemberRollValidator) Handle(ctx context.Context, req atypes.Request) at
...
-			oldSmmr := &maistrav1.ServiceMeshMemberRoll{}
-			err := v.decoder.DecodeRaw(req.AdmissionRequest.OldObject, oldSmmr)
-			if err != nil {
-				logger.Error(err, "error decoding old object in admission request")
-				return admission.ErrorResponse(http.StatusBadRequest, err)
-			}
-			namespacesToCheck.Delete(oldSmmr.Spec.Members...)
+		namespacesToCheck, err := v.findNewlyAddedNamespaces(smmr, req)
+		if err != nil {
+			return admission.ErrorResponse(http.StatusInternalServerError, err)
```
dgn: Should we log the error in this case? Also, returning a UnprocessableEntity error might give more context than internal server error, but I'm not sure what the standard behaviour is for k8s webhooks

pkg/controller/servicemesh/webhooks/validation/memberroll.go
```diff
@@ -107,32 +117,103 @@ func (v *MemberRollValidator) Handle(ctx context.Context, req atypes.Request) at
...
-			}
-			if !allowed {
-				return validationFailedResponse(http.StatusForbidden, metav1.StatusReasonBadRequest, fmt.Sprintf("user '%s' does not have permission to access project/namespace '%s'", req.AdmissionRequest.UserInfo.Username, member))
+		allowed, rejectedNamespaces, err := v.isUserAllowedToUpdatePodsInAllNamespaces(ctx, req, namespacesToCheck)
+		if err != nil {
+			if errors.Is(err, context.DeadlineExceeded) {
+				logger.Error(err, fmt.Sprintf("timeout while performing SAR checks for %d namespaces", len(namespacesToCheck)))
+				return admission.ErrorResponse(http.StatusBadRequest, fmt.Errorf("too many namespaces in ServiceMeshMemberRoll; validating webhook couldn't perform the authorization checks for all namespaces; either try the operation again as a cluster admin, or add fewer namespaces in a single operation"))
 			}
+			return admission.ErrorResponse(http.StatusInternalServerError, err)
```
dgn, line 132: We probably want to log the error again here, will make future debugging easier

dgn (approved):
Looks good, just a few minor comments

pkg/controller/common/helm.go
rcernich: Agreed.  I think originally, the only options were specific to helm rendering (e.g. chart directory), but it has since expanded (obviously).

luksa: +1

pkg/controller/servicemesh/webhooks/validation/memberroll.go
luksa: You're right. This should be a 400 BadRequest, since the error occurs only when the oldObject can't be decoded. 

But because this is a client error, it shouldn't be logged as an error. The user gets the error message in the response and the cluster-admin that manages the operator doesn't want to see client errors in the log. Right?

pkg/controller/servicemesh/webhooks/validation/memberroll.go
luksa: You're absolutely right. I've fixed this here and in one other place, where we also returned a 500 error.

pkg/controller/servicemesh/webhooks/validation/memberroll.go
dgn: Ah, right. No logging then!

rcernich: Good question, but this is an internal error, not a client error.  Client isn't sending oldObject, the apiserver is.  That said, there's not much we can do about it either, so does it need to be logged, and as you said, the user gets the the error?

PR #345, opened 2020-02-20
UX improvements

luksa:


manifests-maistra/1.1.0/servicemeshmemberrolls.crd.yaml
```diff
@@ -16,7 +16,22 @@ spec:
...
+    description: How many of the total number of member namespaces are
+      configured
+    type: string
+    JSONPath: '.status.annotations.configuredMemberCount'
+  - name: Status
+    description: Whether all member namespaces have been configured or why
+      that's not the case
+    type: string
+    JSONPath: .status.conditions[?(@.type=="Ready")].reason
+  - name: Age
```
rcernich, line 29: How is this useful?  Would last modified be better, if you had this field at all?

manifests-maistra/1.1.0/servicemeshmemberrolls.crd.yaml
```diff
@@ -16,7 +16,22 @@ spec:
     status: {}
   version: v1
   additionalPrinterColumns:
-  - JSONPath: .spec.members
+  - name: Ready
+    description: How many of the total number of member namespaces are
+      configured
+    type: string
+    JSONPath: '.status.annotations.configuredMemberCount'
```
rcernich, line 23: Would configured members be better?  I think this would be useful for debugging things like why injection hasn't occurred (e.g. project's not configured).

manifests-maistra/1.1.0/servicemeshcontrolplanes.crd.yaml
```diff
@@ -16,25 +16,30 @@ spec:
...
-  - JSONPath: .status.conditions[?(@.type=="Reconciled")].message
-    name: Reconciliation Status
-    description: The status of the reconciliation process, if the control plane is
-      not up to date with the latest version this resource.
+    JSONPath: .status.lastAppliedConfiguration.version
+  - name: Age
+    description: The age of the object
+    type: date
+    JSONPath: .metadata.creationTimestamp
+  - name: Image HUB
```
rcernich, line 41: If you're going to have hub, why not tag.  I think tag is probably more useful, as it implies an install version.

I'm leery about adding hub, as we might move to using image streams in the future.  Maybe that would imply a "local" hub.

Regardless, would `Image Registry` be a better term?

manifests-maistra/1.1.0/servicemeshcontrolplanes.crd.yaml
```diff
@@ -16,25 +16,30 @@ spec:
     status: {}
   version: v1
   additionalPrinterColumns:
-  - JSONPath: .status.conditions[?(@.type=="Ready")].status
-    name: Ready
+  - name: Ready
+    description: How many of the total number of member namespaces are
```
rcernich, line 20: Looks like this might be mislabeled.  Also, it seems like it might be better to list the components that are not ready vs a count.  That would allow users to easily see what's not up and running.

rcernich (commented):
Should we also add a column for the actual status (e.g. installation paused: waiting for galley to become available).  I think it's important to have this info, because even though we create events for these, the events are not persistent like the status field.

manifests-maistra/1.1.0/servicemeshmemberrolls.crd.yaml
jwendell: MAISTRA-719

manifests-maistra/1.1.0/servicemeshmemberrolls.crd.yaml
luksa: I added this mostly for consistency with all the other k8s resources, as they all have this column.

manifests-maistra/1.1.0/servicemeshcontrolplanes.crd.yaml
luksa: I initially called this column "IMAGE REGISTRY", as it _is_ a better term. But since we use "hub" in the smcp.spec, I thought we should keep it consistent with the spec.

I'll also add the TAG.

manifests-maistra/1.1.0/servicemeshmemberrolls.crd.yaml
luksa: No. Lists that could potentially be very long don't belong in the default output. I've moved the list of configured members to the `-o wide` output option. But even that's an anti-pattern. The list should be capped at around five elements. If users want to see the full list, they should use `oc describe`.

manifests-maistra/1.1.0/servicemeshcontrolplanes.crd.yaml
luksa: This follows the pattern used elsewhere - in deployments, where the READY column shows the number of pods, and in pods, where it shows the number of containers that are ready vs. the total number.

We should follow the same pattern in our own resources.

Showing just the number of "unready" components is kinda weird. And it doesn't show how many components there are in total. The ready/total approach gives you more information and doesn't use up any more width, since the column header is wider than the actual value. Also, naming the column "UNREADY" would just make the column wider.

luksa: Remember, `oc get` should show just the summary info that allows users to quickly see whether everything is OK or not. When something is wrong, users typically use `oc describe` to get further info.

manifests-maistra/1.1.0/servicemeshmemberrolls.crd.yaml
rcernich: Will those be available in `oc describe`?

rcernich (approved):

manifests-maistra/1.1.0/servicemeshmemberrolls.crd.yaml
luksa: Yes. `oc describe` shows the whole `.spec` and `.status` converted into human-readable form for any resource.

jwendell (approved):

PR #348, opened 2020-02-20
MAISTRA-1192 & MAISTRA-1193 Add mutating webhooks for SMMR and SMCP

luksa:


pkg/controller/servicemesh/webhooks/server.go
```diff
@@ -0,0 +1,116 @@
...
+	operatorNamespace := common.GetOperatorNamespace()
+	hookServer, err := webhook.NewServer(
+		componentName,
+		mgr,
+		webhook.ServerOptions{
+			Port:    11999,
+			CertDir: "/tmp/cert",
+			BootstrapOptions: &webhook.BootstrapOptions{
+				ValidatingWebhookConfigName: fmt.Sprintf("%s.servicemesh-resources.maistra.io", operatorNamespace),
+				MutatingWebhookConfigName:   fmt.Sprintf("%s.servicemesh-resources.maistra.io", operatorNamespace),
```
dgn, line 39: I guess this was here before but I'm not sure I understand- what is this ConfigName about?

dgn (approved):
Looks good.

I'm assuming removing the code that e.g. removes the control plane namespaces from memberrolls in the controller will come with a later PR?

rcernich (approved):

pkg/controller/servicemesh/webhooks/server.go
luksa: That's the name of the ValidatingWebhookConfiguration object that the operator creates.

luksa:
> Looks good.
> 
> I'm assuming removing the code that e.g. removes the control plane namespaces from memberrolls in the controller will come with a later PR?

Yes. We can do that later when we also remove the if statements that skip the control plane namespace.

PR #351, opened 2020-02-25
MAISTRA-1213 Deploy prometheus using global hub and tag (fixed sed pa…

dgn:
…ttern)


luksa (approved):

luksa (changes requested):
I take back my approval. There are two containers in the deployment.yaml file. Don't we need to change the image in both?

dgn:
> I take back my approval. There are two containers in the deployment.yaml file. Don't we need to change the image in both?

Only after running the script there are two. The first one is injected using the sed command above the line I am touching.

luksa (approved):
Oh, sorry, my mistake. I tried this by manually modifying the yaml, but I modified the wrong container. Oops.

rcernich (approved):

PR #352, opened 2020-02-25
[Maistra-1215][Maistra-1211] Added collect-olm-manifests

brian-avery:
In https://github.com/Maistra/rpms/commit/bfe95a98d5f923d623b3d3b8b72e65336c4b5e2e, the operator RPM build was updated to use a `BUILD_TYPE` instead of community. These changes were never propagated here. This caused every build to be a community build.

This is part one of three PRs. This PR adds a collect-olm-manifests endpoint which will be used to remove BUILD_TYPE from the RPM build. The other reference to BUILD_TYPE in the RPM build will be reverted to `COMMUNITY=true` or `COMMUNITY=false` to select the build type.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,11 +21,12 @@ IMAGE           ?= docker.io/maistra/istio-ubi8-operator:${MAISTRA_VERSION}
 CONTAINER_CLI   ?= docker
 COMMUNITY       ?= true
 
-SOURCE_DIR        := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
-RESOURCES_DIR     = ${SOURCE_DIR}/resources
-OUT_DIR           = ${SOURCE_DIR}/tmp/_output
-TEMPLATES_OUT_DIR = ${OUT_DIR}/resources/default-templates
-HELM_OUT_DIR      = ${OUT_DIR}/resources/helm
+SOURCE_DIR          := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+RESOURCES_DIR        = ${SOURCE_DIR}/resources
+OUT_DIR              = ${SOURCE_DIR}/tmp/_output
+TEMPLATES_OUT_DIR    = ${OUT_DIR}/resources/default-templates
+HELM_OUT_DIR         = ${OUT_DIR}/resources/helm
+OLM_MANIFEST_OUT_DIR = ${OUT_DIR}/manifests
 
 OFFLINE_BUILD       ?= false
 GIT_UPSTREAM_REMOTE ?= $(shell git remote -v |grep --color=never ':Maistra/istio-operator\.git.*(fetch)' |grep --color=never -o '^[^[:space:]]*')
@@ -41,6 +42,8 @@ BUILD_TYPE = maistra
 else
 BUILD_TYPE = servicemesh
 endif
+
+$(info   Building $(BUILD_TYPE) operator)
 
 export SOURCE_DIR OUT_DIR MAISTRA_BRANCH MAISTRA_VERSION VERSION COMMUNITY BUILD_TYPE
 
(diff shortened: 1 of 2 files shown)


jwendell (commented):
if you run git grep COMMUNITY you'll see other references, better to change all of them for consistency?

build/generate-manifests.sh
```diff
@@ -1,13 +1,12 @@
 #!/usr/bin/env bash
 # to generate Maistra OLM metadata: MAISTRA_VERSION=1.0.1 REPLACES_CSV=1.0.0 tmp/build/generate-manifests.sh
-# to generate ServiceMesh OLM metadata: COMMUNITY=false MAISTRA_VERSION=1.0.1 REPLACES_CSV=1.0.0 tmp/build/generate-manifests.sh
+# to generate ServiceMesh OLM metadata: BUILD_TYPE=maistra MAISTRA_VERSION=1.0.1 REPLACES_CSV=1.0.0 tmp/build/generate-manifests.sh
```
jwendell: servicemesh

build/generate-manifests.sh
```diff
@@ -1,13 +1,12 @@
...
 
 set -e
 
-: ${COMMUNITY:-"true"}
+: ${BUILD_TYPE:-"maistra"}
 : ${MAISTRA_VERSION:?"Need to set maistra version, e.g. 1.0.1"}
-if [[ ${COMMUNITY} == "true" ]]; then
-  BUILD_TYPE="maistra"
+if [[ ${BUILD_TYPE} == "maistra" ]]; then
```
jwendell: remove the `BUILD_TYPE="servicemesh"` line a few lines below

jwendell (approved):
the description should change to reflect it as well... at this point perhaps a new pr is better. do you need a jira for this?

brian-avery:
Thanks. Updated the description and added references to the 2 Maistra issues this is addressing.

Makefile
```diff
@@ -152,6 +155,12 @@ collect-charts: collect-1.0-charts collect-1.1-charts
...
 collect-templates: collect-1.0-templates collect-1.1-templates
 
+.PHONY: collect-olm-manifests
+collect-olm-manifests:
+	rm -rf  ${OLM_MANIFEST_OUT_DIR}
+	mkdir -p ${OLM_MANIFEST_OUT_DIR}
+	cp -ra ${SOURCE_DIR}/manifests-${BUILD_TYPE}/* ${OLM_MANIFEST_OUT_DIR}
+
 .PHONY: collect-resources
 collect-resources: collect-templates collect-charts
```
rcernich: We should add `collect-olm-manifests` here too.

rcernich (commented):
You should update the `Dockerfile` too.

Makefile
```diff
@@ -152,6 +155,12 @@ collect-charts: collect-1.0-charts collect-1.1-charts
 .PHONY: collect-templates
 collect-templates: collect-1.0-templates collect-1.1-templates
 
+.PHONY: collect-olm-manifests
+collect-olm-manifests:
+	rm -rf  ${OLM_MANIFEST_OUT_DIR}
```
rcernich, line 160: This should be part of `make clean`

brian-avery: Make clean already runs `rm -rf "${OUT_DIR}"` so I think it's fine just having it here. If you don't have it here and you run both maistra and servicemesh builds then you end up with manifest directories containing data for both build types so I think it makes sense to leave this here.

brian-avery:
@rcernich what changes do you want in the dockerfile? It's working as is.

rcernich:
This is the bit in the dockerfile: https://github.com/Maistra/istio-operator/blob/maistra-1.1/build/Dockerfile#L7

Since the build now collects the manifests into tmp/_output, we can just copy right out of there (i.e. no need for a build_type build env in the docker file).

rcernich:
fwiw, i use `make clean image` for creating test images for dev.

rcernich (approved):

PR #354, opened 2020-02-26
MAISTRA-1210 Fix inconsistencies in CNI file names (v1.1 -> v1-1)

luksa:


dgn (approved):

jwendell (approved):
Should the 1.0 be named `v1-0...` as well?

luksa:
I didn't want to change the 1.0 file names, because Rob didn't change them either. Probably to keep things backwards-compatible. Although I don't see a real reason to keep the old file names, since they are an implementation detail. And the files are cleaned up by the install-cni script on shutdown.

PR #358, opened 2020-03-03
MAISTRA-1230 Create mesh-user role and binding for SMCP

luksa:


pkg/controller/servicemesh/controlplane/rbac.go
```diff
@@ -0,0 +1,137 @@
...
+	maistrav1 "github.com/maistra/istio-operator/pkg/apis/maistra/v1"
+	"github.com/maistra/istio-operator/pkg/controller/common"
+)
+
+const (
+	roleMeshUser         = "mesh-user"
+	roleBindingMeshUsers = "mesh-users"
+)
+
+func (r *controlPlaneInstanceReconciler) reconcileRBAC(ctx context.Context) error {
```
rcernich, line 21: Where does this get called from?

Also, if this is just adding a `Role` and `RoleBinding`, why not just add them to the charts, so they're managed like any other resource?

luksa: It's called here: https://github.com/Maistra/istio-operator/blob/d941e03fcc0d5d6930057020069c6adff19ebe34/pkg/controller/servicemesh/controlplane/reconciler.go#L219

> Also, if this is just adding a Role and RoleBinding, why not just add them to the charts, so they're managed like any other resource?

That's exactly how I first implemented this, until I realized that the operator will remove all subjects from the RoleBinding every time it reconciles it (and these are the subjects that the mesh admin has to add manually).

rcernich: Thanks.  That makes sense.  Sorry I missed the caller.

rcernich (approved):

dgn (approved):

PR #360, opened 2020-03-04
MAISTRA-1237 Add SMCP UX changes that have disappeared

luksa:


manifests-maistra/1.1.0/servicemeshcontrolplanes.crd.yaml
```diff
@@ -21,7 +21,7 @@ spec:
       configured
     type: string
     JSONPath: '.status.annotations.readyComponentCount'
-  - JSONPath: .status.conditions[?(@.type=="Ready")].reason
+  - JSONPath: .status.conditions[?(@.type=="Reconciled")].reason
```
rcernich: I think there's a difference between `Reconciled` and `Ready`.  The control plane can become unready if something happens to one of its components (e.g. a pod crashes), but it is still in a `Reconciled` state (i.e. all the resources are still present and up to date).

luksa: I agree, but the READY column already shows a detailed readiness status (ready vs total number of components), so it makes sense for the STATUS column to show something else.

Ideally, it would show the reconciled condition during reconciliation, and the readiness condition when the reconciled condition is true. But that can't be expressed in JSONPath, so we'd have to add another status.annotation. 

For now, I think just showing the reconciled condition is good enough, since it doesn't require any additional coding.

rcernich: Having a simple `Ready` state made it pretty easy for users to know if the control plane was good to go.  Are you saying that now they have to look at ready count and compare that to total count? Can you provide an example of what the new `kubectl get` looks like?

luksa:
```
$ kubectl get smcp
NAME              READY   STATUS              TEMPLATE   VERSION   AGE
minimal-install   8/8     InstallSuccessful   default    v1.0      56m

$ kubectl get smcp -o wide
NAME              READY   STATUS              TEMPLATE   VERSION   AGE   IMAGE HUB           TAG     READINESS STATUS MESSAGE                  RECONCILIATION STATUS MESSAGE
minimal-install   8/8     InstallSuccessful   default    v1.0      56m   docker.io/maistra   1.0.4   All component deployments are Available   Successfully installed version unknown-1
```

This is analogous to `kubectl get deployments` and `get pods`, where you also need to look at the READY column to see if all the pods / containers are ready. Also, for pods: when a container becomes unready, the STATUS column still says "Running". 

```
$ kubectl get po
NAME                              READY   STATUS    RESTARTS   AGE
istio-citadel-d69684965-qkp7r     1/1     Running   0          60m
```

rcernich: Works for me.  Thanks!

rcernich (approved):

dgn:
The reason they disappeared is because we generate these files from what's in `deploy/{servicemesh|maistra}-operator.yaml`. See the `generate-product-manifests` target - we'll have to update those files to make sure the changes don't disappear again

luksa:
(*facepalm*) Thank you, Daniel. 

Why do we even have these generated files in VCS? Shouldn't they be generated as part of the build?

dgn:
> (_facepalm_) Thank you, Daniel.
> 
> Why do we even have these generated files in VCS? Shouldn't they be generated as part of the build?

Yeah, they probably should.. I was going to add the checks to `make gen-check`, but it's hard to do because the CSV files contain a date that is always updated when you generate them

luksa:
> Yeah, they probably should.. I was going to add the checks to `make gen-check`, but it's hard to do because the CSV files contain a date that is always updated when you generate them

Yeah, I noticed the timestamp. I don't think we really need it. Who cares when the files were generated - only the version is important.

dgn (approved):

PR #361, opened 2020-03-04
MAISTRA-1076 add upgrade/downgrade validation

rcernich:


pkg/controller/servicemesh/webhooks/validation/controlplane_downgrade_v1_1.go
```diff
@@ -0,0 +1,100 @@
...
+			allErrors = append(allErrors, fmt.Errorf("ca.istio.io/override label on namespace %s is not supported in older version", member.GetName()))
+		}
+	}
+
+	// Any VirtualService http entries use mirrorPercent attribute
+	vsgvk := schema.GroupVersionKind{Group: "networking.istio.io", Version: "v1alpha3", Kind: "VirtualService"}
+	virtualServices := &unstructured.UnstructuredList{}
+	virtualServices.SetGroupVersionKind(vsgvk)
+	// XXX: do we list all in the cluster, or list for each member namespace?
+	if err := v.client.List(context.TODO(), nil, virtualServices); err != nil {
```
rcernich: There are a couple of places where we do things like this.  Should we?

My concerns are with the number of calls to the apiserver and possibly the amount of data returned.  We query a number of deprecated resource types to ensure an upgrade won't break (we could remove that logic), and we query a number of resources to see if they're configured in an unsupported manner (e.g. http- named ports using 443).

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -60,10 +59,23 @@ func (v *ControlPlaneMutator) Handle(ctx context.Context, req atypes.Request) at
...
+			if len(smcp.Status.AppliedVersion) == 0 {
+				// this must have been created before 1.1
+				newSmcp.Spec.Version = maistra.LegacyVersion.String()
+			} else {
+				// don't change the version
+				newSmcp.Spec.Version = smcp.Status.AppliedVersion
+			}
+			log.Info("Setting .spec.version to default value", "version", maistra.LegacyVersion.String())
+			smcpMutated = true
+		}
```
luksa: Do we really need to set the version on update? This code won't be invoked until the user updates the SMCP, which means that the rest of the code always needs to treat an empty version as legacy. So if we need to do this everywhere else, what is the point of having this code here?

pkg/controller/servicemesh/webhooks/validation/controlplane.go
```diff
@@ -77,9 +83,99 @@ func (v *ControlPlaneValidator) Handle(ctx context.Context, req atypes.Request)
...
+	if req.AdmissionRequest.Operation == admissionv1beta1.Update {
+		// verify update
+		oldsmcp := &maistrav1.ServiceMeshControlPlane{}
+		err := v.decoder.DecodeRaw(req.AdmissionRequest.OldObject, oldsmcp)
+		if err != nil {
+			logger.Error(err, "error decoding admission request")
+			return admission.ErrorResponse(http.StatusBadRequest, err)
+		} else if oldsmcp.ObjectMeta.DeletionTimestamp != nil {
+			logger.Info("skipping deleted smcp resource")
+			return admission.ValidationResponse(true, "")
```
luksa: These three lines can be removed, since we already check if the (new) Object is deleted. If the new Object is not deleted, then the old Object also can't be.

pkg/controller/servicemesh/webhooks/validation/controlplane.go
```diff
@@ -55,10 +57,14 @@ func (v *ControlPlaneValidator) Handle(ctx context.Context, req atypes.Request)
...
 
 	if len(smcp.Spec.Version) > 0 {
-		if _, ok := common.GetCNINetworkName(smcp.Spec.Version); !ok {
-			return validationFailedResponse(http.StatusBadRequest, metav1.StatusReasonBadRequest, fmt.Sprintf("invalid Version specified; supported versions are: %v", common.GetSupportedVersions()))
+		if version, err := maistra.ParseVersion(smcp.Spec.Version); err != nil {
+			return validationFailedResponse(http.StatusBadRequest, metav1.StatusReasonBadRequest, fmt.Sprintf("invalid Version specified; supported versions are: %v", maistra.GetSupportedVersions()))
+		} else if err := v.validateVersion(smcp, version); err != nil {
+			return validationFailedResponse(http.StatusBadRequest, metav1.StatusReasonBadRequest, err.Error())
 		}
+
```
luksa: nit: unnecessary blank line

pkg/controller/servicemesh/webhooks/validation/controlplane.go
```diff
@@ -77,9 +83,99 @@ func (v *ControlPlaneValidator) Handle(ctx context.Context, req atypes.Request)
...
+	case maistra.V1_1:
+		return v.validateDowngradeFromV1_1(smcp)
+	default:
+		return fmt.Errorf("upgrade from version %s is not supported", currentVersion.String())
+	}
+}
+
+func (v *ControlPlaneValidator) getSMMR(smcp *maistrav1.ServiceMeshControlPlane) (*maistrav1.ServiceMeshMemberRoll, error) {
+	smmr := &maistrav1.ServiceMeshMemberRoll{}
+	err := v.client.Get(context.TODO(), client.ObjectKey{Namespace: smcp.GetNamespace(), Name: common.MemberRollName}, smmr)
```
luksa, line 184: Context needs to be passed in from calling function.

pkg/controller/servicemesh/webhooks/validation/controlplane_test.go
```diff
@@ -84,9 +91,517 @@ func TestUpdateOfValidControlPlane(t *testing.T) {
...
+					name:      "valid",
+					smcp:      newControlPlaneWithVersion("my-smcp", "istio-system", "v1.0"),
+					configure: func(smcp *maistra.ServiceMeshControlPlane) {},
+					allowed:   true,
+				},
+				{
+					name: "global.proxy.alwaysInjectSelector=false",
+					smcp: newControlPlaneWithVersion("my-smcp", "istio-system", "v1.0"),
+					configure: func(smcp *maistra.ServiceMeshControlPlane) {
+						unstructured.SetNestedField(smcp.Spec.Istio, false, strings.Split("global.proxy.alwaysInjectSelector", ".")...)
```
luksa: Oooh, I like this approach. How about reducing the boilerplate even further by introducing a function like this:
```go
func setNestedField(obj map[string]interface{}, path string, value interface{}) {
    unstructured.SetNestedField(obj, value, strings.Split(path, ".")...)
}
```

This would make the tests even easier to read:
```go
  setNestedField(smcp.Spec.Istio, "global.proxy.alwaysInjectSelector", false)
```

pkg/controller/common/helm.go
```diff
@@ -56,7 +52,7 @@ var Options = &options{}
 // GetChartsDir returns the location of the Helm charts. Similar layout to istio.io/istio/install/kubernetes/helm.
 func (o *options) GetChartsDir(maistraVersion string) string {
 	if len(maistraVersion) == 0 {
-		maistraVersion = DefaultMaistraVersion
+		maistraVersion = maistra.DefaultVersion.String()
 	}
 	// FIXME: Should not be hardcoded when https://issues.jboss.org/browse/MAISTRA-766 is implemented
```
dgn: nit: I know this is old, but I think this is no longer hard-coded - so the FIXME can be removed :)

pkg/controller/servicemesh/webhooks/validation/controlplane_downgrade_v1_1.go
dgn: Should we maybe use `istio/client-go` to query the objects instead of unstructured?

pkg/controller/servicemesh/webhooks/validation/controlplane_upgrade_v1_0.go
```diff
@@ -0,0 +1,117 @@
...
+			}
+			// we only care about resources in this mesh, which aren't being managed by the operator directly
+			if meshNamespaces.Has(metaObj.GetNamespace()) && !metav1.IsControlledBy(metaObj, smcp) {
+				allErrors = append(allErrors, fmt.Errorf("%s/%s of type %s is not supported in newer version", metaObj.GetNamespace(), metaObj.GetName(), gvk.String()))
+			}
+			return nil
+		})
+	}
+
+	// Any service ports using 443 are using http/http2 in their name (http not allowed on port 443)
```
dgn, line 98: Is this a meaningful check? This is more of a general configuration issue than something that comes up during upgrade, right?

pkg/controller/servicemesh/webhooks/validation/controlplane_v1_0.go
```diff
@@ -0,0 +1,33 @@
+package validation
+
+import (
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	maistrav1 "github.com/maistra/istio-operator/pkg/apis/maistra/v1"
+)
+
+func (v *ControlPlaneValidator) validateV1_0(smcp *maistrav1.ServiceMeshControlPlane) error {
```
dgn: I think this is a good start for better validation!

pkg/controller/servicemesh/webhooks/validation/controlplane_upgrade_v1_0.go
```diff
@@ -0,0 +1,117 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	maistrav1 "github.com/maistra/istio-operator/pkg/apis/maistra/v1"
+)
+
+var (
+
+	// These are unsupported in v1.1
+	unsupportedOldResourcesV1_1 = []schema.GroupVersionKind{
```
luksa: I'm not sure we should be rejecting the upgrade if these resources are present. By doing so, we're forcing users to first delete these resources and then update the SMCP.  But users usually update multiple resources in a single kubectl apply command. Plus, doesn't the operator itself create some of these resources when you create the SMCP? Does this mean they have to delete them manually even though they never created them?

pkg/controller/servicemesh/webhooks/validation/controlplane_downgrade_v1_1.go
luksa: We should use caching. It's not appropriate for a webhook to produce multiple API requests. Also consider one of my other comments: should we really check for the existence of other resource types?

pkg/controller/servicemesh/webhooks/validation/controlplane_upgrade_v1_0.go
rcernich: I guess it depends on your general philosophy of what we should be checking.  If the user is using http/http2 protocols on port 443, an upgrade will break their application.  The logic in these upgrade/downgrade checks is intended to ensure an upgrade or downgrade won't break their applications.  I think it would be beneficial for the operator to migrate things like this when possible (nothing in 1.0 <-> 1.1, but I could see migrating Policy to AuthorizationPolicy in the future).

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
rcernich: What if the user removes the `Version` field during an update?  I think this will need to be reset to the previously `AppliedVersion` or else a `1.0` control plane will be installed, whether they want that or not.  Regarding the legacy initialization, I think it does help, as currently, reconciliation only occurs when the SMCP resource has been modified, so this would simplify reconciliation.  However, if we add code to trigger reconciliation in response to user modifications of managed resources, we would need to ensure empty values for `Version` are supported in the reconciler.

pkg/controller/servicemesh/webhooks/validation/controlplane_downgrade_v1_1.go
rcernich: We should, but I was weary of adding that as a dependency.  I'll see how much effort is involved.

pkg/controller/servicemesh/webhooks/validation/controlplane.go
rcernich: I'll remove this.

pkg/controller/servicemesh/webhooks/validation/controlplane.go
rcernich: I'll update this.

pkg/controller/servicemesh/webhooks/validation/controlplane_upgrade_v1_0.go
rcernich: Good question.  First, controller managed resources are filtered from the check, so those will not prevent an upgrade.  Regarding user created CRs, I did not check to see if there was an upgrade path.  These were deprecated in Istio 1.2 and removed in Istio 1.3, and since we're moving from Istio 1.1 to 1.4, it's likely there isn't anything in Istio 1.1 that supports migration to a new resource.  Is it possible to return a warning from validation?

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
luksa: Oh, I wouldn't allow unsetting the version, once it's set. I mean, the mutating webhook should default the version to the one in the oldObject (if set).

Also, I do think we'll start triggering reconciliation much more frequently than we do now. It's really not ok to fix the owned resources only when the user updates the SMCP. That goes against the basic idea of declarative configuration.

pkg/controller/servicemesh/webhooks/validation/controlplane_upgrade_v1_0.go
rcernich: I just realized that we do check for `AuthorizationPolicy` on downgrade.  Should we add logic to migrate that to Policy/RbacConfig on downgrade?

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
rcernich: Okay, I'll modify this to initialize it to the `oldSmcp.Spec.Version` and leave it at that.  This will mean that `Version == ""` will imply `v1.0`.

luksa: We could add a `GetEffectiveVersion()` to the SMCP struct to centralize this, but I don't know how we'd force ourselves to use the function instead of simply reading the field.

pkg/controller/servicemesh/webhooks/validation/controlplane_downgrade_v1_1.go
rcernich: These types have been removed from istio/client-go@1.4.5, so we're stuck using unstructured.

rcernich:
Ready for another review.

I was unable to integrate istio/client-go, as the deprecated resources are not available in 1.4.5.
Version is now defaulted to the old version if it's blank in the updated resource.

luksa:
> Version is now defaulted to the old version if it's blank in the updated resource.

Do we still need the `AppliedVersion` field now?

rcernich:
re `AppliedVersion`...good question.  I think so, as this is the only way the reconciler knows whether or not it is performing an upgrade of the control plane.  That said, we don't actual have any special upgrade/downgrade logic, so it wouldn't be used.

pkg/controller/servicemesh/webhooks/validation/controlplane_downgrade_v1_1.go
dgn: I see, thanks for checking!

rcernich:
@luksa, @dgn, @knrc, did we want to merge this for 1.1?

dgn:
> @luksa, @dgn, @knrc, did we want to merge this for 1.1?

I'll test this for a bit today

pkg/controller/servicemesh/webhooks/validation/controlplane_downgrade_v1_1.go
```diff
@@ -0,0 +1,100 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	maistrav1 "github.com/maistra/istio-operator/pkg/apis/maistra/v1"
+	"github.com/maistra/istio-operator/pkg/controller/common"
+)
+
+var (
+	// These are unsupported in v1.1
+	unsupportedNewResourcesV1_0 = []schema.GroupVersionKind{
+		schema.GroupVersionKind{Group: "security.istio.io", Version: "v1beta1", Kind: "AuthorizationPolicy"},
```
dgn: The operator doesn't have permissions to list these

pkg/controller/servicemesh/webhooks/validation/controlplane_downgrade_v1_1.go
```diff
@@ -0,0 +1,100 @@
...
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	maistrav1 "github.com/maistra/istio-operator/pkg/apis/maistra/v1"
+	"github.com/maistra/istio-operator/pkg/controller/common"
+)
+
+var (
+	// These are unsupported in v1.1
```
dgn: should be `in v1.0`

pkg/controller/servicemesh/webhooks/validation/controlplane_upgrade_v1_0.go
dgn: > I guess it depends on your general philosophy of what we should be checking. If the user is using http/http2 protocols on port 443, an upgrade will break their application. 

Didn't know this was the case! Makes total sense to check this then, sorry

dgn (commented):
Found one permission issue.

dgn (approved):
I think this is good to go as soon as the permission issue is resolved, I also consider it low-risk enough to be merged at this time. Will let final vote go to @knrc and @luksa. Thanks Rob!

pkg/controller/servicemesh/webhooks/validation/controlplane_upgrade_v1_0.go
```diff
@@ -0,0 +1,117 @@
...
+	// Any service ports using 443 are using http/http2 in their name (http not allowed on port 443)
+	for namespace := range meshNamespaces {
+		memberServices := &corev1.ServiceList{}
+		// listing for each member namespace, as we expect a large number of services in the whole cluster
+		if err := v.client.List(ctx, client.InNamespace(namespace), memberServices); err != nil {
+			return pkgerrors.Wrapf(err, "error listing Service resources in namespace %s", namespace)
+		}
+		for _, service := range memberServices.Items {
+			for _, port := range service.Spec.Ports {
+				if port.Port == 443 && (strings.HasPrefix(port.Name, "http-") || strings.HasPrefix(port.Name, "http2-")) {
```
dgn: this doesn't catch the case where the port is simply named `http` or `http2` (which is what I used for testing, coincidentally)

rcernich: Thanks @dgn.  I'll update this and add a test case for it.

pkg/controller/servicemesh/webhooks/validation/controlplane_downgrade_v1_1.go
rcernich: will fix

pkg/controller/servicemesh/webhooks/validation/controlplane_downgrade_v1_1.go
rcernich: will update the roles

rcernich: Side note, should we look at deprecating ServiceMeshRbacConfig (our version of ClusterRbacConfig, which was replaced by AuthorizationPolicy) in favor of using AuthorizationPolicy?

rcernich:
I've processed @dgn's last round of comments and am currently looking at manually adding typed resources for the istio types used in the upgrade/downgrade validation, to get cache support from the controller-runtime client.

pkg/controller/servicemesh/webhooks/validation/controlplane_downgrade_v1_1.go
knrc: +1, rbac should be deprecated in 1.4

rcernich:
This last round of changes integrates @dgn's work for MAISTRA-1160 into 1.0 validation.  Validation will only be performed on 1.0 SMCP resources when dowgrading from 1.1.  I think this should be safe, because an upgrade would be prevented if those settings are misconfigured.  The only setting that would prevent downgrading is the use of `kiali.jaegerInClusterURL` and some other 1.1 specific settings.  Users can remove the offending settings when downgrading.

(This topic came up because of MAISTRA-1160.  In 1.0 it might have been possible for users to use various configurations with zipkin.address set, which are flagged as invalid in 1.1.  Validation for these cases is only checked in 1.0 when downgrading from 1.1, and since they can't upgrade with a misconfigured system, downgrading shouldn't be a problem either.)

rcernich:
This also includes some minor tweaks to the default logic, as we were treating `smcp.Spec.Version == ""` as `v1.1`, when it should be treated as legacy `v1.0`.  In other words, whenever `Version` is not set, it should be assumed to be `v1.0`, always.  This was not consistent before.  (Note, the mutating webhook will initialize `Version` to the latest version, which is `v1.1` today, i.e. defaulting occurs in the mutating webhook.)

rcernich:
I've verified that the "typed" resources used in the validation check work in a real install, so this should be good to go. @knrc, @luksa, it requires your final blessing before it can be merged.  Please review and approve if you are satisfied.

knrc (approved):

PR #366, opened 2020-03-09
MAISTRA-1214 Fail patch scripts if a sed command does not match

dgn:


build/download-charts.sh
```diff
@@ -61,6 +61,16 @@ function retrieveIstioRelease() {
   )
 }
 
+function sed() {
+  state=$(git diff)
+  command sed "$@"
+  difference=$(diff <(echo "${state}") <(git diff) || true )
+  if [[ -z "${difference}" ]]; then
+    echo "ERROR: nothing changed, sed seems to not have matched. Exiting"
```
luksa, line 78: Would be nice if this showed the actual sed command that failed (maybe just print out $@).

dgn: I included the failing call and found some other issues with the git compare- I'm just straight up diffing the file itself now, PTAL

luksa (approved):

build/download-charts.sh
```diff
@@ -61,6 +61,19 @@ function retrieveIstioRelease() {
   )
 }
 
+function sed() {
```
jwendell, line 64: I'd rather call this something else, e.g, `sed_wrapper`. This is to avoid confusion with actual sed command.

jwendell (commented):
Although this is better than our current state, it's still not 100% accurate, as we have sed invocations that do more than one substitution, and this patch does not check if each of them actually succeeded.

The solution would be to replace `sed -e ... -e ...` with many individuals `sed_wrapper` calls.

build/download-charts.sh
dgn: is that a problem? the idea was that you don't have to know this to have it apply

dgn:
> Although this is better than our current state, it's still not 100% accurate, as we have sed invocations that do more than one substitution, and this patch does not check if each of them actually succeeded.
> 
> The solution would be to replace `sed -e ... -e ...` with many individuals `sed_wrapper` calls.

That's a good point. I guess we could split our sed calls to make sure we're matching all sub-patterns

build/download-charts.sh
jwendell: I'd prefer not to make things hard to debug. Overwriting system "functions" like this can be dangerous when something break but when we copy and paste the sed command in the terminal it works. Then, after having wasted a lot of time one finds out that there's a sed overwrite function.

dgn: Done. This is now just a failing function telling you to use `sed_wrap`

dgn:
I split up all the `sed` invocations and renamed the function to `sed_wrap`, PTAL @jwendell

jwendell (approved):
thanks daniel!

PR #367, opened 2020-03-11
MAISTRA-1217 Pull in chart changes due to 1.4.6 update

dgn:


resources/helm/v1.1/istio/files/injection-template.yaml
```diff
@@ -278,27 +320,22 @@ containers:
...
-    {{ else -}}
-    capabilities:
+      {{- end }}
       drop:
-      - KILL
-      - SETUID
-      - SETGID
-      - MKNOD
-    {{ if .Values.global.sds.enabled }}
+      - ALL
```
dgn, line 333: @luksa I am not 100% sure the SCC controller will be smart enough to not overwrite this - `ALL` should of course cover necessary restrictions, but it might be looking for specific caps?

jwendell (approved):

rcernich (approved):

dgn:
While testing I found a bunch of issues with the updated charts. I fixed them in the follow-up commits. PTAL

resources/helm/v1.1/istio/files/injection-template.yaml
dgn: I added them back

rcernich (approved):

PR #372, opened 2020-03-12
MAISTRA-1241 Remove namespace/finalizers update permissions from galley

dgn:


build/download-charts.sh
```diff
@@ -71,10 +71,10 @@ function sed() {
 function sed_wrap() {
   for filename; do true; done # this retrieves the last argument
   echo "patching $filename"
-  state=$(cat $filename)
+  oldchecksum=$(cat $filename | md5sum)
```
jwendell: nit: you seem to mix `$variable` with `${variable}` usages. The latter is preferred.

also, I'm curious: in what cases a md5sum can catch changes that a diff wouldn't?

also², do you think this is better done in a separate pr?

dgn: You're right, I removed the commit from this PR.

dgn: I changed because while working on the sed pattern I noticed that the diff piping sometimes didn't capture when nothing had changed, I couldn't explain it myself. I then switched over to checksums, which should be more performant anyway, and the problem went away

jwendell (approved):

rcernich (approved):

PR #375, opened 2020-03-16
MAISTRA-1110 cleanup initialization of Status; remove unread MeshVersion

rcernich:

### DIFF
--- a/pkg/controller/common/test/manager.go
+++ b/pkg/controller/common/test/manager.go
@@ -26,7 +26,7 @@ import (
 type FakeManager struct {
 	manager.Manager
 	recorderProvider recorder.Provider
-	reconcileWG sync.WaitGroup
+	reconcileWG      sync.WaitGroup
 }
 
 var _ manager.Manager = (*FakeManager)(nil)
@@ -37,14 +37,19 @@ func StartManager(mgr manager.Manager, t *testing.T) func() {
 	t.Helper()
 	stopChannel := make(chan struct{})
 	startChannel := make(chan struct{})
+
 	go func() {
 		t.Helper()
+		t.Logf("starting manager.Manager")
 		if err := mgr.Start(stopChannel); err != nil {
 			t.Fatalf("Uexpected error returned from manager.Manager: %v", err)
 		}
-		t.Logf("manager stopped")
+		t.Logf("manager.Manager stopped cleanly")
 		close(startChannel)
 	}()
+
+	mgr.GetCache().WaitForCacheSync(stopChannel)
+	t.Logf("manager Cache synchronized")
 	return func() {
 		close(stopChannel)
 		select {
@@ -63,7 +68,7 @@ func NewManager(scheme *runtime.Scheme, tracker clienttesting.ObjectTracker, gro
 	return &FakeManager{
 		Manager:          delegate,
 		recorderProvider: NewRecorderProvider(scheme),
-		reconcileWG: sync.WaitGroup{},
+		reconcileWG:      sync.WaitGroup{},
 	}, nil
 }
 
@@ -251,11 +256,34 @@ func NewUpdateSimulator(tracker clienttesting.ObjectTracker) clienttesting.React
 		// XXX: update fields that would get modified by the api server on an update
 		accessor.SetResourceVersion(fmt.Sprintf("%d", rand.Int()))
 		if accessor.GetDeletionTimestamp() == nil && len(updateAction.GetSubresource()) == 0 {
-			// XXX: this should only be done if the .spec field actually changes.
-			// i don't think we should do this if finalizers, annotations, labels, etc. are updated
-			accessor.SetGeneration(accessor.GetGeneration() + 1)
+			existingObj, err := tracker.Get(action.GetResource(), accessor.GetNamespace(), accessor.GetName())
+			if err == nil && specChanged(obj, existingObj) {
+				// this should only be done if the .spec field actually changes.
+				accessor.SetGeneration(accessor.GetGeneration() + 1)
+			}
 		}
 		err = tracker.Update(updateAction.GetResource(), obj, accessor.GetNamespace())
 		return true, obj, err
 	})
+}
+
+func specChanged(new, old runtime.Object) (changed bool) {
+	func() {
+		defer func() {
+			if r := recover(); r != nil {
+				fmt.Println("panic'd checking spec fields")
+				changed = true
+			}
+		}()
+		newSpec := reflect.ValueOf(new).Elem().FieldByName("Spec")
+		oldSpec := reflect.ValueOf(old).Elem().FieldByName("Spec")
+		if newSpec.IsValid() {
+			if newSpec.CanInterface() {
+				changed = !reflect.DeepEqual(newSpec.Interface(), oldSpec.Interface())
+			} else {
+				changed = !reflect.DeepEqual(newSpec.Elem().Interface(), oldSpec.Elem().Interface)
+			}
+		}
+	}()
+	return
 }
--- a/pkg/controller/common/test/manager.go
+++ b/pkg/controller/common/test/manager.go
@@ -251,6 +251,8 @@ func NewUpdateSimulator(tracker clienttesting.ObjectTracker) clienttesting.React
 		// XXX: update fields that would get modified by the api server on an update
 		accessor.SetResourceVersion(fmt.Sprintf("%d", rand.Int()))
 		if accessor.GetDeletionTimestamp() == nil && len(updateAction.GetSubresource()) == 0 {
+			// XXX: this should only be done if the .spec field actually changes.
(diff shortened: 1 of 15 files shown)


rcernich:
This is ready to go.  Cleans up some cruft (MAISTRA-1110) and makes sure the status gets updated at the end of reconciliation (if it changed, MAISTRA-1236).

rcernich:
I have no idea why the tests keep failing.  They pass for me locally.

pkg/controller/common/test/runner.go
```diff
@@ -18,78 +18,72 @@ import (
...
+				for _, assertion := range event.Assertions {
+					tracker.PrependReaction(assertion)
 				}
-			}
-			for _, addController := range testCase.AddControllers {
-				if err := addController(mgr); err != nil {
+				// insert verifier.  this needs to be the first handler, as it verifies the event, but does not handle it
+				tracker.PrependReaction(event.Verifier)
+				// add failure for events occurring after validation should be complete
+				tracker.PrependReaction(&extraneousActionFailure{verifier: event.Verifier, t: t})
```
dgn, line 71: This seems to be causing the failure

dgn (commented):
Note that the CI server is much slower than our laptops, so the order of events can be different. They pass for me locally, too - most of the time. If I run with `-race` I get timeouts on this one occasionally (apart from the fact other tests have data races).

pkg/controller/common/test/manager.go
```diff
@@ -251,6 +251,8 @@ func NewUpdateSimulator(tracker clienttesting.ObjectTracker) clienttesting.React
 		// XXX: update fields that would get modified by the api server on an update
 		accessor.SetResourceVersion(fmt.Sprintf("%d", rand.Int()))
 		if accessor.GetDeletionTimestamp() == nil && len(updateAction.GetSubresource()) == 0 {
+			// XXX: this should only be done if the .spec field actually changes.
+			// i don't think we should do this if finalizers, annotations, labels, etc. are updated
```
luksa: Correct. Why not add a simple `if !reflect.DeepEquals(oldSpec, newSpec)`?

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -558,12 +559,9 @@ func (r *controlPlaneInstanceReconciler) postReconciliationStatus(ctx context.Co
...
 func (r *controlPlaneInstanceReconciler) skipStatusUpdate() bool {
-	for _, conditionType := range []v1.ConditionType{v1.ConditionTypeInstalled, v1.ConditionTypeReconciled, v1.ConditionTypeReady} {
-		if r.Status.GetCondition(conditionType).Status != r.Instance.Status.GetCondition(conditionType).Status {
-			return false
-		}
-	}
-	return true
+	// only update status if it changed
+	return reflect.DeepEqual(r.Status.StatusType, r.Instance.Status.StatusType) &&
+		reflect.DeepEqual(r.Status.ComponentStatus, r.Instance.Status.ComponentStatus)
```
luksa: Hmm, in my WIP, I just checked the whole status. After all, any change to the status should result in the update of the status. This way, things like `status.annotations` are updated when necessary.

luksa (commented):
I've reviewed the code, but I still need to test this in action. I have the feeling that it doesn't solve the status updates completely.

pkg/controller/common/test/runner.go
rcernich: The error seems to imply another event is coming through the reconcile loop, hence the "get" request for an "smcp" resource named, "test."  I suspect this has to do with the `reflect.DeepEqual()` call in `skipStatusUpdate()`.  I wonder if this is the result of different go versions.  (I needed to modify the `LastTransitionTime` initialization, as `DeepCopy()`ing the object caused the values to be different.)

TBH, I suspect using `reflect.DeepEqual()` might be a little flaky for this use case.

pkg/controller/servicemesh/controlplane/reconciler.go
rcernich: Agreed.  I'll update it, but see comment above about using `DeepEqual`.

pkg/controller/common/test/manager.go
rcernich: It's not quite that simple, as we're only using `runtime.Object`, so we'd need to use reflection to get the `spec` values.  I can look at adding this at some point in the future.

pkg/controller/servicemesh/controlplane/reconciler.go
rcernich: `reflect.DeepEqual()` does not seem to be working.  I'm debugging to figure out which field is causing problems.

rcernich:
I need to find out why the test is failing.  The failure is indicative of extra events going through the reconcile queue, which implies we are updating status when we shouldn't be.

pkg/controller/servicemesh/controlplane/reconciler.go
rcernich: The issue here is that `metav1.Time` serializes out to the second, so a round trip through json/yaml marshaler changes the values of the `LastTransitionTime`, which causes `skipStatusUpdate()` to return false every time.  I've changed the way `LastTransitionTime` is initialized to truncate sub second details.

pkg/controller/common/test/manager.go
rcernich: I addressed this too.

pkg/controller/common/test/runner.go
rcernich: See comments below, but this is working as desired now.

rcernich:
I think the test harness is susceptible to timing issues at startup, which contribute to synchronization issues during test execution.  Sometimes the test hangs after completing successfully, sometimes it takes too long to run and times out, sometimes it errors before that.  I'll need to root through the code to see if there's any way to ensure the manager is fully started before processing events.  I'll also try to verify that there isn't any jittering going on in the event queues.

rcernich:
I can reproduce this fairly consistently using `-cpu=8` (at least 50% of the time, the test hangs).

luksa:
I never see the timeout, but I still saw the occasional WaitGroup-related issue. I've fixed it here: https://github.com/luksa/istio-operator/commit/c0ee6a6151fd1ed647de1dab1f2a1c178e88a2fc

With this fix, I now only see the following failure occasionally:
```
  runner.go:128: unexpected action ocurred: testing.GetActionImpl{ActionImpl:testing.ActionImpl{Namespace:"istio-system", Verb:"get", Resource:schema.GroupVersionResource{Group:"maistra.io", Version:"v1", Resource:"servicemeshcontrolplanes"}, Subresource:""}, Name:"test"}
```

(note: I changed the `controlPlaneNamespace` from `test` to `istio-system`)

rcernich:
@luksa, that's a legitimate error which implies an extra event is passing through the reconcile loop.  I'll pull your changes for the WaitGroup issue and see why it's failing.  The reconcile loop should be idle after the last verification and there should be no events in the handler's queue (which would trigger a reconciliation and thus the unexpected event error).

rcernich:
I think both the `WaitGroup` issue and the spurious "unexpected event" are caused by multiple `reconcile.Request` objects being added to the queue in such a way that they are not being _folded_ into a single request.  The supplied `EventHandler` implementations (e.g. `EnqueueRequestForObject`) create two requests for each update event, adding them to the queue in rapid succession.  In order for either error to occur, the controller must retrieve the first request from the queue before the second is added.  This is a very small window.

Notice the `q.Add()` calls in [EnqueueRequestForObject.Update()](https://github.com/kubernetes-sigs/controller-runtime/blob/release-0.1/pkg/handler/enqueue.go#L49-L67).

[q.Add()](https://github.com/kubernetes/client-go/blob/release-10.0/util/workqueue/queue.go#L120-L123) behavior when a similar request is being processed.  The request is marked as _dirty_, but is not added to the queue.

[q.Get()](https://github.com/kubernetes/client-go/blob/release-10.0/util/workqueue/queue.go#L156-L157) behavior.  The request is no longer _dirty_, but is being _processed_.

[q.Done()](https://github.com/kubernetes/client-go/blob/release-10.0/util/workqueue/queue.go#L171-L175) behavior when request is _dirty_.  When the reconciler has completed the request, the _dirty_ object is added to the end of the queue.

Looking at the code above, you can see that for the duplicate request to not get collapsed into the first, the thread invoking `q.Get()` must get hold of the queue's lock between those two `q.Add()` calls - not a big window.

I don't know what the solution is and I can't reproduce this very consistently.  The proper fix is probably to modify the EventHandler so it only generates a single request on updates.  This will probably work for all `kinds` of events, with the exception of owned objects or `ServiceMeshMember` objects (where the owner or referenced control plane might change).  For `EnqueueRequestForObject` it makes no sense to generate two events, as you can't modify an object's name or namespace.

luksa (approved):

dgn (approved):

PR #385, opened 2020-03-24
MAISTRA-1224 Skip flaky test

dgn:


jwendell (approved):

brian-avery (approved):

rcernich:
I think we should make the test not flaky.  This test is actually useful for what I'm doing with MAISTRA-1236, as it verifies the complete processing of a single event through the reconciler.  If the issue is with the timeout, then we should increase that.  It may be better to just use the normal test timeout instead of specifying it in the test case.  (Default test timeout is 30s.  I think I added this because I wanted to test _not_ seeing objects, which I was using to gate the end of the test, but that was before I added code to synchronize reconcile completion, so the only "hole" is whether or not another reconcile has been triggered.)

luksa:
+1 on skipping for now. I'll try to fix the test if no-one else is looking at it yet.

rcernich:
There are two issues with the test.  The primary issue we are seeing is that it times out when running in prow.  I'm working on removing the timeout now, so it will just use the normal go test timeout.  If you look at the results for the failed prow tests, this is all of the failures not caused by legitimate test errors.

The other I have seen is a race condition with the `waitForReconcileCompletion()` call, but that only happens on _really_ fast systems (basically, `Wait()` is called on the `WaitGroup` before `Add()` has been called.  I suppose this could also happen if wait is called in between events on the reconcile queue, as `Add()` and `Done()` surround `Reconciiler.Reconcile()`.  The ideal solution is to integrate synchronization into the reconciler's queue, as that would tie up all the loose ends in the test harness.  That said, I have only seen this timing issue a handful of times.

rcernich:
Found the issue with the hanging tests.  There is only one issue left, which occurs very rarely (I've only seen it a handful of times over the last few months) and may be related to poorly written test cases, or cases that should fail in other ways (i.e. reconcile queue is not empty).

luksa:
@dgn I believe you can now close this.

PR #387, opened 2020-03-26
MAISTRA-1160 Make sure custom Jaeger instances are accessible through…

dgn:
… Routes

### DIFF
--- a/resources/helm/overlays/istio/templates/networkpolicy.yaml
+++ b/resources/helm/overlays/istio/templates/networkpolicy.yaml
@@ -63,3 +63,64 @@ spec:
     - namespaceSelector:
         matchLabels:
           network.openshift.io/policy-group: ingress
+
+# if users are running a custom Jaeger instance in the control plane namespace,
+# we need to make sure that it's accessible through its Route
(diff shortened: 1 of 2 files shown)


rcernich (approved):

luksa (approved):

knrc (changes requested):
Do we not have to remove the istio-jaeger-ingress NetworkPolicy?

dgn:
> Do we not have to remove the istio-jaeger-ingress NetworkPolicy?

I didn't remove it which leads to following behaviour:
- if we deploy Jaeger ourselves, we only create a single NP with the correct assumptions about labels (that is still istio-jaeger-ingress, which is only deployed when `tracing.enabled==true`)
- if the user specifies a custom Jaeger instance, we fall back to creating two NPs to match both all-in-one and production (the NPs in this PR)

I agree that the simpler solution would be to always create two NPs, but my assumption was that we should create the minimum necessary based on what we know. Let me know if I should change that.

knrc:
I'm happy with the current approach if you are, thanks very much

knrc (approved):

PR #388, opened 2020-03-27
Controller test framework improvements

rcernich:


rcernich:
The big change here is that `testmanager.WaitForReconcileCompletion()` now looks at the controller queues and verifies that they are all empty before processing assertions.  This should fix the issue with the `panic` on the `WaitGroup` (`Add()` called after `Wait()` with zero size, and the outstanding issue to ensure the reconciler is idle before completing the test.

I also made the extraneous action assertion optional.  This will help prevent those intermittent "unexpected action occurred..." failures, and have disabled the check for the one integration test.  This will also simplify test development going forward, as the tests will only need to verify the events that are important to them, not requiring them to always verify the last expected action.  (For example, if a test is not concerned with bootstrapping, it can omit all verification and assertions that are in the current test, i.e. just post a create smcp event and wait for processing to complete.  The next test event could be updating cni status and verify the next phase of reconciliation.)

I've also scanned the documentation and updated it, where necessary.

From here on, this test framework should be good to start using for more complex, black box testing of the controllers.

dgn (approved):

luksa (approved):
Code LGTM. I have also seen zero test failures in more than 100 test runs.

PR #390, opened 2020-03-30
MAISTRA-1194 Remove applied version from SMCP

luksa:


pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -444,13 +443,13 @@ func (r *controlPlaneInstanceReconciler) renderCharts(ctx context.Context) error
...
 		// initialize version
-		if len(r.Status.AppliedVersion) == 0 {
+		if len(r.Instance.Spec.Version) == 0 {
 			// this must be from a 1.0 operator
 			r.Status.LastAppliedConfiguration.Version = maistra.LegacyVersion.String()
 		} else {
 			// use the version that's already been applied
 			// users must manually upgrade their mesh version
-			r.Status.LastAppliedConfiguration.Version = r.Status.AppliedVersion
+			r.Status.LastAppliedConfiguration.Version = r.Instance.Spec.Version
```
dgn: I think this can be simplified. This `else` block here seems like it can never be hit anyway

We're writing `r.Instance.Spec` into `LastAppliedConfiguration` above in line 443, so `LastAppliedConfiguration.Version` will always be equal to `r.Instance.Spec.Version`. I think this inner `if-else` block can be removed

pkg/controller/servicemesh/memberroll/controller_test.go
```diff
@@ -285,23 +285,19 @@ func TestReconcileReconcilesAddedMember(t *testing.T) {
 			name:                "default",
 			operatorVersion:     operatorVersionDefault,
 			meshVersion:         "",
```
dgn, line 287: this can be removed I guess? the whole field can be removed from the struct

pkg/controller/servicemesh/controlplane/reconciler.go
luksa: Good catch. Fixed.

pkg/controller/servicemesh/memberroll/controller_test.go
luksa: No, it's used to initialize `SMCP.spec.version` in line 297

pkg/controller/servicemesh/controlplane/reconciler.go
rcernich: Except for the case where `Version == ""`, in which case it gets set to `v1.0`.

rcernich (approved):
lgtm.  my only preference would be that instead of using `""` for the default cases, we use the default operator version, `v1.1` in this case.  This is a nit.

pkg/controller/servicemesh/controlplane/reconciler.go
luksa: @rcernich Yup, I kept that part.

pkg/controller/servicemesh/memberroll/controller_test.go
dgn: Sorry, my bad; only looked at one side of the diff

dgn (approved):

PR #391, opened 2020-03-31
MAISTRA-1160 Simplify external Jaeger NetworkPolicy to ignore instanc…

dgn:
…e name

### DIFF
--- a/resources/helm/overlays/istio/templates/networkpolicy.yaml
+++ b/resources/helm/overlays/istio/templates/networkpolicy.yaml
@@ -68,11 +68,11 @@ spec:
 # we need to make sure that it's accessible through its Route
 {{ if and (not .Values.tracing.enabled) (.Values.global.tracer.zipkin.address) }}
 ---
-# This allows traffic for Routes to a Jaeger all-in-one instance in the control plane namespace
+
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
-  name: istio-jaeger-ingress-all-in-one
+  name: istio-external-jaeger-ingress
   namespace: {{ .Release.Namespace }}
   labels:
     app: jaeger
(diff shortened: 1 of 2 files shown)


dgn:
@objectiser @FilipB could you take a look? I tested this successfully

dgn:
An operator image for testing is available at
```
registry.gitlab.com/dgrimm/istio/istio-operator:1160
```

rcernich (approved):

objectiser (approved):
LGTM - but assumes only one jaeger instance installed in namespace, which I think is reasonable.

dgn:
> LGTM - but assumes only one jaeger instance installed in namespace, which I think is reasonable.

Well, it will allow ingress traffic from the router for all of them, if there are >1. I think that's okay.

FilipB:
Testing with registry.gitlab.com/dgrimm/istio/istio-operator:1160 and following jaeger cr:
```
# This CR deploys Jaeger and Elasticsearch with oauth proxy
apiVersion: jaegertracing.io/v1
kind: Jaeger
metadata:
  name: external-jaeger
  namespace: istio-system
spec:
  strategy: production
  storage:
    type: elasticsearch
    elasticsearch:
      nodeCount: 1
      resources:
        requests:
          cpu: 200m
          memory: 1Gi
        limits:
          memory: 1Gi
  ingress: # Setting Up Authentication
    enabled: true
    security: oauth-proxy
    openshift:
      # This limits user access to the Jaeger instance to users who have access
      # to the control plane namespace. Make sure to set the correct namespace here
      sar: '{"namespace": "istio-system", "resource": "pods", "verb": "get"}'
      htpasswdFile: /etc/proxy/htpasswd/auth
  volumeMounts:
  - name: secret-htpasswd
    mountPath: /etc/proxy/htpasswd
  volumes:
  - name: secret-htpasswd
    secret:
      secretName: htpasswd
```

FilipB:
I'm having troubles to get registry.gitlab.com/dgrimm/istio/istio-operator:1160 into my cluster so I'm testing with latest nightly operator and creating network policy manually.
Following is not working:
```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: istio-external-jaeger-ingress
  namespace: istio-system
  labels:
    app: jaeger
  annotations:
    "maistra.io/internal": "true"
spec:
  podSelector:
    matchExpressions:
    # the value of this label depends on the deployment strategy used
    - key: app.kubernetes.io/component
      operator: In
      values:
      - all-in-one
      - query
    matchLabels:
      app.kubernetes.io/part-of: jaeger
      app.kubernetes.io/managed-by: jaeger-operator
  ingress:
  - from:
    - namespaceSelector:
        matchLabels:
          network.openshift.io/policy-group: ingress
```
and following IS working:
```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: istio-external-jaeger-ingress
  namespace: istio-system
  labels:
    app: jaeger
  annotations:
    "maistra.io/internal": "true"
spec:
  podSelector:
    matchExpressions:
    # the value of this label depends on the deployment strategy used
    - key: app.kubernetes.io/component
      operator: In
      values:
      - all-in-one
      - query
    matchLabels:
      app.kubernetes.io/part-of: jaeger
      app.kubernetes.io/managed-by: jaeger-operator
  ingress:
  - ports:
    port: 8443
```

knrc (approved):

knrc:
@FilipB Sorry, didn't see your comment until after I had approved.  @dgn it does look as if these NetworkPolicies are different from the one we use when we set up the CR, which matches @FilipB's config from above.

https://github.com/Maistra/istio-operator/blob/maistra-1.1/resources/helm/v1.1/istio/charts/tracing/templates/networkpolicy.yaml#L29-L31

FilipB:
@knrc @dgn No problem. I could use the latest nightly build of istio operator now when it contains this change and I can confirm that my comments from above still apply. I can see that istio operator creates 'istio-external-jaeger-ingress' NP but it's NOT working. Problem is in ingress part:
```
ingress:
  - from:
    - namespaceSelector:
        matchLabels:
          network.openshift.io/policy-group: ingress
```
If I change that part to:
```
  ingress:
  - ports:
    port: 8443
```
it starts working.

dgn:
It worked fine for me, strange- is this a difference in Jaeger operators? I'm testing with the latest production Jaeger from OperatorHub

FilipB:
@dgn I'm using Jaeger 1.17.1 installed via OLM from rh-verified-operators operator source, which should be the version used in SM 1.1.

dgn:
Looks like my crc cluster behaves weirdly.

PR #394, opened 2020-04-02
MAISTRA-1316 Prevent SMCP/SMMR/SMM resources from being created in the operator namespace

luksa:


rcernich (approved):

Makefile
```diff
@@ -69,7 +69,7 @@ compile:
 ################################################################################
 .PHONY: test
 test:
-	go test -v -timeout ${TEST_TIMEOUT} -mod=vendor ./...
```
dgn: let's make it optional at least. it helps with debugging failures on prow. we can make it dependant on an env var - on prow we already set `CI=prow`

dgn (approved):

Makefile
luksa: oh, I thought we added that mostly by mistake.. I'll remove it from this PR, so we can address it properly

dgn: Thanks

PR #405, opened 2020-04-14
OSSM-110 add relatedImages to csv file

rcernich:


brianredbeard:
Out of curiosity, why is viper being pulled in?  I don't see any context in the commit message.  Is this to support some later CLI mechanism for manual testing/execution?

rcernich:
I'm using viper to parse configuration for the operator.  This allows specification of the image names in one of two ways: configuration file and environment variable.  For this PR, I've elected to use a properties file to specify the related images into the deployment using a downward api volume.  This made the deployment resource a lot less verbose (we have 23 related images).

For reference, here are some examples:

```properties
olm.relatedImage.v1_1.cni=registry.redhat.io/openshift-service-mesh/istio-cni-rhel8:1.1.0
```

```bash
export OLM_RELATEDIMAGE_V1__1_CNI="registry.redhat.io/openshift-service-mesh/istio-cni-rhel8:1.1.0"
```

HTH

rcernich:
FWIW, I had initially thought about adding the ability to pass the image names in as command line arguments, which is what initially got me looking at viper.  In the end, it really came down to the fact that it supported reading in config through either an environment variable or file, and supported unmarshaling the parsed configuration into an object.

luksa:
When creating a v1.0 SMCP, the first Deployment that is created has an invalid image (e.g. docker.io/maistra/registry.redhat.io/openshift-service-mesh/citadel-rhel8:1.0.10:1.0.10). The 1.0 charts don't check if the image name already contains the hub and tag. The 1.1 charts do. 

Is there a PR that fixes this? I couldn't find it.

luksa:
I've opened https://github.com/Maistra/istio-operator/pull/411 to fix the 1.0 charts.

rcernich:
It should be part of this PR

https://github.com/Maistra/istio-operator/pull/405/commits/ad13a12fa5ab625267d2c728ba558a9f1b74e6ca#diff-b8a523fbc9e23335e648496932633b06

luksa:
Oh, weird. I ran the operator from this PR locally and it created the istio-citadel deployment with an invalid image name. Looks like I forgot to regenerate the charts in `tmp/_output/`.  Seems to work properly now.

luksa (approved):

knrc (approved):

PR #411, opened 2020-04-22
MAISTRA-1369 Fix handling of fully qualified image names

luksa:


rcernich:
I think we need to modify patch-charts.sh so these changes aren't lost in a future change.

dgn:
> I think we need to modify patch-charts.sh so these changes aren't lost in a future change.

+1, either that or modify the `maistra-1.0` branch of maistra/istio to fix them, that's where these charts are pulled in from

luksa:
Yes, that's exactly what this PR is :)

luksa:
Oh, s**t. For some reason, I thought they were pulled in from here. I get confused with all these repos and copying from one to another. I'll submit a PR against Maistra/istio.

luksa:
Ah, nevermind. Rob already patches the charts in `patch-container-image.sh`. Closing this.

PR #413, opened 2020-04-23
MAISTRA-1373 default resource request for oauth-proxy container

a7vicky:


dgn:
Hi @garbagego - appreciate the effort! You'll however notice that the tests will fail on this; these charts are generated, from a mixture of (in this order):
- charts in repo maistra/istio
- charts in resources/overlays
- sed replacements done through scripts in `build/patch-charts.sh`

You'll find that the right place to edit this is in the patch scripts, that's where the oauth-proxy container is added to the deployments.

/ok-to-test

jwendell:
Also, I'm not sure if we should use `.Values.resources` or if we should create a new section in values to hold the oauth container... something like `.Values.oauth.resources`

mathianasj:
The title of this is a bit confusing, is this to address all pod requests or just the oauth-proxy sidecar?

a7vicky:
@dgn thanks for pointing to the right builder script. I will update the Prometheus and Grafana path script.
@jwendell I am not sure using `.Values.resources` is better approach or better to have `.Values.oauth.resources`

@mathianasj This only for oauth-proxy container which is required to have resource request set in deployment in case namespace have resource quota defined.

dgn:
@garbagego note that we already use `global.oauthproxy.hub` and `global.oauthproxy.tag`, so `global.oauthproxy.resources` would be a good fit

dgn:
Superseded by https://github.com/Maistra/istio-operator/pull/441

PR #416, opened 2020-04-27
OSSM-110 use ImageStream for oauth proxy image

rcernich:


build/patch-container-image.sh
```diff
@@ -27,6 +27,28 @@ function patch_container_image() {
...
+\2image: {{ .Values.global.oauthproxy.hub }}/{{ .Values.global.oauthproxy.image }}:{{ .Values.global.oauthproxy.tag }}\
+\{\{- end \}\}\
+\1+
+            }
+        }' "${file}"
+    else
+        echo "ERROR: file does not exist: ${file}"
+        return 1
+    fi
+}
```
dgn, line 50: IIRC we're injecting oauth-proxy in another sed call, wouldn't it be easier to change it there?

rcernich: I updated it in those scripts.  This still needs to hang around to patch the old 1.0 charts, unfortunately.

knrc (approved):

knrc (approved):

brian-avery (approved):

build/patch-container-image.sh
dgn: gotcha, missed that. Thanks!

PR #419, opened 2020-04-30
MAISTRA-1371 refactor version customization

rcernich:

Forgive the scope of the changes, but I needed to address some import cycles in the common package.


rcernich:
This moves all version specific logic into the versions package.  No more `switch version...` statements littered throughout.

pkg/controller/common/helm/helm.go
```diff
@@ -15,65 +15,52 @@ import (
...
-	DefaultTemplatesDir string `json:"defaultTemplatesDir,omitempty"`
-	// TemplatesDir is the base dir to user supplied templates files.
-	UserTemplatesDir string `json:"userTemplatesDir,omitempty"`
-}
-
-var Options = &Config.Rendering
-
 // GetChartsDir returns the location of the Helm charts. Similar layout to istio.io/istio/install/kubernetes/helm.
-func (o *renderingOptions) GetChartsDir(maistraVersion string) string {
+func GetChartsDir(maistraVersion string) string {
```
dgn: nit: why not make this type-safe and use `versions.Version` while we're at? I know technically the ValidatingWebhook will make sure we don't get an invalid version here, but it's more elegant imo. Also, we could avoid the "defaulting to v1.0" in this func if we move it to `ParseVersion()`, see other comment

pkg/controller/versions/versions.go
```diff
@@ -0,0 +1,167 @@
...
+
+	v1 "github.com/maistra/istio-operator/pkg/apis/maistra/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// InvalidVersion is not a valid version
+	InvalidVersion version = iota
+	// UndefinedVersion is...undefined
+	UndefinedVersion
```
dgn: Should we maybe remove `UndefinedVersion`? In the end it always means `V1_0` - if we return that in `ParseVersion()` then we don't have to duplicate that logic everywhere else

dgn (approved):
lgtm, but I'd prefer moving the defaulting logic to a central location (in `versions`)

luksa (approved):

pkg/controller/versions/versions.go
rcernich: The issue is that it means `DefaultVersion` if it's seen in the mutating webhook, and `LegacyVersion` in the processing.  I suppose mutating webhook could check for an empty version string instead of `UndefinedVersion`.

pkg/controller/common/helm/helm.go
rcernich: Makes sense.  I'll update the PR.

pkg/controller/versions/versions.go
rcernich: Empty version now returns V1_0.  It was much easier to check the field for empty string in the mutating webhook than to have the code littered with `if version == versions.UndefinedVersion` all over the place.

PR #420, opened 2020-04-30
MAISTRA-1419 use new smcp when validating upgrade/downgrade

rcernich:

### DIFF
--- a/pkg/controller/servicemesh/webhooks/validation/controlplane.go
+++ b/pkg/controller/servicemesh/webhooks/validation/controlplane.go
@@ -146,13 +146,13 @@ func (v *ControlPlaneValidator) validateUpdate(ctx context.Context, old, new *ma
 	// fails because feature X is no longer supported, but was added back in 1.3).
 	if oldVersion.Version() < newVersion.Version() {
 		for version := oldVersion.Version(); version < newVersion.Version(); version++ {
-			if err := v.validateUpgrade(ctx, version, old); err != nil {
+			if err := v.validateUpgrade(ctx, version, new); err != nil {
 				return validationFailedResponse(http.StatusBadRequest, metav1.StatusReasonBadRequest, fmt.Sprintf("cannot upgrade control plane from version %s to %s: %s", oldVersion.String(), newVersion.String(), err))
 			}
 		}
 	} else {
 		for version := oldVersion.Version(); version > newVersion.Version(); version-- {
-			if err := v.validateDowngrade(ctx, version, old); err != nil {
+			if err := v.validateDowngrade(ctx, version, new); err != nil {
 				return validationFailedResponse(http.StatusBadRequest, metav1.StatusReasonBadRequest, fmt.Sprintf("cannot downgrade control plane from version %s to %s: %s", oldVersion.String(), newVersion.String(), err))
 			}
 		}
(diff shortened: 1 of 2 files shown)


pkg/controller/servicemesh/webhooks/validation/controlplane.go
```diff
@@ -146,13 +146,13 @@ func (v *ControlPlaneValidator) validateUpdate(ctx context.Context, old, new *ma
 	// fails because feature X is no longer supported, but was added back in 1.3).
 	if oldVersion.Version() < newVersion.Version() {
 		for version := oldVersion.Version(); version < newVersion.Version(); version++ {
-			if err := v.validateUpgrade(ctx, version, old); err != nil {
+			if err := v.validateUpgrade(ctx, version, new); err != nil {
```
dgn, line 149: I'm not sure I understand this part. If I'm reading the comment right, shouldn't we be validating an upgrade from `version -> version+1`? For that, we'd need to set `new.Spec.Version = version+1` in every loop iteration

pkg/controller/servicemesh/webhooks/validation/controlplane.go
```diff
@@ -146,13 +146,13 @@ func (v *ControlPlaneValidator) validateUpdate(ctx context.Context, old, new *ma
...
 				return validationFailedResponse(http.StatusBadRequest, metav1.StatusReasonBadRequest, fmt.Sprintf("cannot upgrade control plane from version %s to %s: %s", oldVersion.String(), newVersion.String(), err))
 			}
 		}
 	} else {
 		for version := oldVersion.Version(); version > newVersion.Version(); version-- {
-			if err := v.validateDowngrade(ctx, version, old); err != nil {
+			if err := v.validateDowngrade(ctx, version, new); err != nil {
```
dgn, line 155: same as above

luksa (approved):

pkg/controller/servicemesh/webhooks/validation/controlplane.go
luksa: Since this code is in the 1.1 branch, the for loop is misleading, since there are only two possible versions. The body of the loop is only executed once.

Also, the validation code never looks at `.spec.version`, so there's no need to update it here.

What this PR fixes is the scenario where the user does two things in the same update operation:
- they change the version, and
- make the required changes to the SMCP so that its spec is valid for the _new_ version. 

Previously, users would have to make the changes to the SMCP first and only then change the version.

dgn: Ah, I get it. Thanks @luksa 

I guess we'll need to implement a `validateUpgrade(fromVersion, toVersion, smcp)` func for maistra-1.2 then

dgn (approved):

brian-avery (approved):

pkg/controller/servicemesh/webhooks/validation/controlplane.go
rcernich: This has changed with the version refactoring in maistra-1.2.  Each version strategy will validate downgrade from it to the previous version and upgrade from the previous version to it, e.g. 1.2 strategy would validate upgrade from 1.1 and downgrade to 1.1.  The looping logic persists in 1.2, but as @luksa said, with only two versions, it only loops through once.

PR #422, opened 2020-05-04
MAISTRA-1370 Add OpenAPI schema to Maistra CRDs

luksa:


build/generate-manifests.sh
```diff
@@ -41,37 +41,57 @@ function checkDependencies() {
     echo "Please install yq package, e.g. 'pip install --user yq'"
     exit 1
   fi
-}
 
-function generateServiceMeshControlPlanesCrd() {
-  yq -s -y --indentless '.[] | select(.kind=="CustomResourceDefinition" and .metadata.name=="servicemeshcontrolplanes.maistra.io") | .' ${DEPLOYMENT_FILE} > ${BUNDLE_DIR}/servicemeshcontrolplanes.crd.yaml
+  if ! [ -x "$(command -v operator-sdk)" ]; then
```
rcernich: As we have a dependency on operator-sdk, could we just `go run ...`?

pkg/apis/maistra/v1/helmvalues.go
```diff
@@ -0,0 +1,134 @@
...
+				return nil, false, nil
+			}
+		} else {
+			return nil, false, fmt.Errorf("%v accessor error: %v is of the type %T, expected map[string]interface{}", strings.Join(fields[:i+1], "."), val, val)
+		}
+	}
+	return val, true, nil
+}
+
+func (h *HelmValues) GetBool(path string) (bool, bool, error) {
```
rcernich, line 34: Would it be better to use the code in `unstructured` for this?  Saves us having to maintain similar code.

build/generate-manifests.sh
```diff
@@ -41,37 +41,57 @@ function checkDependencies() {
...
-function generateServiceMeshMemberRollsCrd() {
-  yq -s -y --indentless '.[] | select(.kind=="CustomResourceDefinition" and .metadata.name=="servicemeshmemberrolls.maistra.io") | .' ${DEPLOYMENT_FILE} > ${BUNDLE_DIR}/servicemeshmemberrolls.crd.yaml
+function generateCRDs() {
+  operator-sdk generate crds --crd-version v1
+  mv deploy/crds/maistra.io_servicemeshcontrolplanes_crd.yaml ${BUNDLE_DIR}/servicemeshcontrolplanes.crd.yaml
+  mv deploy/crds/maistra.io_servicemeshmemberrolls_crd.yaml ${BUNDLE_DIR}/servicemeshmemberrolls.crd.yaml
+  mv deploy/crds/maistra.io_servicemeshmembers_crd.yaml ${BUNDLE_DIR}/servicemeshmembers.crd.yaml
+  rm -rf deploy/crds/
+
+  cat ${BUNDLE_DIR}/servicemeshcontrolplanes.crd.yaml >deploy/src/crd.yaml
```
rcernich, line 53: Is there a better way to do this?  `oc apply -f` can work off a directory, so is it necessary to merge all this into a single file?  (Maybe that's required to support the manifest generation, if so, this is an incremental step toward a more conventional operator-sdk structured project?)

build/generate-manifests.sh
luksa: I'm not sure how one can add a dependency to `operator-sdk/cmd`. We currently don't have that package in `vendor/`. Any ideas?

pkg/apis/maistra/v1/helmvalues.go
luksa: You mean use `unstructured` instead of `HelmValues`? Or just call `unstructured.NestedBool` in the `HelmValues' GetBool` function? That might be doable. I wanted to simplify stuff, but I see I can keep the simplifications, while still calling the existing unstructured code.

build/generate-manifests.sh
luksa: Yeah, I wanted to keep the existing files for now. If we move to the `apply -f dir/` solution, we'll need to split the `src/` dir into two - `maistra` and `servicemesh`.

build/generate-manifests.sh
rcernich: Look at `tools.go`  I think you can add an import there for the tools.  I think this might be how the newer versions of operator-sdk setup the project anyway.

pkg/apis/maistra/v1/helmvalues.go
rcernich: I was thinking just implement `HelmValues.GetBool()` using the unstructured helpers.

build/generate-manifests.sh
rcernich: Sounds good.

build/generate-manifests.sh
luksa: Thanks! Importing it works, but until we upgrade the operator-sdk, we can't use it like this. So, I'm leaving things as they for now.

pkg/apis/maistra/v1/helmvalues.go
luksa: Fixed. Thanks!

dgn:
please ignore the release job; that's part of my `feature builds` work. sorry for the trouble, I made it opt-in now

build/generate-manifests.sh
luksa: I've now fixed this. `operator-sdk` is run via `go run`.

manifests-servicemesh/1.2.0/servicemeshoperator.v1.2.0.clusterserviceversion.yaml
```diff
@@ -597,9 +584,6 @@ spec:
                 - name: operator-olm-config
                   mountPath: /etc/operator/olm
                   readOnly: true
-                - name: webhook-tls-volume
```
rcernich: I think these might have been missed when you split out operator.yaml.

rcernich:
Is there still a way to do something like `oc create -f deploy/` to manually install the operator for either service mesh or maistra?

manifests-servicemesh/1.2.0/servicemeshoperator.v1.2.0.clusterserviceversion.yaml
luksa: Ugh, yeah, thanks! Looks like I didn't notice these changes when I rebased today.  

I've now reviewed the files again to make sure nothing else has disappeared.

luksa:
> Is there still a way to do something like `oc create -f deploy/` to manually install the operator for either service mesh or maistra?

Yes, `oc create -f deploy/maistra-operator.yaml` works the same as before, but the yaml files are now built from the files in `deploy/src`.

rcernich (approved):

dgn (approved):

PR #427, opened 2020-05-06
MAISTRA-1431 ControlPlaneStatus.ComponentStatus array shouldn't be an array of pointers

luksa:


luksa:
This PR also contains #422, which must be merged first.

rcernich (approved):

jwendell (commented):
Isn't it better to have them as pointers to avoid copies?

luksa:
> Isn't it better to have them as pointers to avoid copies?

Semantically they shouldn't be pointers. Also, the `FindComponentByName()` function will still return a pointer to the array element, since the existing code expects that modifications of the object will be applied to the array element. We never really copy the object except when making a copy of the whole SMCP. But then we always make a deep copy, so even if we used a pointer, the object would be copied.

dgn:
please ignore the release job; that's part of my `feature builds` work. sorry for the trouble, I made it opt-in now

luksa:
@jwendell can you review this again? thanks

jwendell (approved):

PR #433, opened 2020-05-13
MAISTRA-1472 Feature builds

dgn:


dgn:
Images are available at
```
hub: quay.io/maistra-dev
tag: 1.2.8df33368d2039245d1d2a693ce0c1d79754b09fc
```

deps.sh
```diff
@@ -0,0 +1,26 @@
+# when specifying a commit that is not on the main repository yet but in a PR,
+# you'll have to specify the GitHUB PR number so that we can fetch it 
+
+export RPMS_COMMIT=
+export RPMS_PR=97
```
dgn: as you can see I'm referencing unmerged PRs here; these are the draft PRs that contain the changes necessary to make all this magically work

brian-avery:
Can we move this to an independent repository? Both the RPM builds and image builds are uncoupled from any one component. Seems odd to couple this one to the operator.

dgn:
> Can we move this to an independent repository? Both the RPM builds and image builds are uncoupled from any one component. Seems odd to couple this one to the operator.

We could. I just wanted to avoid people having to create additional PRs to use this; most more complex changes will need an operator change anyway (if you change the charts in Maistra/istio for example, you need an operator PR to pull them in). It's not a lot of code, really, but if people want a separate repo that works for me, too.

dgn:
COPR doesn't want to build :(

brian-avery:
Version changes require operator changes, but there's quite a bit that don't require operator changes right? I think there's quite a lot that can change in Envoy, CNI, Proxy, etc that don't require a change to the operator. It also bleeds building everything into the concerns of the operator.

build/maistra-release.sh
```diff
@@ -0,0 +1,62 @@
...
+    rm -rf ${WORKDIR}
+}
+
+pushd ${WORKDIR}
+
+trap cleanup EXIT
+
+# build RPMs
+
+git clone https://github.com/Maistra/rpms.git
```
brian-avery: I think Jonh is working on breaking the rpms repo apart.

dgn: Yes, that's coming down the line

dgn:
> Version changes require operator changes, but there's quite a bit that don't require operator changes right? I think there's quite a lot that can change in Envoy, CNI, Proxy, etc that don't require a change to the operator. It also bleeds building everything into the concerns of the operator.

Fully agree. I can move this script into `test-infra`, that probably makes more sense anyway. However, the main reason this is currently triggered from here is another:

Building full image sets from every possible repo is hard, and this is the first iteration of this work.

It is absolutely planned to have the possibility to trigger full builds from any repo, but we have to start somewhere, and for now, that'll be `istio-operator`. Please feel free to help with setting this up to work from other repos.

jwendell:
@dgn What's the status of this? Safe to close?

dgn:
closing as we'll likely do this another way now that we decided to do multi-stage builds for community

PR #439, opened 2020-05-26
MAISTRA-1494: Add support for the metadata version

knrc:
I still have to test this PR, will do that tomorrow but wanted to get other eyes on the changes.

This PR adds knowledge of the metadata version into the operator's deployment so we can use this as the basis for making decisions on whether a reconciliation is necessary.  The current code uses the operator build version which, with multiple streams, is no longer guaranteed to change.


luksa (approved):

knrc:
I'm going to merge this and do a build, if there are improvements/changes to be made I'll submit a follow-up PR.

PR #442, opened 2020-05-26
MAISTRA-1496 Ensure lastTransitionTime isn't updated when no transition occurs

luksa:
Resetting the renderings to nil caused the operator to set all the
component conditions to false in the next reconciliation attempt (due
to the renderings being nil). The operator then changed the conditions
of all the components that are already deployed back to true, resetting
the lastTransitionTime in the process, since the condition went from
false to true. This then caused the SMCP status to be updated, which
then caused the operator to go into a tight loop, as the status update
watch event triggered another reconciliation and by-passed the back-off
mechanism.

There's no need to reset the chart renderings if an error occurs when
we try to deploy them, as the renderings will be identical the next time
we render them. If the error occurs during the rendering itself, the
renderings won't even be set, so we don't have to set the field to nil
in that case either.

### DIFF
--- a/pkg/controller/servicemesh/controlplane/reconciler.go
+++ b/pkg/controller/servicemesh/controlplane/reconciler.go
@@ -116,7 +116,6 @@ func (r *controlPlaneInstanceReconciler) Reconcile(ctx context.Context) (result 
 		// error handling
 		defer func() {
 			if err != nil {
-				r.renderings = nil
 				r.lastComponent = ""
 				updateReconcileStatus(&r.Status.StatusType, err)
 			}
--- a/pkg/apis/maistra/v1/status.go
+++ b/pkg/apis/maistra/v1/status.go
@@ -185,12 +185,12 @@ func (s *StatusType) SetCondition(condition Condition) *StatusType {
 	// match the time in our cached status during a reconcile.  We truncate here
 	// to save any problems down the line.
 	now := metav1.NewTime(time.Now().Truncate(time.Second))
-	for i := range s.Conditions {
-		if s.Conditions[i].Type == condition.Type {
-			if s.Conditions[i].Status != condition.Status {
+	for i, prevCondition := range s.Conditions {
+		if prevCondition.Type == condition.Type {
+			if prevCondition.Status != condition.Status {
 				condition.LastTransitionTime = now
 			} else {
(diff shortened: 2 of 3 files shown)


pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -116,7 +116,6 @@ func (r *controlPlaneInstanceReconciler) Reconcile(ctx context.Context) (result
 		// error handling
 		defer func() {
 			if err != nil {
-				r.renderings = nil
```
rcernich, line 119: Will the renderings be regenerated?

luksa: Do they need to be? Wouldn't the regenerated renderings be identical? AFAICT the line I removed only caused the renderings to be regenerated when an error occurred during creation of the resources in these renderings. The next reconcile attempt generated the exact same renderings.

Of course, if a user updates the SMCP, its generation changes and the renderings are then set to nil here: https://github.com/Maistra/istio-operator/blob/53de3522d4b4297de7e6ba6b0b2740da2a211448/pkg/controller/servicemesh/controlplane/reconciler.go#L687

rcernich (approved):

pkg/controller/servicemesh/controlplane/reconciler.go
rcernich: Okay.  I thought I put that in for some reason, e.g. some other place checking to see if `renderings == nil`, but maybe that's what's causing the issue you're seeing (i.e. it was used to revert the component conditions).

dgn:
> @luksa: The following test **failed**, say `/retest` to rerun all failed tests:
> gen-check 	[1b8cf45](https://github.com/Maistra/istio-operator/commit/1b8cf45e3842325d0220d21bf4ba661442a41fb8) 	[link](https://prow.maistra.io/view/gcs/maistra-prow-testing/pr-logs/pull/Maistra_istio-operator/442/gen-check/1265280421558489089) 	`/test gen-check`

test failure will resolve itself once https://github.com/Maistra/istio-operator/pull/443 is merged (needs rebase)

rcernich:
@Mergifyio refresh

dgn (approved):

PR #444, opened 2020-05-26
MAISTRA-1494: Use the sanitised version for reconciliation

knrc:


knrc:
Merging by hand to rebuild the metadata

PR #447, opened 2020-05-27
MAISTRA-1500 Fix race condition in post-install readiness check

luksa:
After deploying a component, if the cache isn't updated in time, we never
wait for the component to become ready (its has no entry in the
readinessMap; we consider this to mean that the chart has no resources
with readiness information, so we consider the whole chart as ready).

If the component is missing in readinessMap, we must take that as a signal
that the component is NOT ready. But if we do that, we'll take all components
that have no resources with readiness information as not ready.

So, before checking readiness, we should determine if the chart has any
resources that contain readiness information. If it doesn't we should
count the component as ready.

### DIFF
--- a/pkg/controller/servicemesh/controlplane/readiness.go
+++ b/pkg/controller/servicemesh/controlplane/readiness.go
@@ -100,12 +100,20 @@ func (r *controlPlaneInstanceReconciler) updateReadinessStatus(ctx context.Conte
 
 type isReadyFunc func(runtime.Object) bool
 
+// keep this in sync with kinds in calculateComponentReadiness()
+var kindsWithReadiness = sets.NewString("Deployment", "StatefulSet", "DaemonSet")
+
+func (r *controlPlaneInstanceReconciler) hasReadiness(kind string) bool {
+	return kindsWithReadiness.Has(kind)
+}
+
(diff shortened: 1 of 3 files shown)


pkg/controller/servicemesh/controlplane/readiness.go
```diff
@@ -100,12 +100,20 @@ func (r *controlPlaneInstanceReconciler) updateReadinessStatus(ctx context.Conte
...
+	return kindsWithReadiness.Has(kind)
+}
+
 func (r *controlPlaneInstanceReconciler) calculateComponentReadiness(ctx context.Context) (map[string]bool, error) {
 	readinessMap := map[string]bool{}
 	typesToCheck := []struct {
 		list  runtime.Object
 		ready isReadyFunc
 	}{
+		// keep this in sync with kindsWithReadiness
```
rcernich, line 116: Should we create a simple type for this?  Maybe make `kindsWithReadiness` a map, so that all of the related code is in one place?

luksa: Yeah, I tried that, but found no simple solution, because we need a List-type in one place, and the base type in the other. And if I have to add type manipulation code, it's just going to be much more complicated that it is right now. On the other hand, this map and the set won't really change much (if ever), so I don't know if it makes sense to waste time improving this.

rcernich (approved):

dgn (approved):

PR #449, opened 2020-05-28
Configure mergify to add "ready to merge" label to PRs

luksa:
This is to make it easy for us to see which PRs have been approved
and are only missing the "okay to merge" label.

This should also make it easy to see which PRs still need at least
one additional approval (PRs that aren't marked as WIP and have no
"ready to merge" label).

I experimented with also adding a "needs review" label, but then
the PR discussion would have a bunch of 'added/removed "needs review"'
events, because switching from WIP to nonWIP and approvals & their
revocations would cause the label to be added/removed.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -14,6 +14,9 @@ pull_request_rules:
       merge:
         method: squash
         strict: smart
+      label:
+        remove:
+        - "ready to merge"
   - name: automatic merge (without squash) with strict when reviewed and not work in progress
     conditions:
       - '#approved-reviews-by>=2'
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -29,19 +32,34 @@ pull_request_rules:
...
         method: rebase
         strict: smart
+      label:
+        remove:
+        - "ready to merge"
   - name: label work in progress PRs
     conditions:
       - label != "work in progress"
-      - title ~= '^WIP\b'
+      - title ~=^WIP\b
```
luksa, line 41: I had to change this, since mergify no longer likes the previous format. It wouldn't trigger these actions as the regex apparently no longer matched.

rcernich (approved):

dgn (approved):

luksa:
Merging manually, since Mergify refuses to merge its own config change.

PR #452, opened 2020-06-01
MAISTRA-1504 Fix import k8s.io/kubernetes/pkg/apis/rbac

luksa:


go.mod
```diff
@@ -25,7 +25,6 @@ require (
 	k8s.io/code-generator v0.17.1
 	k8s.io/helm v2.16.1+incompatible
 	k8s.io/kubectl v0.0.0
-	k8s.io/kubernetes v1.16.2
```
jwendell, line 28: It would be good if you run `go mod tidy` and commit the results of go.sum as well.

luksa: I have run it, but it produces no changes in go.sum.

luksa: I have run it, but it produces no changes in go.sum.

rcernich (approved):

go.mod
jwendell: ah ok, it must be other modules depending on it.

jwendell (approved):

PR #457, opened 2020-06-02
MAISTRA-1496 Ensure lastTransitionTime isn't updated when no transition occurs

luksa:
Resetting the renderings to nil caused the operator to set all the
component conditions to false in the next reconciliation attempt (due
to the renderings being nil). The operator then changed the conditions
of all the components that are already deployed back to true, resetting
the lastTransitionTime in the process, since the condition went from
false to true. This then caused the SMCP status to be updated, which
then caused the operator to go into a tight loop, as the status update
watch event triggered another reconciliation and by-passed the back-off
mechanism.

There's no need to reset the chart renderings if an error occurs when
we try to deploy them, as the renderings will be identical the next time
we render them. If the error occurs during the rendering itself, the
renderings won't even be set, so we don't have to set the field to nil
in that case either.

### DIFF
--- a/pkg/apis/maistra/v1/status.go
+++ b/pkg/apis/maistra/v1/status.go
@@ -185,12 +185,12 @@ func (s *StatusType) SetCondition(condition Condition) *StatusType {
 	// match the time in our cached status during a reconcile.  We truncate here
 	// to save any problems down the line.
 	now := metav1.NewTime(time.Now().Truncate(time.Second))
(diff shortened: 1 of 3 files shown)


luksa:
Note: This is a cherry-pick of #442 into maistra-1.2

rcernich (approved):

jwendell (approved):

PR #460, opened 2020-06-05
MAISTRA-1374: support multiple multus CNI interfaces

knrc:
This needs to be updated once https://github.com/Maistra/istio/pull/117 is merged

### DIFF
--- a/build/patch-charts.sh
+++ b/build/patch-charts.sh
@@ -256,7 +256,7 @@ function patchSidecarInjector() {
 
   # add annotation for Multus & Istio CNI
   sed_wrap -i -e 's/^\(.*injectedAnnotations:.*\)$/\1\
-    \{\{- if .Values.istio_cni.enabled \}\}\
+    \{\{- if and (.Values.istio_cni.enabled) (not .Values.sidecarInjectorWebhook.injectPodRedirectAnnot) \}\}\
       k8s.v1.cni.cncf.io\/networks: \{\{.Values.istio_cni.istio_cni_network\}\}\
     \{\{- end \}\}/' ${HELM_DIR}/istio/templates/sidecar-injector-configmap.yaml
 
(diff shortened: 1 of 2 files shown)


build/patch-charts.sh
```diff
@@ -254,12 +254,6 @@ function patchSidecarInjector() {
       - SETUID
   }' ${HELM_DIR}/istio/files/injection-template.yaml
 
-  # add annotation for Multus & Istio CNI
-  sed_wrap -i -e 's/^\(.*injectedAnnotations:.*\)$/\1\
-    \{\{- if .Values.istio_cni.enabled \}\}\
-      k8s.v1.cni.cncf.io\/networks: \{\{.Values.istio_cni.istio_cni_network\}\}\
-    \{\{- end \}\}/' ${HELM_DIR}/istio/templates/sidecar-injector-configmap.yaml
```
dgn, line 261: I'm afraid we still need this, as the injection of `podRedirectAnnot` is disabled by default in 1.1. We should replace the check for `.Values.istio_cni.enabled` (as that'll always be true with Maistra) with a check for `not .Values.sidecarInjectorWebhook.injectPodRedirectAnnot`

cc @luksa

dgn (changes requested):

build/patch-charts.sh
knrc: I thought we had recently enabled this, I'll double check today and fix

knrc: I've verified it's not enabled by default, apologies for this.  I'll rework the fix.

knrc: @dgn Can you please take another look?

dgn (approved):

build/patch-charts.sh
dgn: LGTM, thanks

rcernich (approved):

PR #461, opened 2020-06-05
MAISTRA-1523 Install some control plane components in parallel

luksa:
The "security" and "galley" have a circular dependency between each other.
It makes sense to deploy them at the same time. Also, deploying some
components in parallel will speed up the installation process.


rcernich:
This looks okay, but I think it would be better to defer installation of istio.io resources until after galley is up and running.  I think `security` needs to be broken up into `citadel` and `mtls`, as it's the mtls bit which causes the problems.  Ideally, any component which installs istio.io resources should be installed after galley is up and running to avoid issues.

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -32,31 +32,27 @@ import (
...
-	"istio/charts/galley",
-	"istio/charts/mixer",
-	"istio/charts/pilot",
-	"istio/charts/gateways",
-	"istio/charts/sidecarInjectorWebhook",
-	"istio/charts/grafana",
-	"istio/charts/kiali",
+var orderedCharts = [][]string{
+	{"istio"}, // core istio resources
+	{"istio/charts/security", "istio/charts/galley"}, // both are deployed at the same time because each is a dependency of the other
```
dgn: I agree with @rcernich that this one could be problematic. If the webhookconfig from `galley` is already present when the istio.io resources from `security` are deployed, but the `galley` pod isn't ready yet, we're going to see errors

dgn: I just realized that from 1.6 onwards upstream does not deploy istio.io configuration on installation anymore, so this problem will likely disappear. I suggest we postpone the merge of security and galley until then

rcernich: We'll still need this for backward compatibility.

luksa:
> This looks okay, but I think it would be better to defer installation of istio.io resources until after galley is up and running. I think `security` needs to be broken up into `citadel` and `mtls`, as it's the mtls bit which causes the problems. Ideally, any component which installs istio.io resources should be installed after galley is up and running to avoid issues.

Yeah, I agree. I've reverted the part where the security and galley are installed at the same time so that they are now installed as they were before. I've left the other charts to be installed in parallel.

dgn (approved):

rcernich (approved):
lgtm

I assume this does not address the issue with security and galley, where security install would fail if validating webhook was installed, but galley deployment failed to come up (okay, because this was an existing issue; just want to confirm)

luksa:
> lgtm
> 
> I assume this does not address the issue with security and galley, where security install would fail if validating webhook was installed, but galley deployment failed to come up (okay, because this was an existing issue; just want to confirm)

Yup, this needs to be addressed in a different PR. Or not, since we're switching to istiod anyway.

PR #462, opened 2020-06-11
MAISTRA-1071 Add manifest namespace validation

dgn:
This is two parts:
- validatingWebhook code to give users early and comprehensible feedback on why they can't use a namespace
- reconciler.validateManifests function that makes sure we'll never deploy resources into namespaces that are not part of the mesh


pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -646,6 +656,56 @@ func (r *controlPlaneInstanceReconciler) renderCharts(ctx context.Context) error
...
+				// if it doesn't have a metadata section, ignore
+				continue
+			}
+			objNs, ok := metadata["namespace"].(string)
+			if !ok {
+				// if namespace is not set, ignore
+				continue
+			}
+
+			for _, ns := range meshNamespaces {
```
rcernich: Would it be easier to return the set and just use that?

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -646,6 +656,56 @@ func (r *controlPlaneInstanceReconciler) renderCharts(ctx context.Context) error
...
 
+func (r *controlPlaneInstanceReconciler) validateManifests(ctx context.Context) error {
+	log := common.LogFromContext(ctx)
+	allErrors := []error{}
+	// validate resource namespaces
+	smmr := &v1.ServiceMeshMemberRoll{}
+	if err := r.Client.Get(context.TODO(), client.ObjectKey{Namespace: r.Instance.GetNamespace(), Name: common.MemberRollName}, smmr); err != nil {
+		if !apierrors.IsNotFound(err) {
+			// log error, but don't fail validation: we'll just assume that the control plane namespace is the only namespace for now
+			log.Error(err, "failed to retrieve SMMR for SMCP")
```
rcernich: It's probably worth returning the error and letting the reconciliation restart.  If there's a gateway that references a namespace that is legitimate, then we'll error when we validate, which will be a false error message.

pkg/controller/servicemesh/controlplane/reconciler.go
dgn: Good point

pkg/controller/servicemesh/controlplane/reconciler.go
dgn: So you want to restart reconciliation without logging an error in this case?

dgn: The thing is that this will probably only happen rarely, and even if it happens, it will only have an impact if you have a gateway outside the control plane namespace. The idea with just logging and continuing was that 99% of the time it'll be alright, because the gateway is in the cp namespace anyway

rcernich: I believe you can just return the error from the reconcile loop and let it do its thing.  I don't remember if we need to log it and return it, or just return it.  @luksa?

rcernich: I agree that it's probably unlikely that it will fail and unlikely that a namespace other than the control plane will be used.  Maybe keep track of the fact that the smmr lookup failed and then _if_ there's a validation failure, you can return the smmr error instead. wdyt?

dgn: Yeah, I think that makes sense

dgn: I fixed it now. While at it, I found another issue; I had forgotten to split the manifests (by "---") so was only looking at the first manifest in the output of each helm template (which for the gateways are multiple).

dgn:
FYI I ran the validation function 1k times and it takes 21ms on average on my machine, so should be okay

rcernich (approved):

pkg/controller/servicemesh/controlplane/reconciler_test.go
```diff
@@ -171,6 +173,181 @@ func TestInstallationErrorDoesNotUpdateLastTransitionTimeWhenNoStateTransitionOc
...
+					},
+				},
+			},
+		},
+	}
+
+	operatorNamespace := "istio-operator"
+	InitializeGlobals(operatorNamespace)()
+
+	for _, tc := range testCases {
```
luksa, line 296: Missing `t.Run(tc.name, func(t *testing.T) {` ?

luksa (approved):
LGTM but I'm wondering why we need to consider the member namespaces when checking the manifests. Aren't all resources deployed in the control plane namespace? If yes, then we could perform the validation when we render the charts.

dgn:
> LGTM but I'm wondering why we need to consider the member namespaces when checking the manifests. Aren't all resources deployed in the control plane namespace? If yes, then we could perform the validation when we render the charts.

The gateways are an exception in that they can be deployed to any namespace by providing a 
`istio.gateways.some-gateway.namespace: someNamespace`; this is meant to allow for multiple ingress/egress gateways in separate namespaces.

The reason I added this in a generic fashion is that things like this could come up again in future versions of Istio, and it might not just be restricted to gateways

pkg/controller/servicemesh/controlplane/reconciler_test.go
dgn: Good point, fixed now

dgn:
Adding a hold as this will go into 1.1.5

PR #467, opened 2020-06-26
MAISTRA-1575 create new SMCP structure for Istio 1.6 

rcernich:
New SMCP structure that better organizes (I hope) configuration settings.


rcernich:
The first cut at a new structure for SMCP is complete.  Still to do:

* Verify installation still works with 1.0/1.1
* Integrate istio 1.6 charts
* Write conversion code from v1 to v2
* Add conversion webhook

pkg/apis/maistra/conversion/gateways.go
```diff
@@ -0,0 +1,276 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	v2 "github.com/maistra/istio-operator/pkg/apis/maistra/v2"
+	"github.com/maistra/istio-operator/pkg/controller/versions"
+)
+
+const (
+	imageNameSDS = "node-agent-k8s"
```
dgn, line 15: what's this used for? upstream 1.5+ does not have this anymore, and we never shipped it

rcernich:
@dgn, here's a link to some example yaml for the new structure: https://github.com/maistra/istio-operator/blob/6121c576086fbe56edb04f957f8fff24193d657c/pkg/apis/maistra/v2/smcp_new.yaml

I've also sprinkled the code with `XXX` comments where I'm unsure about things or need to come back and fix/decide what to do (e.g. maybe some settings don't map to values.yaml, but should be supported).

rcernich:
I'm currently using enabled = !null in a lot of places, and I'm not sure whether or not I should include an explicit "enabled" field.

rcernich:
SMCP reconciler has been converted to use v2 resource and all unit tests are passing.  I think there are some differences in the way things are defaulted, e.g. gateways.ingress and gateways.egress must be specified, as should policy, telemetry, and the various addons, which I think were all enabled by default.  (See my comment above about having a separate `enabled` field, as opposed to using `enabled = field != nil`.)

rcernich:
This is working for v1.1 installs.

knrc (approved):

manifests-maistra/1.2.0/servicemeshcontrolplanes.crd.yaml
```diff
@@ -101,6 +101,8935 @@ spec:
...
+                  information is to make it available to kubectl, which does not yet
+                  allow counting objects in JSONPath expressions.
+                type: object
+              components:
+                items:
+                  description: ComponentStatus represents the status of an object
+                    with children
+                  properties:
+                    children:
+                      description: 'TODO: can we remove this? it''s not used anywhere
```
brian-avery, line 123: Are you planning on removing hte TODO before merge?

rcernich: I anticipate a couple of rounds of cleanup, so not at this point.  I need to get this in so we can start getting 2.0 installations going.  This is a pretty big change and needs a lot of eyeballs and a proper design review.  I don't anticipate this being the final revision, but a starting point to get it dialed in.

pkg/apis/maistra/v2/smcp_new.yaml
```diff
@@ -0,0 +1,410 @@
...
+                termination: reencrypt
+          runtime:
+            deployment: {}
+            pod: {}
+
+    tracing:
+      type: Jaeger # or Zipkin, Lightstep, Datadog, Stackdriver
+      jaeger:
+        name: jaeger # name of Jaeger CR to reference/create
+        install: # create a jaeger CR if it doesn't already exist
```
dgn, line 353: how do you trigger installation? just by settings any values under `install`?

pkg/apis/maistra/v2/smcp_new.yaml
```diff
@@ -0,0 +1,410 @@
...
+            ingress:
+              contextPath: /grafana
+              tls:
+                termination: reencrypt
+          runtime:
+            deployment: {}
+            pod: {}
+      kiali:
+        name: kiali # kiali CR to create/reference
+        install: # install kiali CR if not present
```
dgn, line 396: I don't think we should nest the options underneath `install`, because every addon setting we provide will be an installation setting. maybe `install` should just be the bool that determines whether the addon is managed?

dgn (approved):
Thanks for starting this, Rob. I added a few comments but they don't have to be addressed in this PR. Generally, I think we should finalize the spec asap and then start working on documentation - because that'll become really important with this change

pkg/apis/maistra/conversion/gateways.go
rcernich: I was just trying to cover the full range of items and wasn't sure if this was something special for gateways.  It can be removed.

pkg/apis/maistra/v2/smcp_new.yaml
rcernich: Yes, if `install` is not null, maistra will create a CR with the settings.  If a CR already exists, no action will be taken.

pkg/apis/maistra/v2/smcp_new.yaml
rcernich: I'm trying to preserve the existing behavior, where the maistra operator installs and manages these things.  The difference is, once installed, maistra will only manage certain fields in the CR.  For example, in the case of Kiali, the settings for grafana, prometheus, and jaeger, and possibly the version (to ensure it is compatible with the control plane version).

PR #477, opened 2020-07-02
Add VMware NSX Container Plugin as CNI option

yasensim:
Hello,

I tried to file an issue but cannot find where, so I am sending PR. 
We are introducing VMware NSX Container Plugin support for Openshift 4. Without this PR service Mesh doesn't work on Openshift cluster when the CNI is NSX NCP.   
Could you please merge it?

Thanks in advance,
Yasen

### DIFF
--- a/pkg/controller/servicemesh/memberroll/namespace_reconciler.go
+++ b/pkg/controller/servicemesh/memberroll/namespace_reconciler.go
@@ -26,6 +26,7 @@ import (
 
 const networkTypeOpenShiftSDN = "OpenShiftSDN"
 const networkTypeCalico = "Calico"
+const networkTypeNsxNcp = "ncp"
 
 type namespaceReconciler struct {
 	common.ControllerResources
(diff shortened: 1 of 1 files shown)


pkg/controller/servicemesh/memberroll/namespace_reconciler.go
```diff
@@ -26,6 +26,7 @@ import (
 
 const networkTypeOpenShiftSDN = "OpenShiftSDN"
 const networkTypeCalico = "Calico"
+const networkTypeNsxNcp = "ncp"
```
brianredbeard, line 29: This should be a more descriptive value.  `ncp` would have no meaning to most users who are reviewing/attempting to understand a configuration file with a value like this.

yasensim: Thanks for reviewing @brianredbeard ! This is the value the certified cluster operator watches in the network crd during openshift bootstrapping. It means NSX Container Plugin.  Users that have NSX installed and plan to use it for Openshift know this abbreviation. Furthermore, we have already customers and changing this value might trigger issues in their clusters during upgrade. Please accept the constant name a more descriptive here.

salv-orlando: I understand the name might not tell much, but it's unfortunately the name of the networking solution "NSX Container Plugin" as Yasen wrote.

dgn:
Closing as we're now defaulting to NetworkPolicy for unknown CNI providers: https://github.com/maistra/istio-operator/pull/848

PR #479, opened 2020-07-14
MAISTRA-1071 Add manifest namespace validation

dgn:
Manual cherry-pick of https://github.com/maistra/istio-operator/pull/462


pkg/controller/versions/strategy_v1_1.go
```diff
@@ -80,6 +82,22 @@ func (v *versionStrategyV1_1) Validate(ctx context.Context, cl client.Client, sm
 			}
 		}
 	}
+	smmr := &v1.ServiceMeshMemberRoll{}
```
rcernich, line 100: Do we need this same check for the 1.0 and 1.2/2.0 validators?

rcernich (approved):

pkg/controller/versions/strategy_v1_1.go
dgn: Yes, but I wanted to wait for your adjustment (re)introducing the loop over gateways

dgn:
~Note that this is broken right now because I started work on it before the v2 stuff merged. Will rebase today~

I rebased the branch to be compatible with the v2 smcp changes

luksa (approved):

PR #480, opened 2020-07-17
 MAISTRA-1633 integrate charts from maistra/istio maistra-2.0

rcernich:
2.0 installs are stubbed out, but not tested (need 2.0 builds of control plane images).


Makefile
```diff
@@ -137,23 +141,25 @@ collect-1.1-templates:
...
-# maistra v1.2
+# maistra v2.0
 ################################################################################
-.PHONY: update-1.2-charts
-update-1.2-charts:
-	HELM_DIR=${RESOURCES_DIR}/helm/v1.2 ISTIO_VERSION=1.4.0 ${SOURCE_DIR}/build/download-charts.sh
-	HELM_DIR=${RESOURCES_DIR}/helm/v1.2 ${SOURCE_DIR}/build/patch-container-image.sh
+.PHONY: update-2.0-charts
+update-2.0-charts:
+	HELM_DIR=${RESOURCES_DIR}/helm/v2.0 ISTIO_VERSION=1.6.0 ${SOURCE_DIR}/build/download-charts.sh
```
jwendell, line 148: The accuracy of the value of `ISTIO_VERSION` does not matter, right? I mean, we are on 1.6.5 now.

jwendell (approved):

Makefile
rcernich: Yeah, I don't think that's used anywhere.  We should remove it at some point.

knrc (approved):

Makefile
knrc: It's only used for the release directory

PR #489, opened 2020-07-21
MAISTRA-1400, MAISTRA-1685, MAISTRA-1348: Changes to IOR

jwendell:
This PR addresses the following changes to IOR:

- MAISTRA-1400: IOR now belongs to pilot, it is no longer a standalone deployment
- MAISTRA-1685: Move IOR chart configuration to its own level, it's no longer part of gateway config
- MAISTRA-1348: Enable IOR by default.


resources/helm/v2.0/istio-control/istio-discovery/templates/ior-roles.yaml
```diff
@@ -0,0 +1,35 @@
...
+    maistra-version: 2.0.0
+    app: istiod
+    release: {{ .Release.Name }}
+rules:
+- apiGroups: ["route.openshift.io"]
+  resources: ["routes", "routes/custom-host"]
+  verbs: ["get", "list", "watch", "create", "delete"]
+- apiGroups: [""]
+  resources: ["services", "pods"]
+  verbs: ["get", "list"]
```
rcernich: Can we just add these rules to existing clusterrole.yaml?

resources/smcp-templates/v2.0/base
```diff
@@ -95,4 +94,6 @@ spec:
           cpu: 10m
           memory: 128Mi
         limits:
+    ior:
```
rcernich: Does it make sense to move this under `pilot`?  Also, since we're moving it, we might see if we can come up with a more descriptive name.

Also, we need to add a field into smcp v2 to enable/disable this feature.

resources/helm/v2.0/istio-control/istio-discovery/templates/ior-roles.yaml
jwendell: @rcernich  These are just roles, not cluster roles. Also, having them in a separate files means one less `sed` hacking.

resources/smcp-templates/v2.0/base
jwendell: @rcernich Have we decided on a name yet? No objections in moving it under `pilot`.
Where is this smcp v2 stuff?

resources/helm/v2.0/istio-control/istio-discovery/templates/ior-roles.yaml
rcernich: There's a difference between a role and a role binding.  We use cluster roles because they can be referenced by role bindings in other namespaces.  When a namespace is configured to be a part of the mesh, we create role bindings for istio (galley, pilot, etc.) which reference the cluster role.  If we used namespaced roles, we'd have to create the roles in each namespace as well.

As far as being separate, it just means we need to create an extra role binding too.

jwendell: @rcernich done.

resources/smcp-templates/v2.0/base
jwendell: @rcernich done. Called it `routeManager`

build/patch-charts.sh
```diff
@@ -146,9 +146,14 @@ function patchGalley() {
...
-  echo '  - apiGroups: ["maistra.io"]
+  echo '
+  # Maistra specific
+  - apiGroups: ["maistra.io"]
     resources: ["servicemeshmemberrolls"]
-    verbs: ["get", "list", "watch"]' ${HELM_DIR}/istio-control/istio-discovery/templates/clusterrole.yaml
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["route.openshift.io"]
+    resources: ["routes", "routes/custom-host"]
+    verbs: ["get", "list", "watch", "create", "delete"]' >> ${HELM_DIR}/istio-control/istio-discovery/templates/clusterrole.yaml
```
rcernich: do you need `update` too?

build/patch-charts.sh
```diff
@@ -179,6 +184,15 @@ base:
...
 \      istio.io/rev: {{ .Values.revision | default "default" }}' \
     $deployment
+
+  # IOR
+  sed_wrap -i -e '/env:/ a\
+{{- $iorEnabled := "true" }}\
+{{- if and .Values.pilot.routeManager (eq .Values.pilot.routeManager.enabled "false") }}\
+{{- $iorEnabled := "false" }}\
+{{- end }}\
+          - name: ENABLE_IOR\
```
rcernich, line 195: i'm not sure what convention upstream uses for things like this, but do we want env, arg, or mesh config (obviously, we can't use mesh config)?  also, for env variable names, should this be prefixed with `ISTIO_META_`?

rcernich (changes requested):
You need to incorporate the new setting into the SMCP v2 structure, as well as add the proper conversion logic for moving between v1 and v2.  I don't know where the best place is for this new setting.  It could go in the cluster section or the gateways section.  I think gateways probably makes more sense, as the setting modifies the behavior of gateways (at least how their managed by the control plane).

Also, please update the v2 example files.

build/patch-charts.sh
dgn: the `ISTIO_META_` prefix is for proxies; they will advertise all env vars with this prefix to pilot during initial connection

build/patch-charts.sh
```diff
@@ -179,6 +184,15 @@ base:
...
+            value: "{{ $iorEnabled }}"' "${deployment}"
```
dgn, line 196: Couldn't you put the condition right in here?
`value: "{{ and .Values.pilot.routeManager (eq .Values.pilot.routeManager.enabled "false") }}"`

build/patch-charts.sh
jwendell: @rcernich Strictly speaking, no. IOR doesn't update a route, it deletes and creates a new one. That said, I think it makes sense to add `update` too, for completeness.

build/patch-charts.sh
jwendell: In pilot/istiod, this is called something like feature gate. We enable/disable certain features by making use of env variables. This is the less intrusive approach.

build/patch-charts.sh
jwendell: @dgn That value should be `true` by default, for example, if that key doesn't exist.

rcernich (approved):

rcernich (approved):

brian-avery (approved):

PR #492, opened 2020-07-24
MAISTRA-1631 implement SMCP v1 to v2 conversion

rcernich:


rcernich:
These changes are now complete.  There were some minor modifications to some of the structs to make things more consistent.  I haven't tested an install yet, but will give it a go in the morning.

PR #503, opened 2020-08-05
MAISTRA-1720 Fix CRD generation for SMCPv2

dgn:
Apparently Kubernetes 1.18 introduces CRD validation that fails on core types :face_with_head_bandage: 

This is the quick fix to get things working on OCP 4.5+, the real fix would be to change either the core type or the CRD validation code in K8s

Note that I didn't re-generate the CRDs in this PR because it hasn't been done for a while and would introduce a lot of other changes. I did test this though


dgn:
`gen-check` fails until https://github.com/maistra/istio-operator/pull/504 is merged

rcernich (approved):

build/generate-manifests.sh
```diff
@@ -45,6 +45,10 @@ function checkDependencies() {
 
 function generateCRDs() {
   go run -mod=vendor github.com/operator-framework/operator-sdk/cmd/operator-sdk generate crds --crd-version v1
+
+  # workaround for https://github.com/kubernetes-sigs/controller-tools/issues/457
```
jwendell, line 49: Should we create an issue to track this?

jwendell (approved):

build/generate-manifests.sh
dgn: Good point. I filed https://issues.redhat.com/browse/MAISTRA-1721

PR #505, opened 2020-08-05
MAISTRA-1633 fix issues in 2.0 charts

rcernich:

These fix issues in the 2.0 charts.  I'm seeing errors in istiod that I need to track down (looks like it's using global watch instead of SMMR), but it is being deployed.


jwendell (approved):

dgn (approved):
Thanks, these changes make sense. The `.Values.global.istioNamespace` variable is an artifact of Costin's istio/installer adventure and not needed for our purposes anyway; I'd love to remove it in our 1.0/1.1 charts, too - that would solve some problems with gateways. Not sure we'd make changes like that in older releases though

dgn:
Oh, the unit test actually will need to be fixed; it's counting the CRDs, and you removed one

resources/helm/overlays/istio-control/istio-discovery/templates/meshclusterrole.yaml
```diff
@@ -0,0 +1,17 @@
+# Dedicated cluster role - istiod will use fewer dangerous permissions ( secret access in particular ).
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: istiod-internal-{{ .Values.revision | default "default" }}-{{ .Release.Namespace }}
```
dgn, line 5: why add this?

dgn: I'm asking because there's already a ClusterRole `istiod-default-{namespace}`

resources/helm/overlays/istio-control/istio-discovery/templates/meshclusterrole.yaml
```diff
@@ -0,0 +1,17 @@
...
+    release: {{ .Release.Name }}
+rules:
+  # auto-detect installed CRD definitions
+  - apiGroups: ["apiextensions.k8s.io"]
+    resources: ["customresourcedefinitions"]
+    verbs: ["get", "list", "watch"]
+  # for k8s service account token authentication
+  - apiGroups: ["authentication.k8s.io"]
+    resources: ["tokenreviews"]
+    verbs: ["create"]
```
dgn, line 17: I was able to get istiod to start when I added the following permissions here:
```
  # auto-detect installed CRD definitions
  - apiGroups: ["apiextensions.k8s.io"]
    resources: ["customresourcedefinitions"]
    verbs: ["get", "list", "watch"]
  # for k8s service account token authentication
  - apiGroups: ["authentication.k8s.io"]
    resources: ["tokenreviews"]
    verbs: ["create"]
  - apiGroups: [""]
    resources: ["namespaces", "nodes"]
    verbs: ["get", "list", "watch"]
  - apiGroups: ["networking.k8s.io"]
    resources: ["ingresses", "ingressclasses"]
    verbs: ["get", "list", "watch"]
  - apiGroups: ["networking.k8s.io"]
    resources: ["ingresses/status"]
    verbs: ["*"]
```

Of course we don't want to give it those permissions, but we can solve that in follow-ups

resources/helm/overlays/istio-control/istio-discovery/templates/meshclusterrole.yaml
rcernich: The revision stuff is how multiple control planes are supported, so this ensures a unique name.  Same comment applies to every place else where the name went from <name>-<namespace> to <name>-<revision>-<namespace>.  (Writing this, I'm wondering if there are other places where revision needs to be added, e.g. mixer, prometheus, kiali, etc.)

PR #515, opened 2020-08-12
MAISTRA-1612 Remove injectPodRedirectAnnot option and always inject them

dgn:


luksa (approved):

build/patch-charts.sh
```diff
@@ -252,12 +252,6 @@ function patchSidecarInjector() {
   sed_wrap -i -e 's/runAsGroup: 1337/runAsGroup: {{ .ProxyUID }}/g' ${HELM_DIR}/istio-control/istio-discovery/files/injection-template.yaml
   sed_wrap -i -e 's/fsGroup: 1337/fsGroup: {{ .ProxyUID }}/g' ${HELM_DIR}/istio-control/istio-discovery/files/injection-template.yaml
 
-  # add annotation for Multus & Istio CNI
-  sed_wrap -i -e 's/^\(.*injectedAnnotations:.*\)$/\1\
-    \{\{- if and (.Values.istio_cni.enabled) (not .Values.sidecarInjectorWebhook.injectPodRedirectAnnot) \}\}\
-      k8s.v1.cni.cncf.io\/networks: \{\{.Values.istio_cni.istio_cni_network\}\}\
-    \{\{- end \}\}/' ${HELM_DIR}/istio-control/istio-discovery/templates/istiod-injector-configmap.yaml
-
   sed_wrap -i -e '/- name: istio-proxy/,/resources:/ {
```
knrc, line 255: Why is this being removed?

knrc (changes requested):

build/patch-charts.sh
knrc: Sorry, was mixing this up with the other k8s.v1.cni.cncf.io/networks changes.  Please ignore

knrc (approved):

build/patch-charts.sh
rcernich: Isn't the purpose of this PR to force `injectPodRedirectAnnot` to true, i.e. always inject pod annotations?

knrc: This change covers the case when it was set to false, we had to ensure the `k8s.v1.cni.cncf.io/networks` annotation was set correctly.  With it set to true the annotation is covered elsewhere so this can safely be removed.

rcernich: I don't think it's configurable anymore: https://github.com/maistra/istio/pull/165

rcernich: actually, shouldn't we just change the command line flag to true?

knrc: +1, the configuration option was added by us during 1.1 as this was a feature introduced part way through the stream.  The other PR is aligning with the upstream codebase in that it's not a configurable option, it's always enabled.

PR #519, opened 2020-08-19
MAISTRA-1682 add support for v2 profiles

rcernich:
Also includes `MAISTRA-1575 refine SMCP v2 structure` which includes changes to make working with the new structure easier (based on updating the profiles to v2).


rcernich:
I needed to regenerate the CRDs, so I rolled them back to v1beta1 in this PR.

rcernich:
I've verified that `oc apply` works with the trimmed down SMCP CRD.

pkg/apis/maistra/v2/smcp_new.yaml
```diff
@@ -296,20 +279,29 @@ spec:
...
-          discovery:
-            resources:
-              limits: {}
-              requirements: {}
-    citadel: {} # similar to pilot above
-    galley: {} # similar to pilot above
-    # note, mixer runtime is configured as part of policy.mixer or telemetry.mixer
+      pilot: # same as gateway.runtime structure illustrated above
+        deployment:
+          replicas: 2
```
brian-avery, line 286: is this better to have in the yaml files?

pkg/apis/maistra/v2/smcp_new.yaml
```diff
@@ -344,34 +336,27 @@ spec:
...
-                nodeCount: 3
-                storage: {} # raw jaeger config, .Values.tracing.jaeger.elasticsearch.storage
-                redundancyPolicy: {} # raw jaeger config, .Values.tracing.jaeger.elasticsearch.redundancyPolicy
-                indexCleaner: {} # raw jaeger config, .Values.tracing.jaeger.elasticsearch.esIndexCleaner
-                runtime: {} # .Values.tracing.jaeger.elasticsearch.<runtime-specific-settings>
+          storage:
+            type: Memory # or Elasticsearch
+            memory: # implies tracing.jaeger.template=all-in-one
+              maxTraces: 100000
+            # or
```
brian-avery, line 349: Are this and node count better to have in the yaml files?

pkg/controller/versions/template.go
```diff
@@ -69,58 +70,89 @@ func mergeValues(base map[string]interface{}, input map[string]interface{}) map[
...
+			return nil, nil, fmt.Errorf("template cannot be loaded from user or default directory. Error from user: %s. Error from default: %s", err, defaultErr)
 		}
-		templateContent = defaultTemplateContent
+		profileContent = defaultProfileContent
 	}
 
-	var template v1.ServiceMeshControlPlane
-	if err = yaml.Unmarshal(templateContent, &template); err != nil {
-		return v1.ControlPlaneSpec{}, fmt.Errorf("failed to parse template %s contents: %s", name, err)
+	obj, gvk, err := decoder.Decode(profileContent, nil, nil)
```
brian-avery, line 90: Is there any advantage to using this over the ghodds? Ghodss is simpler, and doesn't require the additional weight of using reflection. 

Looking at https://github.com/kubernetes/kubernetes/pull/74378, it looks like Kubernetes forked ghodss, but it's unclear why.

I'd be curious if making this change slows us down/what the cost/benefit of the additional complexity is.

brian-avery: It looks like they may have switched to a fork to try to support YAML 1.2, which it looks like they may be falling back on. YAML 1.2 introduces breaking changes in handling of booleans, and some other changes. Not sure how that would affect those writing templates. https://github.com/kubernetes/kubernetes/issues/34146

pkg/controller/versions/template.go
```diff
@@ -69,58 +70,89 @@ func mergeValues(base map[string]interface{}, input map[string]interface{}) map[
...
-func (v version) recursivelyApplyTemplates(ctx context.Context, smcp v1.ControlPlaneSpec, visited sets.String) (v1.ControlPlaneSpec, error) {
+func (v version) recursivelyApplyProfiles(ctx context.Context, smcp *v1.ControlPlaneSpec, profiles []string, visited sets.String) (v1.ControlPlaneSpec, error) {
 	log := common.LogFromContext(ctx)
-	if smcp.Template == "" {
-		return smcp, nil
-	}
-	log.Info(fmt.Sprintf("processing smcp template %s", smcp.Template))
 
-	if visited.Has(smcp.Template) {
-		return smcp, fmt.Errorf("SMCP templates form cyclic dependency. Cannot proceed")
```
brian-avery, line 105: I don't think this should be removed. It's really low hanging validation and the presence of circular dependencies means the user configured something improperly. Ignoring it it means that the user will not be achieving the behavior that they specified. We should find a way to surface this as a warning to the user before they encounter unrelated issues.

pkg/apis/maistra/v2/smcp_new.yaml
rcernich: this is just an example file embedded with the struct definition, as it's difficult to get an idea of the structure by looking at the go code (since it's all spread around between files).

also, i think we've decided that we want the profiles to be _the_ default, so we're not relying on values.yaml.  that said, i don't think we'll be there for a while.  also, it means the user will have to have a full config if they're not using the default profile, which means we may want to always apply a default.

brian-avery: the default profile should be used as the base for *everything* shouldn't it? If you don't specify a profile, use default. If you specify a profile, have it inherit at some level from default? I don't think that should be avoidable.

brian-avery: Thanks.

pkg/controller/versions/template.go
rcernich: No, because we don't know which version of SMCP is being loaded.  We need to pass an object to the marshaller and we don't know which one to pass (v1 or v2).

pkg/controller/versions/template.go
rcernich: Yeah, I guess I disagree.  They are getting the behavior they specified.  They just specified something that could break the operator (because we'd continually recurse).  The other problem is that because profiles is now a list, it's possible that a profile could be specified multiple times, without actually being circular.  All this does is break the chain, returning a result.  If the user doesn't get the result they wanted, then the configuration error is their problem, which isnot really any different than if they had specified bogus stuff in a single profile, that wasn't part of a circular chain.

I think the ultimate solution is to incorporate the profiles during the validation phase, so that we reject any invalid smcp before it gets to reconcile.  We should also add the resulting resource to the status, so users can inspect what they've got.  Adding a history would be nice too, but a better option might be to have some sort of tool that users can test with, or have some sort of dry-run capabilities.  Feel free to open a jira.

brian-avery: yeah. Adding the resulting resource to the status does allow our users to go back when they start seeing problems and see what the result after parsing is. And adding the SMCP templates to validation would also help.

pkg/controller/versions/template.go
brian-avery: Ah ok. Other option here would be: 

v2 := &v1.SMCP
if Unmarshal(v2) != error ... 

else unmarshal(v1) != error ...
else fail

Probably faster just to take the hit for reflection at that point though.

brian-avery (approved):

knrc (approved):

dgn (commented):
~I guess this was already merged; it doesn't work on 4.5 however. The apiserver does not accept the CRDs~

Nevermind `maistra-2.0` seems to work. Maybe my branch was outdated

rcernich:
Yeah, the default stuff is not generated/supported for v1beta1.  I left it commented out for when we move back to v1.  That said, I think @brian-avery, was having problems building the OLM index, so I think there's some additional cleanup required wrt the generated crds.  (I went from using operator-sdk to using controller-tools directly.  operator-sdk does not expose all the flags available for crd generation, specifically, MaxDescLen.)

PR #502, opened 2020-08-04
Remove ca_file option in Kiali operator CR

israel-hdez:
Kiali has moved away from the deprecated service serving interface. The
Kiali operator is providing good defaults for ca_file in OpenShift, so
it's no longer needed that maistra operator specifies these options.

PR #520, opened 2020-08-19
Remove ca_file option in Kiali operator CR

israel-hdez:
Same as #502 but for `maistra-1.1` branch.
For fixing: https://issues.redhat.com/browse/MAISTRA-1719

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ oc create -n observability -f https://raw.githubusercontent.com/jaegertracing/ja
 To install the Kiali operator, execute the following command:
 
 ```
-bash <(curl -L https://git.io/getLatestKialiOperator) --operator-image-version v1.0.0 --operator-watch-namespace '**' --accessible-namespaces '**' --operator-install-kiali false
+bash <(curl -L https://kiali.io/getLatestKialiOperator) --operator-watch-namespace '**' --accessible-namespaces '**' --operator-install-kiali false
 ```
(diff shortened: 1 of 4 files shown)


israel-hdez:
cc @rcernich @dgn

rcernich:
@israel-hdez, you'll need to create a separate PR for the 1.0 changes, which will apply to the maistra-1.0 branch.

brian-avery (approved):

israel-hdez:
Should I create the PR for 1.0 branch, or is this going to be cherry-picked automatically?

rcernich:
I don't think we can cherry-pick, as the file structure differs, e.g. there is no helm/v1.1 directory in maistra-1.0.

JFYI, the idea here was that we'd be maintaining the charts specific to each istio release in its corresponding maistra branch, e.g. istio 1.1 charts are maintained in maistra-1.0, istio 1.4 charts in maistra-1.1, and istio 1.6 charts in maistra-2.0.  The rationale was that if we maintained them in a single branch, we'd need to maintain multiple patching scripts, etc.  In hindsight, though, I don't know if what we're doing today is any better than just maintaining all of them in a single branch.

israel-hdez:
Here is the PR for 1.0 branch: https://github.com/maistra/istio-operator/pull/529

rcernich (approved):

PR #521, opened 2020-08-25
MAISTRA-1780 Use v4.6.0 of json-patch to speed up generation of patch

luksa:


go.mod
```diff
@@ -51,3 +51,6 @@ replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+inc
 replace k8s.io/cli-runtime => k8s.io/cli-runtime v0.18.3
 
 replace k8s.io/client-go => k8s.io/client-go v0.18.3
+
+// use json-patch 4.6.0, since earlier versions cause json patch generation to be very slow (MAISTRA-1780)
+replace github.com/evanphx/json-patch => github.com/evanphx/json-patch v0.0.0-20191220184958-bf22ed931162
```
rcernich: Should we use a specific tag?

luksa: Version 4.6.0 contains a {{go.mod}} file and can't be referenced by tag. But I've now replaced it with v4.9.0, which removes the {{go.mod}} file specifically to solve this problem.

dgn (approved):

rcernich (approved):

PR #523, opened 2020-08-25
MAISTRA-1781 support v2 resources in mutating/validating webhooks

rcernich:


pkg/controller/versions/strategy_v1_0.go
```diff
@@ -79,12 +91,24 @@ func (v *versionStrategyV1_0) Validate(ctx context.Context, cl client.Client, sm
 	return utilerrors.NewAggregate(allErrors)
 }
 
-func (v *versionStrategyV1_0) ValidateDowngrade(ctx context.Context, cl client.Client, smcp *v1.ServiceMeshControlPlane) error {
+func (v *versionStrategyV1_0) ValidateV2(ctx context.Context, cl client.Client, smcp *v2.ServiceMeshControlPlane) error {
```
dgn, line 94: I'm starting to think that maybe the v2 resource should not be used to deploy v1 control planes - in the end, we dropped a lot of options that were specific to the old way of deploying Istio. Not sure how to enforce that though, with Kubernetes only accepting one resource version to be the default at all times

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -35,52 +39,50 @@ var _ admission.DecoderInjector = (*ControlPlaneMutator)(nil)
...
+			}
 		}
 	}
 
-	if smcp.Spec.Template == "" {
-		log.Info("Setting .spec.template to default value", "template", maistrav1.DefaultTemplate)
-		newSmcp.Spec.Template = maistrav1.DefaultTemplate
-		smcpMutated = true
+	if len(mutator.GetProfiles()) == 0 {
+		log.Info("Setting .spec.profiles to default value", "profiles")
```
luksa: Missing argument. The function call panics with the following message: 
```
odd number of arguments passed as key-value pairs for logging   {"ServiceMeshControlPlane": "istio-system/minimal-install", "ignored key": "profiles"}
```

pkg/controller/servicemesh/webhooks/mutation/controlplane_test.go
```diff
@@ -26,8 +27,29 @@ func TestDeletedControlPlaneIsAlwaysAllowed(t *testing.T) {
...
+func TestDeletedV2ControlPlaneIsAlwaysAllowed(t *testing.T) {
+	controlPlane := newControlPlaneV2("my-smcp", "istio-system")
+	controlPlane.Spec.Version = ""
+	controlPlane.Spec.Template = ""
+	controlPlane.DeletionTimestamp = now()
+
+	mutator, _, _ := createControlPlaneMutatorTestFixture()
+	response := mutator.Handle(ctx, newCreateRequest(controlPlane))
+	assert.DeepEquals(response, acceptWithNoMutation, "Expected mutator to accept deleted ServiceMeshControlPlane", t)
+}
```
luksa: Can this and `TestDeletedV1ControlPlaneIsAlwaysAllowed()` be turned into a single test with two test cases?

pkg/controller/servicemesh/webhooks/mutation/controlplane_test.go
```diff
@@ -26,8 +27,29 @@ func TestDeletedControlPlaneIsAlwaysAllowed(t *testing.T) {
...
+	controlPlane := newControlPlaneV1("my-smcp", "not-watched")
+	controlPlane.Spec.Version = ""
+	controlPlane.Spec.Template = ""
+	mutator, _, _ := createControlPlaneMutatorTestFixture()
+	mutator.namespaceFilter = "watched-namespace"
+	response := mutator.Handle(ctx, newCreateRequest(controlPlane))
+	assert.DeepEquals(response, acceptWithNoMutation, "Expected mutator to accept ServiceMeshControlPlane whose namespace isn't watched", t)
+}
+
+func TestV2ControlPlaneOutsideWatchedNamespaceIsAlwaysAllowed(t *testing.T) {
```
luksa: Same here - single test function with two test cases.

pkg/controller/versions/strategy_v1_0.go
rcernich: It's something to consider, although it would be ideal if we could ensure as many settings as possible are mapped.  Part of the reason for having v2 is to declare a known set of supported options.  It's certainly possible that some fringe settings could be missed in the new schema.

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
rcernich: fixed

pkg/controller/servicemesh/webhooks/mutation/controlplane_test.go
rcernich: done.  moved all v1/v2 specific test functions into single test function with multiple cases.

rcernich:
@dgn, @luksa, are you okay with the last round of changes?

luksa (approved):

rcernich:
@dgn, Could you give this a final look please?

dgn (approved):
lgtm

PR #527, opened 2020-08-31
MAISTRA-1776 fix issue with null params in handler resources for v1.x…

rcernich:
… installs


dgn (approved):

luksa:
In this case, the OpenAPI schema seems to be wrong. `params:` and `params: {}` should be equivalent, so I think it would be better to fix the schema. But otherwise, this lgtm.

luksa (approved):

PR #530, opened 2020-09-02
Disable Prometheus in minimal example

jwendell:


dgn (approved):

luksa:
PR needs to be updated. The schema has been modified, moving the prometheus config from `.spec.addons.metrics.prometheus` to `.spec.addons.prometheus`.

But LGTM otherwise.

deploy/examples/maistra_v2_servicemeshcontrolplane_cr_minimal.yaml
```diff
@@ -11,6 +11,9 @@ spec:
       jaeger:
         name: jaeger
         install: {}
+    metrics:
```
brian-avery: Should the same be present in maistra.io?

brian-avery (approved):

rcernich:
If we're disabling prometheus, should we also disable telemetry?  Also, was prometheus disabled in the old minimal example?  Is there a JIRA for this work?

PR #531, opened 2020-09-03
Backport changes to make daily build work

brian-avery:
Backport changes from daily build to operator.

### DIFF
--- a/build/generate-manifests.sh
+++ b/build/generate-manifests.sh
@@ -56,6 +56,7 @@ function generateCRDs() {
   mv deploy/crds/maistra.io_servicemeshcontrolplanes.yaml ${BUNDLE_DIR}/servicemeshcontrolplanes.crd.yaml
   mv deploy/crds/maistra.io_servicemeshmemberrolls.yaml ${BUNDLE_DIR}/servicemeshmemberrolls.crd.yaml
   mv deploy/crds/maistra.io_servicemeshmembers.yaml ${BUNDLE_DIR}/servicemeshmembers.crd.yaml
+
   rm -rf deploy/crds/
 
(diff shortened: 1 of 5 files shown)


build/generate-manifests.sh
```diff
@@ -56,13 +56,20 @@ function generateCRDs() {
...
 
   cat ${BUNDLE_DIR}/servicemeshcontrolplanes.crd.yaml >deploy/src/crd.yaml
   echo -e "\n---\n" >>deploy/src/crd.yaml
   cat ${BUNDLE_DIR}/servicemeshmemberrolls.crd.yaml >>deploy/src/crd.yaml
   echo -e "\n---\n" >>deploy/src/crd.yaml
   cat ${BUNDLE_DIR}/servicemeshmembers.crd.yaml >>deploy/src/crd.yaml
+
+  currentDir=$(pwd)
+  cd ${BUNDLE_DIR}
+  sed -i -e '/x-kubernetes-list-map-keys:/,/x-kubernetes-list-type: map/ { /- protocol/d }' *.crd.yaml
```
rcernich, line 70: Can you move this to the original fix, which was on line 53?

brian-avery: Sure. I thought that was a separate fix since this doesn't modify default values or descriptions.

brian-avery:
Aargh. Problem working with the same branch name in my remote. @jwendell @rcernich Should I revert or would you just like to review as is? I removed the code from line 53 that removes defaults.

build/generate-manifests.sh
rcernich: It fixes the same problem, but it only works on 4.5+.  Having the comment too is a hint about what's going on so we can remove it once we're no longer supporting 4.4.

rcernich:
Just create a new pr.

PR #532, opened 2020-09-03
Move sed a few lines higher

brian-avery:
Closed the previous PR without addressing the request to move the sed. This PR accomplishes that.


build/generate-manifests.sh
```diff
@@ -44,11 +44,15 @@ function checkDependencies() {
...
 function generateCRDs() {
-  #go run -mod=vendor github.com/operator-framework/operator-sdk/cmd/operator-sdk generate crds --crd-version v1
   go run -mod=vendor sigs.k8s.io/controller-tools/cmd/controller-gen crd \
       paths=./pkg/apis/maistra/... \
       crd:maxDescLen=0,preserveUnknownFields=false,crdVersions=v1beta1 \
       output:crd:dir=./deploy/crds
+   # workaround for https://github.com/kubernetes-sigs/controller-tools/issues/457
+  #sed -i -e "s/\( *\)\(description\: The IP protocol for this port\)/\1default: TCP\n\1\2/" \
+  #    deploy/crds/maistra.io_servicemeshcontrolplanes.yaml
+  sed -i -e '/x-kubernetes-list-map-keys:/,/x-kubernetes-list-type: map/ { /- protocol/d }' deploy/crds/*
```
rcernich: I really think this should be constrained to just the SMCP crd.

build/generate-manifests.sh
```diff
@@ -44,11 +44,15 @@ function checkDependencies() {
...
+  sed -i -e '/---/d' deploy/crds/*
```
rcernich, line 57: Maybe add a comment about why we're doing this.  It's not obvious, because it shouldn't be a problem applying the resource and there's nothing to tie it back to problems with OLM metadata builds.

build/generate-manifests.sh
brian-avery: Sure. I can constrain this one to just the SMCP crd. Are you ok with the next one on all CRDs? All of them are currently broken.

config/maistra.io_servicemeshmemberrolls.yaml
```diff
@@ -0,0 +1,148 @@
+
+---
```
brian-avery: @rcernich what makes these?

rcernich: run `make generate-manifests`

brian-avery: Ran `make generate-manifests`. All of the new files are still there though. I'm assuming that's good.

rcernich (approved):

jwendell (approved):

PR #533, opened 2020-09-03
MAISTRA-1802 Add support for exiting Jaeger and Kiali resources

rcernich:


rcernich:
Moving back to wip, as it seems kiali cr is getting deleted during prune operators after modifying smcp resource.

rcernich:
Everything appears to be working as it should now.

PR #534, opened 2020-09-04
MAISTRA-1800 Use service-ca-operator to inject CA bundle

luksa:


pkg/controller/servicemesh/webhookca/controller.go
```diff
@@ -210,18 +237,30 @@ func (wm *webhookCABundleManager) UpdateCABundle(ctx context.Context, cl client.
...
+		logger.Info("could not get CA bundle", "caBundleSourceConfig", caBundleSourceConfig, "error", err)
 		return nil
 	}
-	return currentConfig.UpdateCABundle(ctx, cl, caRoot)
+	return currentConfig.UpdateCABundle(ctx, cl, caBundle)
+}
+
+// GetCABundleFromSource retrieves the CA bundle from a secret or config map
+func (wm *webhookCABundleManager) getCABundleFromSource(ctx context.Context, cl client.Client, config caBundleSourceConfig) ([]byte, error) {
+	switch config.Kind {
```
rcernich, line 256: Would it make sense to create an interface for `CABundleSource` and put a `GetCABundle()` method on them?

rcernich (approved):
lgtm

pkg/controller/servicemesh/webhookca/controller.go
luksa: I don't like to introduce additional layers of abstraction until there's a clear benefit. Since this here is the only place where we have different behavior for the two source kinds, I don't think it's enough to warrant a new layer. Introducing an interface would just move the switch statement to the place we instantiate the struct.

rcernich:
rebased and replaced by #543

PR #547, opened 2020-09-22
MAISTRA-1799 Reconciler now checks if Prometheus is enabled when Kiali is enabled

luksa:


rcernich:
I think we need to process the profiles in the validator, as I think the checks will fail if users are applying an empty SMCP resource that uses the default profile.

luksa:
@rcernich I've now reimplemented this so that the validation is performed in the reconciler. If the validation fails, the SMCP's status will now show `ValidationError` instead of `ReconcileError`.

dgn (approved):

pkg/controller/versions/strategy_v2_0.go
```diff
@@ -404,6 +412,12 @@ func (v *versionStrategyV2_0) Render(ctx context.Context, cr *common.ControllerR
 	return renderings, nil
 }
 
+func validateAppliedSpec(appliedSpec *v2.ControlPlaneSpec) error {
```
rcernich: Would it not be better to try to run this through the regular validation process, i.e. use the same validation logic that's used by the webhook?

luksa: It can't be the same, as that would mean that the webhook would reject the SMCP in cases where the SMCP becomes valid only after applying the profiles. But I guess we could call the webhook validation logic here and add the additional checks that require the profiles to be applied beforehand. I'll do this.

luksa: OK, I've now changed this so that full validation is performed just before the charts are rendered.

rcernich (approved):

dgn (approved):
lgtm. should we look at doing the template merges as part of the validating webhook, for post 2.0?

luksa:
> lgtm. should we look at doing the template merges as part of the validating webhook, for post 2.0?

That's the plan. But we may introduce a resource that will hold the merged SMCP spec by then, so we may just end up performing all the validation in the webhook for this resource.

PR #550, opened 2020-09-22
MAISTRA-1833 Don't check CNI readiness during SMCP installation

luksa:


pkg/controller/servicemesh/controlplane/integration_test.go
```diff
@@ -141,10 +141,8 @@ func TestBootstrapping(t *testing.T) {
 							Verify("create").On("customresourcedefinitions").IsSeen(),
 							// verify that CNI is installed
 							Verify("create").On("daemonsets").Named(cniDaemonSetName).In(operatorNamespace).IsSeen(),
-							// verify CNI readiness check during reconcile
-							Verify("list").On("daemonsets").In(operatorNamespace).IsSeen(),
```
rcernich, line 145: I think we should leave this, as it verifies that the install tries to check readiness of cni.

luksa: No, this is the exact readiness check that this PR removes in reconciler.go. The other one in {{VerifyReadinessCheckOccurs()}} was removed in my other PR (it was invoked during the normal post-install readiness check).

rcernich (approved):

PR #552, opened 2020-09-22
MAISTRA-1825 fix prometheus integration; MAISTRA-1804 specify istio c…

rcernich:
…onfig map name for kiali


dgn:
I think I'm lacking context here. EnvoyFilter resources in the control plane namespace apply to all mesh namespaces- so why do we copy them around?

pkg/controller/servicemesh/memberroll/namespace_reconciler.go
```diff
@@ -326,6 +339,65 @@ func (r *namespaceReconciler) reconcileRoleBindings(ctx context.Context, namespa
...
+	labelSelector := map[string]string{common.MemberOfKey: r.meshNamespace}
+	err := r.Client.List(ctx, &namespaceEnvoyFilters, client.MatchingLabels(labelSelector), client.InNamespace(namespace))
+	if err != nil {
+		reqLogger.Error(err, "error retrieving EnvoyFilter resources for namespace")
+		return err
+	}
+
+	allErrors := []error{}
+
+	// add required role bindings
```
dgn, line 355: nit: envoy filters

rcernich:
Yep.  I need to rework this.  The problem is that the version is not being set correctly in the proxy, so the filters aren't being applied.

rcernich:
Superseded by https://github.com/maistra/istio-operator/pull/555

PR #556, opened 2020-09-23
MAISTRA-1844: Update operator to support maistra 1.1.9

knrc:


rcernich (approved):

knrc:
This should be good to go once #557 is merged

resources/helm/v1.1/istio/charts/kiali/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 description: Kiali is an open source project for service mesh observability, refer to https://www.kiali.io for details.
 name: kiali
-version: 1.9.1
```
dgn: was this change intentional? it doesn't look like anything but the version number changed here

knrc: This change was pulled in by the Makefile, let me check what is going on

knrc: cleaning everything out and rerunning the Makefile fixes this, I'll push up a new version now

dgn (approved):

PR #558, opened 2020-09-24
MAISTRA-1831 use objectSelector in injector webhook

rcernich:

### DIFF
--- a/resources/smcp-templates/v2.0/base
+++ b/resources/smcp-templates/v2.0/base
@@ -24,7 +24,8 @@ spec:
     level: warn
 
   proxy:
-    autoInject: true
+    injection:
+      autoInject: false
     networking:
       protocol:
         autoDetect:
@@ -88,3 +89,6 @@ spec:
   techPreview:
     wasmExtensions:
       enabled: false
+    sidecarInjectorWebhook:
+      objectSelector:
+        enabled: false
--- a/build/patch-charts.sh
+++ b/build/patch-charts.sh
@@ -220,7 +220,7 @@ function patchSidecarInjector() {
       -e '/metadata:/a\
 \  name: istiod-{{ .Values.revision | default "default" }}-{{ .Release.Namespace }}' \
     $webhookconfig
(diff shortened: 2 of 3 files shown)


rcernich:
I haven't tested this yet.  I left the logic for `istio.io/rev` always enabled to allow users to control which injector they want to use for the app.  This allows fine grained control over the upgrade process.  That said, I think we need to put a little more thought into how we will manage revisions going forward.  (I could see using `ServiceMeshMember` to select which control plane (revision) for the namespace and then put the correct `istio.io/rev` label on the namespace as part of SMMR reconciliation.)

dgn (approved):
Thanks, I think this makes sense.

Regarding the revision name, I think in the future we'll want to make it <smcp.name>-<smcp.generation> so we can actually perform upgrades of the same SMCP using revisions. wdyt?

rcernich:
I was thinking over night that requiring the revision might be a little burdensome for applications, as they'd need to know the name of the SMCP resource they want to use.  Furthermore, this is a break from 1.1, where the only requirement was adding the inject label to their pod.  Given that, I think it's best to remove the revision requirement, possibly adding it to the namespace selector in the future.

Regarding the name-generation, I'm not sure we'd want to do that as I think this could cause pods that were injected to not be injected after an update occurs, because it's referencing the older generation, e.g. if the pod were rescheduled because of cluster maintenance, the new pod would not get injected.

resources/smcp-templates/v2.0/base
```diff
@@ -24,7 +24,7 @@ spec:
     level: warn
 
   proxy:
-    autoInject: true
+    autoInject: disabled
```
dgn: why `disabled`? I think this has to be `false`, otherwise it will cause the `if` condition to be `true`

rcernich: doh! mixing contexts: new smcp and old values.yaml.  that saved an extra build/test cycle.  thanks!

dgn (approved):
just needs a rebase now

rcernich:
And I need to test it too. ;)

PR #560, opened 2020-09-24
MAISTRA-1851 STATUS column should only show readiness status if the SMCP has been reconciled successfully

luksa:


rcernich:
This looks okay to me, but do we also want a state that works for readiness? Or is that covered by the new work for deployment status?  I mention this because it's possible that a pod dies and the capabilities of the control plane would be degraded.

brian-avery:
I think having the readiness information available as well is likely useful. I could imagine that users would potentially be more concerned with whether their environment is ready than whether the operator has read their SMCP.

luksa:
> I think having the readiness information available as well is likely useful. I could imagine that users would potentially be more concerned with whether their environment is ready than whether the operator has read their SMCP.

The readiness information is in the {{READY}} column. This is similar to the output of {{oc get pods}}. The pod's status is shown as {{Running}} even if one of the containers is not ready. Of course, if the container actually crashes, the pod status column does  indicate that. So maybe we need to do the same here. 

As a user, ideally I want {{oc get smcp}} to answer the following questions (listed in order of importance):
- Is the SMCP reconciled fully (i.e. are all deployments created and up-to-date)?
- Is each component of the control plane running? (If a component runs multiple replicas, one is enough for the component to be considered ok. Also, if a component is running, but is not (yet) ready, it should also be considered as being fine).
- Are all the control plane components ready?

If the SMCP isn't reconciled, I don't really care about whether all its components are running or ready. 

The way it works right now is that the status is either {{ComponentsReady}} or {{ComponentsNotReady}}. There's no indication that reconciliation itself has failed. The changes in this PR ensure that the reconciliation status is shown, while the readiness is shown in the READY column. What's missing is the answer to "Are all control plane components actually running?". 

So, yes, maybe we need to change how we set the reason in the {{Ready}} condition. Maybe that's what we had initially, but our changes to the code that determines the readiness have somehow changed the reason.

luksa:
> This looks okay to me, but do we also want a state that works for readiness? Or is that covered by the new work for deployment status? I mention this because it's possible that a pod dies and the capabilities of the control plane would be degraded.

If a pod dies, the READY column would show that one of the components isn't ready. But I agree that also stating this in the STATUS column would be better. I just don't like to see actual readiness information in the STATUS column (e.g. pods are running fine, but just aren't ready yet).

rcernich:
I agree with everything you've said @luksa.  I wonder if for component status, we could have something like: readyCount/degradedCount/unavailable/totalCount, where "ready" is fully available, "degraded" is available, but not fully rolled out (e.g. 3/5 replicas running), "unavailable" means no replicas running (for components with multiple deployments, I think if any have a zero available count, then the whole component should be marked unavailable), and total is the total number of components.  Then users have a complete view of their control plane runtime status.

brian-avery:
Thanks @luksa I agree. Thinking about customers who, say, wait for all components to be up before beginning testing, or use oc get smcp to see if their mesh is ready. This sounds reasonable.

brian-avery (approved):

luksa:
I've now reimplemented this change so that the STATUS column still shows the readiness condition, but the Ready condition itself now has the same reason & message as the Reconciled condition if the SMCP hasn't been reconciled yet. Only then will the Ready condition reflect whether all components are ready or not.

This ensures that the STATUS column shows the relevant information.

brian-avery (approved):
Looks great. Thanks. Looks like gen check needs something, but approving.

rcernich (approved):

PR #562, opened 2020-09-25
MAISTRA-1844: Update operator to support maistra 1.1.9

knrc:


knrc:
Merging manually for release

PR #570, opened 2020-10-01
Web console fixes

luksa:


build/manifest-templates/clusterserviceversion.yaml
```diff
@@ -19,47 +19,31 @@ metadata:
...
-          "apiVersion": "maistra.io/v1",
+          "apiVersion": "maistra.io/v2",
           "kind": "ServiceMeshControlPlane",
           "metadata": {
-            "name": "basic-install",
+            "name": "basic",
             "namespace": "control-plane-namespace"
           },
           "spec": {
             "version": "v2.0",
```
brian-avery, line 29: This is the field that I was seeing the error for.

luksa: Yeah, but this change isn't what fixes it. The fact that we had a v1 SMCP in `alm-examples` meant that no field was actually initialized. This caused the name field to be set to "example", but there's a bug in the UI that makes it appear as though this is actually enter in the field, when it's actually not. So when you submit the form, there is no name set, hence the error.

Anyway, the change to v2 fixes it.

brian-avery: Thanks. Do we need the default in the profiles as well? If you install via CLI, you don't get any of this, correct?

deploy/maistra-operator.yaml
```diff
@@ -6975,7 +6980,7 @@ spec:
         olm.relatedImage.v1_1.proxyv2: docker.io/maistra/proxyv2-ubi8:1.1.8
         olm.relatedImage.v1_1.sidecar-injector: docker.io/maistra/sidecar-injector-ubi8:1.1.8
 
-        olm.relatedImage.v2_0.3scale-istio-adapter: quay.io/3scale/3scale-istio-adapter:v2.0.0
+        olm.relatedImage.v2_0.3scale-istio-adapter: quay.io/3scale/3scale-istio-adapter:v1.0.0
```
brian-avery: I think 3scale is publishing 2.0.0 for 2.0

brian-avery: and we don't usually test 3scale either. That's usually done by the 3scale team.

brian-avery (changes requested):
Should probably leave the 3scale version at 2.0.0. Not sure if we want to add the v2 default in this PR or in a separate one.

deploy/maistra-operator.yaml
luksa: This change came in when I ran `make generate-manifests`, so it was already in the codebase somewhere. I've now removed it from this PR.

dgn: I have the same change in my PR. please just leave it in- if we want to make it 2.0.0 we'll have to change the template that sets it to 1.0.0 in a separate PR. I will also make sure we're checking this in `gen-check` in the future

dgn (approved):

deploy/maistra-operator.yaml
brian-avery: Ok. The threescale guys made the change to upgrade it to 2.0. Perhaps they made it in the wrong location? Approving.

brian-avery (approved):

PR #571, opened 2020-10-01
MAISTRA-1878 Align ServiceMeshExtension CRD with maistra/istio

dgn:
Mostly, there was a rogue/outdated YAML in the manifest directory that broke the OLM bundle. While at it, I aligned the types.go with what we have in maistra/istio


jwendell (approved):

deploy/maistra-operator.yaml
```diff
@@ -6975,7 +6977,7 @@ spec:
         olm.relatedImage.v1_1.proxyv2: docker.io/maistra/proxyv2-ubi8:1.1.8
         olm.relatedImage.v1_1.sidecar-injector: docker.io/maistra/sidecar-injector-ubi8:1.1.8
 
-        olm.relatedImage.v2_0.3scale-istio-adapter: quay.io/3scale/3scale-istio-adapter:v2.0.0
+        olm.relatedImage.v2_0.3scale-istio-adapter: quay.io/3scale/3scale-istio-adapter:v1.0.0
```
brian-avery: Shouldn't this be 2.0? I think 3scale is shipping 2.0 with maistra 2.0.

deploy/servicemesh-operator.yaml
```diff
@@ -6975,7 +6977,7 @@ spec:
         olm.relatedImage.v1_1.proxyv2: registry.redhat.io/openshift-service-mesh/proxyv2-rhel8:1.1.8
         olm.relatedImage.v1_1.sidecar-injector: registry.redhat.io/openshift-service-mesh/sidecar-injector-rhel8:1.1.8
 
-        olm.relatedImage.v2_0.3scale-istio-adapter: registry.redhat.io/openshift-service-mesh/3scale-istio-adapter-rhel8:2.0.0
+        olm.relatedImage.v2_0.3scale-istio-adapter: registry.redhat.io/openshift-service-mesh/3scale-istio-adapter-rhel8:1.0.0
```
brian-avery: Same here

brian-avery (changes requested):
I think we should avoid changing the 3scale version. Other than that, looks good.

dgn:
> I think we should avoid changing the 3scale version. Other than that, looks good.

See https://github.com/maistra/istio-operator/pull/573

rcernich:
Do we want this installed as part of the operator, or should it be installed with the other Istio CRDs?

dgn:
> Do we want this installed as part of the operator, or should it be installed with the other Istio CRDs?

Good point. I was thinking to install it with the operator because it's managed by us directly; it is consumed by the control planes though. Might make sense to include it with the Istio CRDs. wdyt?

brian-avery (approved):

rcernich:
Yeah, I think we should move it into `istio-init/files`, although I don't thing there's any harm, other than tying the install up with OLM, which could cause issues in the future if we remove it from the operator metadata.

dgn:
I moved the CRD into the `istio-init` chart and split out CRD generation into its own script to accommodate that change. PTAL

brian-avery (approved):

rcernich (approved):

PR #573, opened 2020-10-02
MAISTRA-1885 Include manifest generation in `gen` target

dgn:


luksa (approved):

dgn:
Looks like we're using yqv2 in istio-operator but everywhere else we're using yqv3...

jwendell (approved):

rcernich:
I believe the 3scale versions differ between product and community, one with the `v` prefix and one without.

dgn:
> I believe the 3scale versions differ between product and community, one with the `v` prefix and one without.

Oh, you're right. Product doesn't use the `v` prefix. Will push a fix

PR #588, opened 2020-10-07
MAISTRA-1920 preserve unmapped fields from SMCP v1 resources

rcernich:

The only areas where fields may be lost is when the object is part of a list, which are gateway secret and config volumes (e.g. gateways.<name>.secretVolumes), label selectors for anti/affinity, and tolerations.

Any values remaining after conversion to v2 will be stuffed into the techPreview field.  This should provide lossless conversion between v1 and v2, i.e. v1->v2->v1 should equal the original object.


dgn:
> v1->v2->v1 should equal the original object

This part is not really achievable I'm afraid. I wrote a quick test that does it, and I can't get it to pass, namely because we're setting a bunch of defaults (like `global.meshExpansion.enabled=false`) that don't convert back cleanly

dgn:
I pushed my test, which is failing for the reasons I mentioned above. I really think we should try to make it pass, for two reasons:

1. I think it's going to avoid quite a few possible regressions regarding 1.1 deployments
2. We introduced yet another place where we apply defaults, which we shouldn't do IMO. I think we should put those new defaults into their values.yaml counterparts

If this turns out not to work, we can still skip the test

rcernich:
Regarding defaults, specifically for meshExpansion, the idea behind the new config is to allow as much as possible to be configured automatically.  This means the code that configures clustering does a lot of work based on a few settings, and expansion is not configured at all if clustering is not enabled.  For example, default expansion ports are automatically added to the cluster ingress gateway, as are networks, dns suffixes, etc. when multicluster is enabled.  That said, I was considering adding an enabled field to all of this, so at least the properties could be maintained, even if they weren't being used.  I didn't do that because I didn't think many folks would be using this, but...good catch.

rcernich:
I've added enablement flags to the multi-cluster and mesh expansion fields to allow them to be stubbed out.  This should preserve multi-cluster configuration from v1 resources.  However, if unspecified, multi-cluster and mesh expansion are disabled (i.e they need to be manually enabled).  Also, the current logic is that if multi-cluster is enabled, so is mesh expansion.

I modified the test to remove the results of fields fanning out from v2 to v1.  For example, enabling tracing in v1 causes the tracer to be set to jaeger too (default tracer if none is specified).  This may have some effect if default values differ, but we should be configuring things so they match the default.  For example, to enable tracing with v1, you would have to set all those fields, but some of those may have been the same defaults used in values.yaml, or may have been coded in the charts for correct behavior.  The v2 resource provides explicit support for things like this, so changing the tracing provider doesn't require changes to numerous other fields (although it may look a little verbose for cases that are defaulted sensibly in values.yaml).

rcernich:
This also appears to partially fix MAISTRA-1902, although only by way of preserving the field from v1.  For v2, it does not get converted into telemetry.mixer.

pkg/apis/maistra/conversion/conversion_test.go
```diff
@@ -23,6 +22,92 @@ var (
...
+					"multiCluster": map[string]interface{}{
+						"enabled": false,
+						"multiClusterOverrides": map[string]interface{}{
+							"expansionEnabled":    nil,
+							"multiClusterEnabled": nil,
+						},
+					},
+					// a result of enabling tracing, default provider is jaeger
+					"proxy": map[string]interface{}{
+						"tracer": string("jaeger"),
```
dgn: this should now be 'zipkin', right?

dgn (approved):
lgtm so far. I think we can do the fix of MAISTRA-1902 (fixing the v1->v2 conversion) in a separate PR

pkg/apis/maistra/conversion/conversion_test.go
rcernich: /test istio-operator_unittests

probably, after i rebase

rcernich: That is only used to remove those keys after conversion, so the checks pass, i.e. this documents what keys are added, but not values.

rcernich:
This fixes many issues with the conversion code as well as cleaning things up a bit.  The only changes to the structure of the SMCP spec are the addition of enabled fields to multiCluster, meshExpansion, and autoScaling (where previously they were enabled if not null).  I've also added more tests, specifically around profiles, and fixed bugs that I found there.  This should get merged as soon as possible.

rcernich:
@luksa, mind giving this a review?  Sorry it's so big.  I'm mostly concerned about the updates to the validation logic.  I added logic to ensure the smcp resource is configured consistently, as some mixer settings are shared between policy and telemetry in v1.x.

jwendell (approved):

PR #589, opened 2020-10-08
MAISTRA-1722: Remove cluster-scoped permissions from istiod

bison:
This removes the temporary cluster-scoped permissions from the
istiod-internal ClusterRole.  The ability to create TokenReviews is
all that's left here, which is not an issue.


dgn (approved):
Thanks, I quickly verified this works on my cluster

jwendell (approved):

rcernich (approved):

PR #591, opened 2020-10-09
MAISTRA-1816 Set status.configuredMembers to [] when SMCP is deleted

dgn:
When deleting an SMCP and re-creating it with a different version, the control plane requires different RoleBindings in the mesh's member namespaces. Keeping the status.configuredMembers across SMCP deletions can therefore cause problems that can be hard to recover from (only by deleting the SMMR manually), because the new control plane (with a different Istio version) will assume it has access to the mesh's member namespaces while it fact it doesn't, because the RoleBindings are not correct or point to ClusterRoles and Roles that no longer exist.


rcernich (approved):

luksa (approved):

dgn:
Damn. Turns out this doesn't actually work. I had tested a previous version that does. Will push a follow-up soon

PR #596, opened 2020-10-12
MAISTRA-1931 remove runAsUser/Group 1337 from mixer deployments

rcernich:

### DIFF
--- a/build/patch-charts.sh
+++ b/build/patch-charts.sh
@@ -441,12 +441,24 @@ function patchMixer() {
     /args/ a\
 \          - --memberRollName=default\
 \          - --memberRollNamespace=\{\{ .Release.Namespace \}\}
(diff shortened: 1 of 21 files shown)


jwendell (approved):

rcernich:
@dgn, @brian-avery, I copied the envoy config over from maistra/istio and updated based on what was in the envoy config for telemetry.  I'm not sure how correct it is, but it appears to work.  Anyway, with this PR and maistra/istio#192, mixer is now running in 2.0, with and without control plane security (envoy config is only used when control plane security is enabled, which is the default).

dgn (approved):

dgn:
There are some references to galley in that envoy config that we'll likely want to remove, but it's not important and we can do that in a follow-up

rcernich:
Yeah, those were in the telemetry config and I wasn't sure if they were needed or not, so I added them in.

PR #597, opened 2020-10-13
MAISTRA-1846 Add globs to service name regexes

dgn:
This fixes the problem of missing control plane metrics. The issue was that we're using revisions, which the prometheus config was not compatible with.


rcernich:
Is the issue number correct for this?  MAISTRA-1924 seems to be related to pilot crashing when namespaces are deleted.

rcernich (approved):

dgn:
> Is the issue number correct for this? MAISTRA-1924 seems to be related to pilot crashing when namespaces are deleted.

Good catch, thanks :)

jwendell (approved):

PR #604, opened 2020-10-22
MAISTRA-1938 Ensure all ports are forwarded to the proxy by default

luksa:
When using the init container, all ports are forwarded, but the
template for CNI is such that iptables rules are configured to
only forward the ports that are specified in the pod definition,
while all others go straight to the application container.

This causes the AuthorizationPolicy to not have any effect on
these other ports.

### DIFF
--- a/build/patch-charts.sh
+++ b/build/patch-charts.sh
@@ -421,6 +421,12 @@ function patchSidecarInjector() {
   # use the correct cni network defintion
   sed_wrap -i -e '/podRedirectAnnot:/,$s/istio-cni/{{ .Values.istio_cni.istio_cni_network }}/' \
       ${HELM_DIR}/istio-control/istio-discovery/files/injection-template.yaml
+
+  # MAISTRA-1938 Ensure all ports are forwarded to the proxy by default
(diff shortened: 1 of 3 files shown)


rcernich:
Hey @luksa, I think we should change the default value for `includeInboundPorts` in global.yaml instead.

knrc (approved):

rcernich (approved):

rcernich:
fwiw, i don't think the global proxy setting has ever worked.  I think the fix should be:
```
traffic.sidecar.istio.io/includeInboundPorts: "{{ annotation .ObjectMeta `traffic.sidecar.istio.io/includeInboundPorts` | default .Values.global.proxy.includeInboundPorts | default (includeInboundPorts .Spec.Containers) }}"
```

And we should change the default in global.yaml to `includeInboundPorts: "*"`

knrc:
@luksa do you think it's worth switching to @rcernich's suggestion?

luksa:
Yes, definitely. All the other settings can be set via `global.proxy.xxx` defaults.

rcernich:
I've opened a new PR which incorporates the smcp setting and uses '*' as the default.  #605

PR #605, opened 2020-10-22
MAISTRA-1938 includeInboundPorts should default to '*'

rcernich:

### DIFF
--- a/build/patch-charts.sh
+++ b/build/patch-charts.sh
@@ -421,6 +421,16 @@ function patchSidecarInjector() {
   # use the correct cni network defintion
   sed_wrap -i -e '/podRedirectAnnot:/,$s/istio-cni/{{ .Values.istio_cni.istio_cni_network }}/' \
       ${HELM_DIR}/istio-control/istio-discovery/files/injection-template.yaml
+  # use global.proxy.includedInboundPorts if specified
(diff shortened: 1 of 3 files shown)


knrc (approved):

brian-avery (approved):

rcernich:
There were a couple of other changes required to get this working correctly, namely the status port and port 15090 need to be excluded to prevent weird data showing up in kiali.

As part of my testing, I discovered that the AppliedSpec and AppliedValues fields were not being updated correctly.  This was cause by using merge patches to update the status.  This works great when fields change, but not so great when fields are removed.

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -421,7 +423,7 @@ func (r *controlPlaneInstanceReconciler) PostStatus(ctx context.Context) error {
 	log.Info("Posting status update", "conditions", r.Status.Conditions)
 	if err := r.Client.Get(ctx, client.ObjectKey{Name: r.Instance.Name, Namespace: r.Instance.Namespace}, instance); err == nil {
 		instance.Status = *r.Status.DeepCopy()
-		if err = r.Client.Status().Patch(ctx, instance, common.NewStatusPatch(instance.Status)); err != nil && !(apierrors.IsGone(err) || apierrors.IsNotFound(err)) {
+		if err = r.Client.Status().Update(ctx, instance); err != nil && !(apierrors.IsGone(err) || apierrors.IsNotFound(err)) {
```
luksa: This will have a significant effect on the entire operation of the reconciler, because the UPDATE operation will fail on conflicts, whereas PATCH won't. A failed update causes the reconciler to imediately run again, at which point it may fail again until the local cache is finally updated by the watch event.

luksa: I fixed how we perform the patch operation in https://github.com/maistra/istio-operator/pull/606. Please review the PR and then revert this change.

rcernich: done.  i was worried about that.  thanks for fixing it correctly.

PR #611, opened 2020-10-27
MAISTRA-1902: Fix disablePolicyChecks logic in yaml template

jwendell:


rcernich (approved):

dgn:
Aren't we already setting a default value in `global.yaml`? We can just put in the value as it is

jwendell:
@dgn The issue is when we set that value to `false`. Currently, because of ` | default true`, it is **always** set to true.
See for instance: https://github.com/helm/helm/issues/3308

dgn:
> @dgn The issue is when we set that value to `false`. Currently, because of ` | default true`, it is **always** set to true.
> See for instance: [helm/helm#3308](https://github.com/helm/helm/issues/3308)

Ah, I see! What I had in mind was just doing:

```yaml
  disablePolicyChecks: {{ .Values.global.disablePolicyChecks }} 
```

-- because we know the value will be set (it's defined in global.yaml, which is always applied first in the operator). But I'm fine with the current approach, too.

dgn (approved):

PR #612, opened 2020-10-27
MAISTRA-1964 use MergeFrom patch instead of Merge patch during reconcile

rcernich:

### DIFF
--- a/pkg/controller/common/helm/patch.go
+++ b/pkg/controller/common/helm/patch.go
@@ -2,10 +2,9 @@ package helm
 
 import (
 	"context"
-	"fmt"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/kubectl/pkg/util"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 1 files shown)


brian-avery (approved):

pkg/controller/common/helm/patch.go
```diff
@@ -26,22 +25,30 @@ func NewPatchFactory(k8sClient client.Client) *PatchFactory {
...
-	newObj runtime.Object
+	oldObj *unstructured.Unstructured
+	newObj *unstructured.Unstructured
 }
 
 func (p *basicPatch) Apply(ctx context.Context) (*unstructured.Unstructured, error) {
-	if err := p.client.Patch(ctx, p.newObj, client.Merge, client.FieldOwner("istio-operator")); err != nil {
-		return nil, err
+	var patch client.Patch
+	if originalBytes, err := util.GetOriginalConfiguration(p.oldObj); err == nil && len(originalBytes) > 0 {
```
dgn, line 40: This part is a bit strange to me. The problem is that the lastAppliedConfiguration annotation is explicitly set by us, but it won't be updated when making changes using `kubectl` (at least not for me) - so we're creating a Patch based on possibly outdated information. I'm not sure that was intended here? It's neat, because it allows you to change a container image for example without the operator overwriting your change, but it will also mean that changes made by the user are never overwritten except the SMCP change triggering the patch is touching those very fields

dgn (commented):
I just tested this and I'm not sure I have enough context on what problem this is supposed to solve. What I saw was that my changes to any of the resources created by the operator would never be overwritten

luksa:
A few things:
- If the user modifies the object using `kubectl edit` or `kubectl patch`, the `lastAppliedConfiguration` annotation remains unchanged. The operator won't override the user's changes. 
- If the user applies changes using `kubectl apply`, the annotation is updated to include those changes. On the next reconcile, the operator will therefore override the user's changes. 
- The operator updates Kiali in a different part of the code; these changes won't affect those updates (the Jira mentions Kiali, but this PR doesn't seem to address that part).
- Before this PR, the difference between the old and the new resource state was calculated in the API server. Now, the operator calculates it from its own view of the world (i.e. the cached resource). If the cached resource is stale, the resulting update may not be exactly what we want. This may not be a real problem, but we need to be aware of it.
- The operator should render the charts again (or cache the rendered charts) and calculate the diff from them instead of the `lastAppliedConfiguration`. It should probably also take advantage of `metadata.managedFields` to determine whether to update a field or not (this probably depends on whether the SMCP was updated). I was planning on making these changes in 2.1.0. Not sure if it's wise to push a partial solution into 2.0.0.

luksa (commented):
A few things:
- If the user modifies the object using `kubectl edit` or `kubectl patch`, the `lastAppliedConfiguration` annotation remains unchanged. The operator won't override the user's changes. 
- If the user applies changes using `kubectl apply`, the annotation is updated to include those changes. On the next reconcile, the operator will therefore override the user's changes. 
- The operator updates Kiali in a different part of the code; these changes won't affect those updates (the Jira mentions Kiali, but this PR doesn't seem to address that part).
- Before this PR, the difference between the old and the new resource state was calculated in the API server. Now, the operator calculates it from its own view of the world (i.e. the cached resource). If the cached resource is stale, the resulting update may not be exactly what we want. This may not be a real problem, but we need to be aware of it.
- The operator should render the charts again (or cache the rendered charts) and calculate the diff from them instead of the `lastAppliedConfiguration`. It should probably also take advantage of `metadata.managedFields` to determine whether to update a field or not (this probably depends on whether the SMCP was updated). I was planning on making these changes in 2.1.0. Not sure if it's wise to push a partial solution into 2.0.0.

pkg/controller/common/helm/patch.go
rcernich: Yes, that was the intent, but I'm still parsing @luksa's comments below.

rcernich:
Hey @luksa, my understanding is that `Merge` strategy will only update fields that are explicitly set in the new resource.  If the update removes a field, that field will not be updated unless the template explicitly set the field to `null`, which most don't and which probably won't work as most fields are not explicitly `nullable`.  This is the problem that you recently fixed in the status code and the problem that I am trying to fix here.

In 1.1, before switching to `Merge`, we manually calculated the patch using a three way strategy from `lastAppliedConfiguration`, the current configuration, and the updated configuration.  I don't think this new approach is very different from that, other than that it completely ignores changes outside of apply.

I didn't think about users using apply to make their changes, which would cause them to be overwritten after a reconcile.  I think it's beneficial to try to preserve this behavior, but shouldn't be a requirement.  Regarding Kiali and Jaeger, I really need to verify that these never get updated.  The intent with those is to provide an initial install, which can be configured in SMCP, but then delegate their maintenance to the user (except for any fields that are derived from SMCP settings, e.g. Grafana URL, etc. in Kiali).  That said, I believe these resources are still being updated on reconciles.

luksa (approved):

PR #613, opened 2020-10-29
MAISTRA-1968 enable status when analysis is enabled

rcernich:

### DIFF
--- a/resources/helm/v2.0/istio-control/istio-discovery/templates/configmap.yaml
+++ b/resources/helm/v2.0/istio-control/istio-discovery/templates/configmap.yaml
@@ -147,7 +147,7 @@
     defaultConfig:
       #
       # TCP connection timeout between Envoy & the application, and between Envoys.
-      connectTimeout: 10s
+      #connectTimeout: 10s
       #
       ### ADVANCED SETTINGS #############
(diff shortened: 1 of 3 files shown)


dgn (approved):

resources/helm/v2.0/istio-control/istio-discovery/templates/configmap.yaml
```diff
@@ -147,7 +147,7 @@
     defaultConfig:
       #
       # TCP connection timeout between Envoy & the application, and between Envoys.
-      connectTimeout: 10s
+      #connectTimeout: 10s
```
dgn, line 150: did this cause issues?

luksa (approved):

resources/helm/v2.0/istio-control/istio-discovery/templates/configmap.yaml
rcernich: Apparently, it's no longer in the schema and galley was barfing when it tried deserializing it.  I don't think it affected validation, but it did emit an error in the istiod logs.  (Also, it was a bit of an iterative process trying to figure out how to test this, so this might have been a problem I encountered along the way that may have been a red herring.)

PR #621, opened 2020-11-06
MAISTRA-1996 preserve v1 content on conversion errors; MAISTRA-1991 preserve unused jaeger config for v1 to v2 round tripping

rcernich:

### DIFF
--- a/pkg/apis/maistra/conversion/conversion_test.go
+++ b/pkg/apis/maistra/conversion/conversion_test.go
@@ -874,6 +874,221 @@ var (
 				},
 			}),
 		},
+		{
+			name: "MAISTRA-1983.user.2",
+			smcpv1: v1.ControlPlaneSpec{
+				Version: "v1.1",
+				// specifying both Template and Profiles saves having to duplicate
+				// everything in roundTripped, i.e. this will round trip exactly
+				Template: "default",
+				Profiles: []string{"default"},
+				Istio: v1.NewHelmValues(map[string]interface{}{
+					"gateways": map[string]interface{}{
+						"istio-ingressgateway": map[string]interface{}{
+							"autoscaleEnabled":      false,
+							"externalTrafficPolicy": "Local",
+							"ior_enabled":           false,
+							"ports": []interface{}{
+								map[string]interface{}{
+									"name":       "status-port",
+									"nodePort":   int64(30158),
+									"port":       int64(15020),
+									"protocol":   "TCP",
+									"targetPort": int64(15020),
+								},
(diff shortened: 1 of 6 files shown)


pkg/apis/maistra/conversion/conversion_test.go
```diff
@@ -275,6 +276,604 @@ var (
...
+		{
+			name: "MAISTRA-1983.2",
+			smcpv1: v1.ControlPlaneSpec{
+				Version: "v1.1",
+				Istio: v1.NewHelmValues(map[string]interface{}{
+					"tracing": map[string]interface{}{
+						"jaeger": map[string]interface{}{
+							"elasticsearch": map[string]interface{}{
+								"nodeCount":        int64(5),
+								"redundancyPolicy": nil,
```
dgn, line 361: is there a reason you set this to `nil` specifically?

dgn: nevermind I just tested with different string values and it is preserved

dgn (approved):

luksa (approved):

pkg/apis/maistra/conversion/conversion_test.go
rcernich: Yes, this was the issue fixed in MAISTRA-1983.  `unstructured.NestedString()` returns an error if the field is nil.  Personally, I think this is a bug in apimachinery, because if the parent field is nil, or the field doesn't exist at all, the call succeeds, but returns not found.  I suppose you could argue that the field was found and it's not a string, so...

knrc (approved):

pkg/apis/maistra/conversion/conversion_v1_to_v2.go
```diff
@@ -59,15 +60,44 @@ func v1ToV2Hacks(in *v1.ControlPlaneSpec, values *v1.HelmValues) error {
 }
 
 // Convert_v1_ControlPlaneSpec_To_v2_ControlPlaneSpec converts a v1 ControlPlaneSpec to its v2 equivalent.
-func Convert_v1_ControlPlaneSpec_To_v2_ControlPlaneSpec(in *v1.ControlPlaneSpec, out *v2.ControlPlaneSpec, s conversion.Scope) error {
+func Convert_v1_ControlPlaneSpec_To_v2_ControlPlaneSpec(in *v1.ControlPlaneSpec, out *v2.ControlPlaneSpec, s conversion.Scope) (err error) {
```
dgn, line 63: This change wasn't in here when I reviewed

PR #622, opened 2020-11-09
MAISTRA-1963 Mark gateways as unprivileged pods

dgn:
depends on https://github.com/maistra/istio/pull/201


resources/helm/v2.0/gateways/istio-egress/templates/deployment.yaml
```diff
@@ -282,6 +285,10 @@ spec:
           - name: ISTIO_META_MESH_ID
             value: "{{ $.Values.global.trustDomain }}"
           {{- end }}
+          {{- if not $gateway.runAsRoot }}
```
jwendell, line 288: Is it being applied twice?

dgn: No, this change is here because this is a backport of my upstream fix. We're not setting `$gateway.runAsRoot==false` but are instead using `sed` to make the necessary changes

dgn: ie this change comes is from the `maistra/istio` repo, I'm just pulling it in here

jwendell: ah :)

jwendell (approved):

bison (approved):

resources/helm/v2.0/gateways/istio-egress/templates/deployment.yaml
bison: I approved this. It looks fine to me. Though, I wonder why we don't set `$gateway.runAsRoot==false`.

PR #626, opened 2020-11-10
MAISTRA-1981 Mutating webhook should default v1 SMCPs to version v1.1

luksa:
Previously, the webhook defaulted both v1 and v2 SMCPs to version v2.0. Because version
v2.0 can only be deployed in a v2 SMCP, the max version for v1 SMCPs is v1.1. For v1 SMCPs
the webhook should therefore default to v1.1, not v2.0.


dgn (approved):
good catch. this is important, without this customers are getting 2.0 control planes if they didn't specify a version

rcernich (approved):

PR #627, opened 2020-11-10
MAISTRA-2003 Skip reconciliation of SMCPs that failed to convert from v1 to v2

luksa:


dgn (approved):
lgtm so far. confirming on a cluster now...

rcernich (approved):

brian-avery (approved):

dgn:
It does what it's supposed to do in that it doesn't reconcile the resource if the field is present, but it's stuck in an infinite loop (with back-off, but still) trying to reconcile it. Should we maybe just log an error and not return one, to avoid reconcialition to trigger again?

rcernich:
I don't know if there's a way around that.  Don't we now no longer update status though, if the error hasn't changed, so we're not spamming logs/events?

dgn:
> I don't know if there's a way around that. Don't we now no longer update status though, if the error hasn't changed, so we're not spamming logs/events?

We're not spamming events, but logs. It's not a big issue, we can clean this up later

rcernich:
Yeah, it would be nice if we could somehow mark the resource as reconciled, but have it still reflect the bad state.  Same is true if the spec is causing rendering errors, and really any other irrecoverable error, where the only fix is editing the resource or updating the operator.

luksa:
At one point, I changed the code so that the reconcile function didn't return an error. This prevented the loop. But I changed it back since we return the error in some similar cases. 

I agree that these sorts of errors shouldn't be logged as errors, but I guess we should fix this for all such cases.

PR #628, opened 2020-11-20
test case for additional container support in gateway

rcernich:

### DIFF
--- a/pkg/apis/maistra/conversion/gateways_test.go
+++ b/pkg/apis/maistra/conversion/gateways_test.go
@@ -1439,6 +1439,53 @@ var gatewaysTestCases = []conversionTestCase{
 			},
 		}),
 	},
+	{
+		name: "ingress.additionalContainers." + versions.V2_0.String(),
+		spec: &v2.ControlPlaneSpec{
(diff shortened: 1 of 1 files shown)


pkg/apis/maistra/conversion/gateways_test.go
```diff
@@ -1439,6 +1439,53 @@ var gatewaysTestCases = []conversionTestCase{
...
+			"gateways": map[string]interface{}{
+				"istio-ingressgateway": map[string]interface{}{
+					"name":        "istio-ingressgateway",
+					"gatewayType": "ingress",
+				},
+			},
+		}),
+		completeIstio: v1.NewHelmValues(map[string]interface{}{
+			"global": map[string]interface{}{
+				"multiCluster":  globalMultiClusterDefaults,
```
dgn, line 1473: idea for follow-up PR: maybe we should just move adding the defaults into the test code?

dgn (approved):

luksa (approved):
LGTM, but needs a Jira

dgn:
we should probably cherry-pick this to maistra-2.1 to not lose it

PR #632, opened 2020-11-20
WIP: MAISTRA-2004 do not pause when resources are installed outside cp namespace

rcernich:

This PR simply modifies the pause logic to ignore waiting when the components it is waiting on have no unready deployments.  This will fix the case where gateways are deployed outside the control plane namespace, but I think there is also an issue with allowing cluster ingress and egress gateways to be installed in a namespace other than the control plane namespace.  That could probably be split out into another issue.  Marking this as WIP to let folks weigh in with their opinions.

### DIFF
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -158,6 +158,8 @@ func GetMeshNamespaces(controlPlaneNamespace string, smmr *v1.ServiceMeshMemberR
 	}
 	meshNamespaces := sets.NewString(controlPlaneNamespace)
 	if smmr != nil {
+		// XXX: I don't think this can work, since ConfiguredMembers won't be set
+		// until after the initial install succeeds
 		meshNamespaces.Insert(smmr.Status.ConfiguredMembers...)
 	}
 	return meshNamespaces
--- a/pkg/controller/servicemesh/controlplane/reconciler.go
+++ b/pkg/controller/servicemesh/controlplane/reconciler.go
@@ -146,7 +146,7 @@ func (r *controlPlaneInstanceReconciler) Reconcile(ctx context.Context) (result 
 				reconciliationMessage = "Spec is invalid"
 			} else if versions.IsDependencyMissingError(err) {
 				reconciliationReason = status.ConditionReasonDependencyMissingError
-				reconciliationMessage = fmt.Sprintf( "Dependency %q is missing", versions.GetMissingDependency(err))
+				reconciliationMessage = fmt.Sprintf("Dependency %q is missing", versions.GetMissingDependency(err))
 			} else {
 				reconciliationReason = status.ConditionReasonReconcileError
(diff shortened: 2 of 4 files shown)


pkg/controller/common/util.go
```diff
@@ -158,6 +158,8 @@ func GetMeshNamespaces(controlPlaneNamespace string, smmr *v1.ServiceMeshMemberR
 	}
 	meshNamespaces := sets.NewString(controlPlaneNamespace)
 	if smmr != nil {
+		// XXX: I don't think this can work, since ConfiguredMembers won't be set
+		// until after the initial install succeeds
 		meshNamespaces.Insert(smmr.Status.ConfiguredMembers...)
```
luksa, line 163: Why not use `SMMR.Spec.Members` here?

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -248,7 +248,7 @@ func (r *controlPlaneInstanceReconciler) Reconcile(ctx context.Context) (result
 	} else if r.waitForComponents.Len() > 0 {
 		// if we've already begun reconciling, make sure we weren't waiting for
 		// the last component to become ready
-		readyComponents, _, readinessErr := r.calculateComponentReadiness(ctx)
+		readyComponents, unreadyComponents, readinessErr := r.calculateComponentReadiness(ctx)
```
luksa, line 251: I feel like the `calculateComponentReadiness` function should also return the actual readiness status of the resources that aren't in the mesh namespace. Right now, it returns incomplete information, which then requires computing the intersection of `r.waitForComponents` and `unreadyComponents`. This just feels like a complex workaround.

luksa (commented):
Not sure about this change. I think it would be better to properly determine the readiness of all the resources that the controller creates.

pkg/controller/servicemesh/controlplane/reconciler.go
rcernich: Agreed, but that would require a lot more work.  My goal here was to simply unblock the pause.  I had also looked at not setting the component flag, but that required a lot more refactoring.  Perhaps we could list all resources managed by the operator, i.e. not focus solely on those in the control plane's namespace.  We'd also need to ensure that we're receiving updates for those resources and that a reconcile is triggered appropriately.

In general, I think the proper solution is to create a first class gateway deployment resource that can be used to manage the installation and configuration of gateways outside the basic control plane configuration.

pkg/controller/common/util.go
rcernich: Sorry, I just noticed this when I was going through the code and it occurred to me that you could never initially install anything in another namespace with this logic, so it could only apply to updates, which also means the issue being addressed here could only be seen on an update, not a fresh install.

rcernich:
I think this has been fixed through a different PR.

PR #634, opened 2020-12-01
MAISTRA-1898: Sync with maistra/istio

jwendell:
By running `make gen`.

### DIFF
--- a/resources/helm/v2.0/gateways/istio-ingress/templates/preconfigured.yaml
+++ b/resources/helm/v2.0/gateways/istio-ingress/templates/preconfigured.yaml
@@ -14,7 +14,7 @@ spec:
     istio: ingressgateway
   servers:
   - hosts:
-    - "*.global"
+    - "*.{{ .Values.global.multiCluster.globalDomainSuffix | trim }}"
     port:
       name: tls
       number: 15443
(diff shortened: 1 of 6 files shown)


manifests-maistra/2.0.0.2/maistraoperator.v2.0.0.2.clusterserviceversion.yaml
```diff
@@ -12,7 +12,7 @@ metadata:
       The Maistra Operator enables you to install, configure, and manage an instance of Maistra service mesh. Maistra is based on the open source Istio project.
 
     containerImage: maistra/istio-ubi8-operator:2.0.0
-    createdAt: 2020-11-10T03:22:53UTC
+    createdAt: 2020-12-01T15:05:30EST
```
rcernich: Can we remove the csv files from the commit? They have a tendency to cause merge issues when the files actually change.

jwendell: @rcernich I'm assuming those timestamps are ignored by the check-gen job?

jwendell: `make-gen` should be smart enough to not make this change if this is the case

rcernich (approved):

knrc (approved):

PR #640, opened 2020-12-09
MAISTRA-1940 Wait for components when updating control planes

dgn:
We're already waiting for individual components to be ready in order when deploying the first time, but when updating a control plane we just apply everything instantly. This PR uses the same waiting mechanism when updating an existing control plane.

Note the ugly use of `hacks.ReduceLikelihoodOfRepeatedReconciliation`- but I couldn't find another way to make sure the cache is updated before we calculate readiness. This is because we're using `unstructured` to apply the resources, so the cache has no chance of updating. @luksa I'd be especially interested in our opinion here, maybe you know of a better solution.

### DIFF
--- a/build/patch-charts.sh
+++ b/build/patch-charts.sh
@@ -245,6 +245,13 @@ base:
     -e 's/{{ .Values.global.disablePolicyChecks | default "true" }}/{{ hasKey .Values.global "disablePolicyChecks" | ternary .Values.global.disablePolicyChecks true }}/' \
     ${HELM_DIR}/istio-control/istio-discovery/templates/configmap.yaml
 
+  # analysis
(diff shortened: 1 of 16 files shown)


jwendell (approved):

luksa (approved):
LGTM as a quick fix. I'll try to find a way to remove the two second hack.

jwendell:
@rcernich Any clue on this failure? It's failing on `TestExternalJaegerV1_1`

rcernich:
@dgn, sorry, i tried to add patch to fix the issue, but i think i broke the history.  i think a rebase and force push should fix the issue.

jwendell:
Closed in favor of https://github.com/maistra/istio-operator/pull/641

PR #648, opened 2020-12-15
MAISTRA-1351 Redesign ServiceMeshMemberRoll and ServiceMeshMember controllers

luksa:


jwendell:
@luksa Could you check if this failure is legit?

pkg/apis/maistra/v1/servicemeshmember_types.go
```diff
@@ -90,6 +90,16 @@ type ServiceMeshMemberStatus struct {
...
 
+	// The generation of the ServiceMeshControlPlane object observed by the
+	// controller during the most recent reconciliation of this
+	// ServiceMeshMember.
+	ServiceMeshGeneration int64 `json:"meshGeneration,omitempty"`    // TODO: do we need this field at all?
+
+	// The reconciled version of the ServiceMeshControlPlane object observed by
+	// the controller during the most recent reconciliation of this
+	// ServiceMeshMember.
+	ServiceMeshReconciledVersion string `json:"meshReconciledVersion,omitempty"`	// TODO: do we need this field at all?
```
dgn, line 101: what's the version stored here? the resourceVersion or the control plane version?

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -171,364 +183,340 @@ func (r *MemberRollReconciler) Reconcile(request reconcile.Request) (reconcile.R
...
-	if len(pendingMembers) > 0 {
-		instance.Status.SetCondition(v1.ServiceMeshMemberRollCondition{
-			Type:    v1.ConditionTypeMemberRollReady,
-			Status:  corev1.ConditionFalse,
-			Reason:  v1.ConditionReasonReconcileError,
-			Message: fmt.Sprintf("The following namespace(s) could not be configured: %v", pendingMembers),
-		})
+		kialiName := mesh.Status.AppliedSpec.Addons.Kiali.ResourceName()
+		roll.Status.SetAnnotation(statusAnnotationKialiName, kialiName)
+		kialiErr = r.kialiReconciler.reconcileKiali(ctx, kialiName, meshNamespace, configuredMembers.List()) // TODO: should Kiali know about all known members or just configured members?
```
dgn, line 309: I think `configuredMembers` is the only viable choice here, as we don't really do the reconciliation in this controller and have to rely on the results of the member controller. it does mean that we might be reconciling kiali wrong, and knowingly - because we added some members in this run that will only later be picked up by the member controller. that'll then lead to us reconciling kiali every time one of those namespaces gets configured, which is expensive - it triggers a redeploy.

I don't really see a way around that, except some kind of buffering/unbounce mechanism in the operator that buffers these reconciles and will only apply them once it hasn't seen updates after waiting for x seconds. it would delay the update of the configured namespaces in kiali, but it would probably still be faster than having to wait for multiple redeployments

dgn (commented):
first pass lgtm. will take another look later today

pkg/controller/common/finalizers.go
```diff
@@ -4,67 +4,62 @@ import (
...
-		err = cl.Update(ctx, obj)
-		if errors.IsNotFound(err) {
-			// Object was deleted manually before we could add the finalizer to it. This is not an error.
-			return false, nil
-		}
-		return false, err
+func getObjectMeta(obj runtime.Object) meta.Object {
+	oma, ok := obj.(meta.ObjectMetaAccessor)
+	if !ok {
+		panic("object does not implement ObjectMetaAccessor")
```
rcernich, line 62: Do we really want to panic here?  I'm guessing if this were to happen, we'd end up bouncing infinitely, with no way to correct the situation.

pkg/controller/servicemesh/memberroll/controller.go
rcernich: How about this: have a single controller that listens for changes to Member and MemberRoll.  If a Member change is the triggering event and a MemberRoll doesn't exist, that's when one gets created.  The single controller pulls all existing Member resources, merges with whatever's in the MemberRoll, performs the update, then updates the status for all the resources, Members and MemberRoll.  Upon successful completion of an update, it schedules another update event (i.e. Result.RequeueAfter=5s), which would attempt to update kiali, i.e. kiali update would only occur when processing an event where no changes are applied.  Using RequeueAfter would allow for any concurrent Member updates.

luksa: @rcernich I initially wanted to merge the two controllers into one, but the controller would be incredibly complex and error-prone. The way the two controllers' tasks are separated in this PR makes each of them very simple. 

@dgn, as for Kiali - the problem seems to be that updating the Kiali resource causes a full redeploy. I think the problem needs to be fixed there. Here, I was just wondering if Kiali needs to show namespaces that are marked as being part of the mesh, but haven't been configured yet as such. For example, to allow Kiali to display the status of these namespaces. 

Also, `.spec.members` is a bit more stable than `.status.configuredMembers`, as it only depends on the user's actions, not on the actual state of the system (a namespace can move in and out of `configuredMembers` for other reasons that can be beyond the user's control).

pkg/apis/maistra/v1/servicemeshmember_types.go
luksa: This is the operatorVersion and the SMCP generation. The only reason I introduced this field here is because we have the same field in the SMMR status (and the point of this PR is to move some of the functionality from SMMR to SMM). I'm not sure if users actually need it (or the one in SMMR), but they might use it to tell whether the member namespace has been updated to the latest SMCP generation. 

I'd actually like to remove this and the `ServiceMeshGeneration` field from the ServiceMeshMember and not add it until it's actually needed. Neither field is currently used by the operator.

pkg/controller/common/finalizers.go
luksa: I want this to fail fast and hard during development. If this function panics, the developer is passing in an object that they shouldn't. Given that this function is called whenever the developer invokes `AddFinalizer()` or `HasFinalizer()`, it should panic the very first time the developer runs the operator, so I don't see it ever failing in production.

pkg/controller/servicemesh/memberroll/controller.go
rcernich: Understood.  My general thinking was a flow similar to what is in smmr today:

1. Get SMCP
2. Get SMMR
3. Get all SMM associated with SMCP
4. Namespaces to update list is union of SMMR.spec.members and SMM list
5. Process additions/deletions as we do today
6. If no namespaces were processed (i.e. no changes to configured namespaces) and SMCP was not updated, update kiali resource, if necessary.

The only complications I see are:
1. Managing the creation/deletion of SMMR when it is being wholly managed by the operator (i.e. created in response to an SMMR).
2. Status updates for the SMM resources

The idea here is that any change to SMCP, SMM, SMMR, or namespace deletions triggers a reconciliation of member namespaces.  I think this actually simplifies finalization of SMM, as you no longer need to process deletes.  The next reconciliation will remove the namespace because the SMM resource no longer exists, so it will be missing from the list of configuredMembers (although we'll need to remove the finalizers on existing resources after updating).

pkg/controller/common/finalizers.go
rcernich: Sounds good.

pkg/controller/servicemesh/memberroll/controller.go
luksa: OK, I'm now lost. This is (exactly?) what the SMMR controller currently does. It gets the SMCP, SMMR and SMM, calculates the union, and updates Kiali. The SMM controller takes care of configuring the individual namespaces. The tasks are now nicely split across the two controllers instead of the SMMR controller doing everything. 

So what exactly is the problem here? The SMMR controller's job is very straightforward. It aggregates the statuses of the SMMs, updates Kiali and then updates the SMMR status.The SMM controller's job is also straightforward - when a namespace contains a SMM resource, the controller configures the namespace and updates the SMM status. I don't see why we'd have to make the system more complicated than this.

rcernich: Sorry, I misunderstood the purpose of this PR then.  I thought we were trying to simplify the interdependencies between the two controllers, to avoid conflicts and caching issues, and to consolidate the namespace configuration logic.

So, if I understand correctly, your intent is that:

- SMM update causes its namespace to be updated to work with the mesh
- SMMR update causes its list of members to be updated to work with the mesh, adds all successfully reconciled SMM resources to its list of configuredMembers, and updates kiali with the list of configuredMembers.
- SMCP update triggers SMMR and SMM controllers to update their namespaces to the new state of the install

Do I understand correctly?

dgn: > @dgn, as for Kiali - the problem seems to be that updating the Kiali resource causes a full redeploy. I think the problem needs to be fixed there. Here, I was just wondering if Kiali needs to show namespaces that are marked as being part of the mesh, but haven't been configured yet as such. For example, to allow Kiali to display the status of these namespaces.
> 
> Also, `.spec.members` is a bit more stable than `.status.configuredMembers`, as it only depends on the user's actions, not on the actual state of the system (a namespace can move in and out of `configuredMembers` for other reasons that can be beyond the user's control).

I agree that `spec.members` would be more stable. I wonder whether the Kiali operator has safeguards on the namespaces parameter, ie whether it also checks that the user has access to those namespaces. In any case, you are probably right and we should be delegating the SAR check to Kiali in this case. They will have to verify the user has access anyway (at least I think so), so we might as well give them the user input straight away.

rcernich: One problem is that if the namespace isn't configured, the RoleBindings won't be in place to allow pilot, et al. to read resources in them (i.e. pilot will be configured to read the namespace, but the namespace won't be configured to allow it).

As for kiali, it does its own user based access checks, so if the user has access, then they can see the resources regardless of whether or not the namespace has been properly configured into the mesh.  That said, any operations performed through the UI may not produce the desired results, if pilot/galley can't pickup the changes.  (Probably a moot point, since it takes kiali minutes to update to changes to its spec anyway.)

luksa: @rcernich Yes. That's the intent, but there's one more thing: the SMMR controller also creates SMM resources in all namespaces that the mesh admin adds to `SMMR.spec.members` and if a mesh user creates the SMM, that namespace no longer gets added to `SMMR.spec.members`. Thus, the member namespaces are determined by the union of `SMMR.spec.members` and the list of SMM resources. This was one of the goals in the Jira that suggested we merge the two controllers.

As for Pilot not being able to read the namespace: this can't happen, because the namespace only becomes part of `SMMR.status.configuredMembers` when it is properly configured.

rcernich: Perfect.  Thanks for clarifying.

rcernich:
I'm fine with this being merged, once the lgtm issues are addressed.

dgn (approved):
lgtm, sorry for the long wait. I kept putting it off because of the amount of code. thank you very much for the numbered comments, that really helped grasping what's happening.

rcernich (approved):

PR #649, opened 2020-12-15
MAISTRA-2055 make doc generation stable across executions

rcernich:

I can't guarantee that this fixes the problem, but after several runs, I have not seen any changes to the generated docs.

### DIFF
--- a/tools/pkg/doc/gen.go
+++ b/tools/pkg/doc/gen.go
@@ -176,7 +176,7 @@ func (g Generator) Generate(ctx *genall.GenerationContext) error {
 	}
 
 	// Calculate cross references
-	locations, external := g.initializeCrossReferences(kindToSchema)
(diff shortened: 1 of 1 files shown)


jwendell (approved):
Does this need a jira?
Also, don't forget to cherry-pick to 2.1. A comment `/cherrypick maistra-2.1` must be enough.

rcernich:
I don't know.  It's just the tools code and shouldn't affect anything outside the docs folder (i.e. it's specifically related to the generation of those .adoc files).  The problem it fixes is pretty random and causes different docs to be generated intermittently (maybe 20% of the time).  For example, my PR yesterday, docs gen produced different results three times, which prevented the gen check from passing.  That said, I'm happy to add a JIRA, if you think it's necessary.

knrc:
It should have a jira, ideally all commits to the repos should be traceable back to a jira

dgn (approved):

PR #654, opened 2020-12-21
Look for the right yq; Check if it's the correct one

jwendell:


build/generate-manifests.sh
```diff
@@ -51,9 +45,19 @@ function checkDependencies() {
     exit 1
   fi
 
+  # Looks for yq-python (https://github.com/kislyuk/yq) first, as installed in CI image, fallbacks to yq.
+  # Note there's another yq package, written in golang, which is not the one we want: https://github.com/mikefarah/yq
```
brian-avery, line 49: We  should choose one `yq` or the other to rely on. We use the go yq in other locations to read values out of yaml files.

jwendell: @brian-avery This pr is meant to be used with https://github.com/maistra/test-infra/pull/114

If you can make the docs tooling with with the same yq used in operator (or the other way around), I'm fine. Given operator 2.1 tests are blocked because of this, I think we could merge both PR's as a quick solution and you could fix it properly (on the docs side?) when you come back your vacation.

brian-avery: Yeah, that makes sense to me. Is the docs tooling the only place that uses the go version?

brian-avery (approved):

PR #657, opened 2021-01-04
MAISTRA-2060: Point to Red Hat Jira for issues

jwendell:


rcernich (approved):

README.md
```diff
@@ -1,3 +1,7 @@
+# Issues for this repository are disabled
+
+Issues for this repository are tracked in Red Hat Jira. Please head to https://issues.redhat.com/browse/MAISTRA in order to browse or open an issue.
```
brian-avery: https://github.com/microsoft/terminal/pull/8114/files might be a better approach. Use the process users are already used to and redirect them in that path rather than making them look for other info.

jwendell: It doesn't prevent users to just skip it and create issues if they want to.

rcernich (approved):

README.md
jwendell: @brian-avery done

dgn (approved):

PR #658, opened 2021-01-05
MAISTRA-2062: updated maistra CSV to support 1.1.11

knrc:


rcernich (approved):

knrc:
Note the istio-operator_gen-check seems to be failing because of unrelated docs changes

knrc:
and interestingly enough there are no differences if I run this locally other than the createdAt lines

rcernich:
Yeah, the doc gen is unstable.  I have a PR which fixes the problem.

brian-avery (approved):

PR #659, opened 2021-01-06
Update metadata.go to clarify mesh generation

brian-avery:
This clarifies the mesh generation variable and how it's used.

### DIFF
--- a/pkg/controller/common/metadata.go
+++ b/pkg/controller/common/metadata.go
@@ -24,7 +24,7 @@ const (
 	// GenerationKey represents the generation to which the resource was last reconciled
 	GenerationKey = MetadataNamespace + "/generation"
 
-	// MeshGenerationKey represents the generation of the service mesh operator to which the resource was last reconciled. The operator will look at the operator version and generation to determine whether a control plane should be upgraded.
+	// MeshGenerationKey represents the generation of the service mesh to which the resource was last reconciled.  This uniquely identifies an installation, incorporating the operator version and the smcp resource generation.
 	MeshGenerationKey = MetadataNamespace + "/mesh-generation"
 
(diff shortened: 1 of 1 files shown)


pkg/controller/common/metadata.go
```diff
@@ -24,7 +24,7 @@ const (
 	// GenerationKey represents the generation to which the resource was last reconciled
 	GenerationKey = MetadataNamespace + "/generation"
 
-	// MeshGenerationKey represents the generation of the service mesh to which the resource was last reconciled
+	// MeshGenerationKey represents the generation of the service mesh operator to which the resource was last reconciled. The operator will look at the operator version and generation to determine whether a control plane should be upgraded.
```
rcernich: I don't like this wording.  The original wording is correct, so maybe we just use:
```
// MeshGenerationKey represents the generation of the service mesh to which the resource was last reconciled.  This is uniquely identifies an installation, incorporating the operator version and the smcp resource generation.
```

brian-avery: That makes sense. Clarifies that it's the operator. Confusion was is this the service mesh operator or service mesh control plane/data plane version. Thanks.

rcernich (approved):

pkg/controller/common/metadata.go
rcernich: Yeah, it's actually the version of the created resource, which would be defined by user settings (smcp generation), translation to values.yaml (operator version), and chart version (operator version).  Those three things determine the version of the resource that should be installed.

knrc:
@brian-avery wait until we tag the 2.0.1 release and we can get @rcernich's doc patch in

brian-avery:
Added the old. Thanks.

PR #668, opened 2021-01-18
Prepare 2.1.0

rcernich:
Updates version to 2.1.0
Initial generation of charts for v2.1
Updates v2.0 to 2.0.1.1


Makefile
```diff
@@ -218,19 +247,19 @@ ifneq "${OSSM_MANIFEST_DATE}" ""
 endif
 
 .PHONY: update-charts
-update-charts: update-1.0-charts update-1.1-charts update-2.0-charts
+update-charts: update-1.0-charts update-1.1-charts update-2.0-charts update-2.1-charts
```
jwendell, line 250: Can we drop 1.0 yet?

rcernich: We should do that as part of a separate pr/commit.

dgn (approved):

PR #670, opened 2021-01-19
MAISTRA-2089: Update templates for 2.0.2

knrc:


resources/smcp-templates/v1.1/base
```diff
@@ -3,7 +3,7 @@ kind: ServiceMeshControlPlane
 spec:
   istio:
     global:
-      tag: 1.1.11
+      tag: 1.1.12
```
jwendell, line 6: Do you also need to update v2.0 files?

knrc: Yes, good catch

knrc: I've just checked and it doesn't seem to be included in the 2.0 base

knrc: sorry, it is but a different name.  I'll update now

brian-avery (approved):

resources/smcp-templates/v1.1/base
knrc: @jwendell can you check again?  I've just pushed up the 2.0 changes matching the above

knrc:
There appears to be a number of test failures, on the surface they do not appear to be related but I'm investigating

jwendell (approved):

knrc:
I've pushed an update for one test, most of the others appear to be warnings about a retry.

jwendell:
> I've pushed an update for one test, most of the others appear to be warnings about a retry.

It worked \o/

Next time just push another commit, it makes our lives easier to review.

PR #672, opened 2021-01-26
MAISTRA-2102: Add support for a minimum supported version of the mesh…

knrc:
… for CNI daemonset

### DIFF
--- a/pkg/bootstrap/cni.go
+++ b/pkg/bootstrap/cni.go
@@ -2,6 +2,7 @@ package bootstrap
 
 import (
 	"context"
+	"k8s.io/helm/pkg/manifest"
 	"path"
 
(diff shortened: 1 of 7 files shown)


rcernich:
Hey @knrc, the CNI installation looks fine, but shouldn't we also update the validation logic to prevent 1.0 control planes from being created too, i.e. reject spec.version < minSupportedVersion?

knrc:
> Hey @knrc, the CNI installation looks fine, but shouldn't we also update the validation logic to prevent 1.0 control planes from being created too, i.e. reject spec.version < minSupportedVersion?

@rcernich Yes we should, thanks for reminding me.  I'll deal with it now.

rcernich (approved):

PR #676, opened 2021-02-10
MAISTRA-2081 Add maistra.io apiGroup access to aggregate roles

dgn:
This makes sure that normal users can CRUD ServiceMeshExtension resources.

/cherry-pick maistra-2.0

### DIFF
--- a/pkg/bootstrap/crds.go
+++ b/pkg/bootstrap/crds.go
@@ -113,6 +113,15 @@ func installCRDRole(ctx context.Context, cl client.Client) error {
 					Resources: []string{rbacv1.ResourceAll},
 					Verbs:     role.verbs,
 				},
+				{
+					APIGroups: []string{
+						"maistra.io",
+					},
+					Resources: []string{
+						"servicemeshextensions",
+					},
+					Verbs: role.verbs,
+				},
 			},
 		}
 		existingRole := crdRole.DeepCopy()


rcernich:
I wonder if we need to add this crd to the olm manifests.  At this time, olm is managing the other maistra.io resources (e.g. smcp, smmr, smm).

dgn:
> I wonder if we need to add this crd to the olm manifests. At this time, olm is managing the other maistra.io resources (e.g. smcp, smmr, smm).

That's how I had it in the beginning, but now it is managed by the operator, which is better I think, because the control plane is consuming it, not the operator - so I think OLM shouldn't be managing it

rcernich:
Agreed, but I don't think these policies ever get removed anyway, once they're created.  I was worried that this would conflict with anything olm might be doing.

dgn:
> Agreed, but I don't think these policies ever get removed anyway, once they're created. I was worried that this would conflict with anything olm might be doing.

Hmm, you're right- are we allowing SMCP edit access this way? We might have to move ServiceMeshExtensions into another apiGroup then

brian-avery (approved):

rcernich (approved):

dgn:
> @dgn: The following test **failed**, say `/retest` to rerun all failed tests:
> istio-operator_gen-check 	[5386402](https://github.com/maistra/istio-operator/commit/53864024bb1226c0afc0a5d875c6380b90b93ef7) 	[link](https://prow.maistra.io/view/gs/maistra-prow-testing/pr-logs/pull/maistra_istio-operator/676/istio-operator_gen-check/1359882409968930816) 	`/test istio-operator_gen-check`

this failure is fixed by https://github.com/maistra/istio-operator/pull/677

PR #682, opened 2021-02-23
MAISTRA-1717 Use go-based yq in generate-manifests.sh

luksa:


dgn:
I think [Jonh made some changes that resulted in the `yq` binaries being called `yq-go` and `yq-python`](https://github.com/maistra/test-infra/pull/114) (which is why the gen-check job is failing here). If this removes the need for the python variant, we should remove that logic from the Dockerfile

luksa:
The go-based yq tool always indents list entries (there's no config option to stop it from doing this), so the resulting YAMLs have slightly different whitespace. Also, some empty strings are encoded as `""` instead of `''`. The same goes for boolean values (e.g. `"true"` instead of `'true'`). 

The nice benefit of using the go-based yq tool is that it preserves comments in the YAML.

luksa:
@dgn Thanks for the tip. I changed the script to first look for `yq-go`, which should make gen-check work now.

build/generate-manifests.sh
```diff
@@ -40,22 +40,18 @@ fi
...
-  # Note there's another yq package, written in golang, which is not the one we want: https://github.com/mikefarah/yq
-  YQ=$(which yq-python 2>/dev/null || which yq 2>/dev/null || echo "")
-
+  # Looks for go-based yq (https://github.com/mikefarah/yq) first, as installed in CI image.
+  # Note there's another yq package, written in python, which is not the one we want: https://github.com/kislyuk/yq
+  YQ=$(which yq-go 2>/dev/null || which yq 2>/dev/null || echo "")
+  echo "Using yq: ${YQ}"
   if ! [ -x "$(command -v ${YQ})" ]; then
-    echo "Please install the python yq package, e.g. 'pip install --user yq'"
+    echo "Please install the go-based yq package (https://github.com/mikefarah/yq/releases/tag/v4.6.0)"
```
rcernich: Can we just add this as a tool and invoke it using `go run`?

luksa: Best idea ever! Done!

dgn:
I'm not sure how I feel about adding all these things into `vendor/` as they're not runtime dependencies. Is there a way we can `go run` it without that?

luksa:
We've already done it like this for the other build-time dependencies. We could probably not vendor these things, but don't we need to ensure that the build is 100% repeatable and thus vendoring the build-time stuff also makes sense.

jwendell:
+1 on Daniel's comment. What's the problem with relying on yq being present on the host (like the previous iteration of this PR)?

luksa:
@rcernich suggested we use `go run` as we already do for `deepcopy-gen`, `controller-gen`, `conversion-gen`. This way you just run make and everything works.

build/generate-manifests.sh
```diff
@@ -40,25 +40,7 @@ fi
...
-    echo "Please install the python yq package, e.g. 'pip install --user yq'"
-    exit 1
-  else
-    s="yq 2.*"
-    if ! [[ $(${YQ} --version) =~ $s ]]; then
-      echo "Install the correct (python) yq package, e.g. 'pip install --user yq'"
-      exit 1
-    fi
-  fi
+  :
```
jwendell, line 43: You can remove this function

jwendell (commented):
My concern is that we end up like envoy, vendoring everything but the basic compiler.

build/generate-manifests.sh
luksa: I left it in in case we need to add dependencies later. But I'll remove it.

brian-avery:
Yeah, I'm of two minds here. On the one hand, it installs everything it requires. On the other hand, it's adding extra stuff. We handle the dependencies via docker, but if you don't have docker installed this may cause issues. Is there any risk with the go mod solution if you already have, say, the python version installed?

dgn:
> Yeah, I'm of two minds here. On the one hand, it installs everything it requires. On the other hand, it's adding extra stuff. We handle the dependencies via docker, but if you don't have docker installed this may cause issues. Is there any risk with the go mod solution if you already have, say, the python version installed?

No, and that's really the big advantage of doing it like this.

I think I'm fine with merging this as-is, but we should then use the same solution in maistra.github.io

dgn (approved):

brian-avery:
Nevermind. Go run will take care of my concern.

brian-avery (approved):

PR #683, opened 2021-03-02
MAISTRA-2099 maistra updates for istio 1.8

rcernich:

Marking as WIP, as I still don't have an install fully working.


resources/helm/v2.1/istio-control/istio-discovery/templates/deployment.yaml
```diff
@@ -182,24 +181,6 @@ spec:
             value: {{ .Values.global.tls.ecdhCurves }}
 {{- end }}
 {{- end }}
-{{- if .Values.global.tls }}
```
dgn, line 185: don't we still need this?

rcernich: Yes, but it's moved to the charts in maistra/istio, so this is just removing a duplicate section (look above this line).

jwendell (approved):

brian-avery (approved):

PR #684, opened 2021-03-10
MAISTRA-2192 Concurrent reconciliation of member namespaces

luksa:


luksa:
Please review this thoroughly - especially for potential thread-safety issues in the namespace reconciler and the functions that it invokes.

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -482,52 +483,75 @@ func (r *MemberRollReconciler) reconcileNamespaces(ctx context.Context, namespac
...
+	type result struct {
+		ns string
+		err error
+	}
+	results := make(chan result, len(namespacesToProcess))
+
+	// scatter
+	for namespace := range namespacesToProcess {
+		go func(ns string) {
+			var err error
```
dgn, line 506: should we use a waitGroup here? together with a defer func in the goroutine we could make sure that even if the goroutine panics (and thus never throws a result on the channel), we don't wait forever below, but only until all goroutines are done with processing

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -482,52 +483,75 @@ func (r *MemberRollReconciler) reconcileNamespaces(ctx context.Context, namespac
...
+			if namespacesToReconcile.Has(ns) {
+				err = reconciler.reconcileNamespaceInMesh(ctx, ns)
+				if err != nil {
+					if errors.IsNotFound(err) || errors.IsGone(err) { // TODO: this check should be performed inside reconcileNamespaceInMesh
+						reqLogger.Info("namespace to configure with mesh is missing", "namespace", ns)
+						r.EventRecorder.Event(
```
dgn, line 512: should we maybe record events in the outer for loop that processes results? I think it is thread-safe, so we can do it right here, I'm just wondering

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -482,52 +483,75 @@ func (r *MemberRollReconciler) reconcileNamespaces(ctx context.Context, namespac
...
-			nsErrors = append(nsErrors, err)
-			pending.Insert(ns)
 		} else {
-			configured.Insert(ns)
+			nsErrors = append(nsErrors, res.err)
+			pending.Insert(res.ns) // TODO: do we want to have a separate list for namespaces that are being removed, but haven't yet been?
 		}
 	}
+
+	reqLogger.Info("Reconciliation of member namespaces complete", "duration", time.Now().Sub(startTime))
```
dgn, line 554: I think this is a clear sign we should start collecting metrics in the operator :slightly_smiling_face:

dgn (commented):
I think I'd prefer a waitGroup, but other than that lgtm

pkg/controller/servicemesh/memberroll/controller.go
luksa: I don't think it's necessary, as `WaitGroup` is nothing more than a counter, which we've already got (the messages in the channel). If the goroutine panics, the entire operator crashes, so it won't wait forever. The only scenario where it would is if we later add a `return` to the goroutine function without sending anything to the channel. Adding a `WaitGroup` would also not help in that case, as we'd then try to read the same number of results from the channel and end up waiting forever.

pkg/controller/servicemesh/memberroll/controller.go
luksa: I actually had it like that in the first version, but I think it's more semantically correct for the events to be recorded as soon as they occur. I mean, an Event object represents an actual event that has happened in the system. This implies that it should be recorded as soon as the actual event happens, not seconds later (or even minutes, as could theoretically happen in this case).

pkg/controller/servicemesh/memberroll/controller.go
luksa: Yeah. I actually added this so I wouldn't have to calculate the duration from the log line timestamp. Forgot to remove it afterwards. Maybe keep it in until we add metrics.

rcernich:
Overall, this looks good to me.  I'd probably would have used a wait group, as @dgn mentioned, but that's just personal preference.  One thing I was wondering: do we need to reconcile namespaces that have already been configured, if the currently installed control plane version matches the last version we sync'd to?  This would limit the amount of traffic to the api server when folks are just adding/removing members.  Not sure if we do this already or not.  If we go that route, we should consider adding the member-of label last, so if there's any error along the way, it doesn't show up as configured when we do a namespace list.

rcernich (approved):

pkg/controller/servicemesh/memberroll/controller.go
dgn: oops, yeah that's right; panic() will stop the whole process. nevermind then

luksa:
I don't think we do that, but maybe that's not a good idea, as the operator would then fail to revert any manual changes that the user makes to the owned resources. In general, a controller MUST always ensure that the actual state reflects the desired state. This means that it should check the state of each of those resources instead of relying on any other information such as the last sync version. 

As far as the member-of label goes, it's already the last thing that we add.

dgn (approved):

PR #686, opened 2021-03-12
MAISTRA-2206 Remove member-of label from SMCP namespace on deletion

luksa:


pkg/controller/common/metadata.go
```diff
@@ -21,6 +21,9 @@ const (
 	// MemberOfKey represents the mesh (namespace) to which the resource relates
 	MemberOfKey = MetadataNamespace + "/member-of"
 
+	// MemberOfKey represents the mesh (namespace) to which the resource relates
```
brian-avery: Comment is wrong here

luksa: Thanks! Fixed.

jwendell (approved):

brian-avery (approved):

dgn:
Does this also affect 1.1 operators?

luksa:
Most likely yes. I've marked the PR for cherry-picking to 1.1, too. Thanks!

PR #691, opened 2021-03-23
[WIP] MAISTRA-2057 Add missing _affinity.tpl to the istio-control helm templates

shamsher31:


shamsher31:
@dgn @rcernich @knrc 
Do I need to update charts for this PR eg:  running `make gen`  or `make update-2.1-charts` or is this PR sufficient to test?

resources/helm/v2.1/istio-control/istio-discovery/templates/_affinity.tpl
```diff
@@ -0,0 +1,93 @@
+{{/* affinity - https://kubernetes.io/docs/concepts/configuration/assign-pod-node/ */}}
```
dgn, line 1: you'll need to move this file into the `overlays` directory (instead of 2.1) and then run `make gen`, which will copy it here

resources/helm/v2.1/istio-control/istio-discovery/templates/deployment.yaml
```diff
@@ -253,5 +253,15 @@ spec:
       - name: extracacerts
         configMap:
           name: pilot-jwks-extra-cacerts-{{ .Values.revision | default "default" }}
+  {{- end }}
```
dgn, line 256: this change will either have to go into maistra/istio repo (because that's where we pull these charts from) or into a sed script in the build/ directory

shamsher31:
Closing this and created https://github.com/maistra/istio/pull/300

PR #693, opened 2021-03-29
MAISTRA-2225: Update for 2.0.3 and 1.1.13

luksa:


rcernich:
@luksa, you have to update the unit tests for the new versions.

rcernich:
In the future, we should consider pinning the charts to specific tags, so that we only encounter these gen issues when we update the referenced tags to the latest release.

rcernich:
These need to be updated: https://github.com/maistra/istio-operator/blob/maistra-2.1/pkg/controller/servicemesh/controlplane/profiles_test.go#L19-L21

rcernich (approved):

luksa:
Thanks @rcernich

knrc (approved):

brian-avery (approved):

PR #694, opened 2021-03-29
MAISTRA-2227 ignore empty yaml bytes

rcernich:

### DIFF
--- a/pkg/controller/common/helm/manifestprocessing.go
+++ b/pkg/controller/common/helm/manifestprocessing.go
@@ -73,10 +73,15 @@ func (p *ManifestProcessor) ProcessManifest(ctx context.Context, man manifest.Ma
 			allErrors = append(allErrors, err)
 			continue
 		}
+		if len(rawJSON) == 0 || string(rawJSON) == "{}" || string(rawJSON) == "null" {
+			// MAISTRA-2227 ignore empty objects.  this could happen if charts have in empty blocks, e.g. superfluous "---"
+			continue
+		}
 		obj := &unstructured.Unstructured{}
 		_, _, err = unstructured.UnstructuredJSONScheme.Decode(rawJSON, nil, obj)
 		if err != nil {
 			log.Error(err, "unable to decode object into Unstructured")
+			log.V(2).Info(fmt.Sprintf("raw bytes:\n%s\n", raw))
 			allErrors = append(allErrors, err)
 			continue
 		}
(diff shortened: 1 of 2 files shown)


knrc (approved):

pkg/controller/common/helm/manifestprocessing.go
```diff
@@ -73,10 +73,14 @@ func (p *ManifestProcessor) ProcessManifest(ctx context.Context, man manifest.Ma
 			allErrors = append(allErrors, err)
 			continue
 		}
+		if len(rawJSON) == 0 || string(rawJSON) == "{}" || string(rawJSON) == "null" {
+			// MAISTRA-2227 ignore empty objects.  this could happen if charts have in empty blocks, e.g. superfluous "---"
+			continue
+		}
```
jwendell, line 79: Can we have a test for this?

pkg/controller/common/helm/manifestprocessing.go
```diff
@@ -73,10 +73,14 @@ func (p *ManifestProcessor) ProcessManifest(ctx context.Context, man manifest.Ma
...
 		obj := &unstructured.Unstructured{}
 		_, _, err = unstructured.UnstructuredJSONScheme.Decode(rawJSON, nil, obj)
 		if err != nil {
-			log.Error(err, "unable to decode object into Unstructured")
+			log.Error(err, fmt.Sprintf("unable to decode object into Unstructured:\n%s\n", raw))
```
brian-avery: is there any chance we'd log something we don't intend to here?

brian-avery: I think it's ok. I think secrets and similar would be in other resources.

brian-avery (approved):

pkg/controller/common/helm/manifestprocessing.go
rcernich: okay, made this more like a debug entry, leaving the error as it was

pkg/controller/common/helm/manifestprocessing.go
rcernich: done

jwendell: I think the original error was not thrown in this function, was it? If my guess is true, then checking if err != nil in the test would not be enough, right?

dgn:
Is this something that can happen in 2.0, too?

rcernich:
If the charts have superfluous `---`, with no yaml between them and especially if there are comments those empty yaml blocks.

PR #696, opened 2021-04-02
MAISTRA-2023 Add support for specifying kiali resource and scheduling constraints

luksa:


pkg/apis/maistra/conversion/kiali.go
```diff
@@ -59,6 +60,81 @@ func populateKialiAddonValues(kiali *v2.KialiAddonConfig, values map[string]inte
...
+					}
+				}
+			}
+		}
+
+		nodeSelector := deploymentConfig.NodeSelector
+		if nodeSelector != nil {
+			if values, err := toValues(nodeSelector); err == nil {
+				if len(values) > 0 {
+					if err := setHelmValue(kialiValues, "deployment.nodeSelector", values); err != nil {
```
knrc, line 116: @luksa should this not be node_selector?  See


      {% if kiali_vars.deployment.node_selector|length > 0 %}
            nodeSelector:
                  {{ kiali_vars.deployment.node_selector | to_nice_yaml(indent=0) | trim | indent(8) }}
      {% else %}
            nodeSelector: null
      {% endif %}

knrc (changes requested):

pkg/apis/maistra/conversion/kiali.go
knrc: never mind, I see we have used nodeSelector in our template but are setting node_selector

knrc (approved):

pkg/apis/maistra/conversion/kiali.go
luksa: Yeah, unfortunately, Kiali uses the `node_selector` notation for its values, whereas SMCP uses `nodeSelector`. It was hard for me to decide where to perform the switch, but the existing code in `conversion/kiali.go` already used the latter format, so I kept things consistent.

jwendell (approved):

PR #699, opened 2021-04-12
MAISTRA-1956 Stabilize accessible_namespaces in the Kiali resource

luksa:
Previously, the operator reset the accessible_namespaces field in the
Kiali resource to only the control plane namespace every time the SMCP
was updated. Now, the Helm chart for the Kiali resource no longer
contains the accessible_namespaces field. This allows the
ServiceMeshMemberRoll controller to be the only component that updates
this field and ensures that the list of namespaces reflects the list
of member namespaces at all times.

Additionally, the list of namespaces now contains all the member
namespaces (either configured in SMMR.spec.members or those that
contain a ServiceMeshMember object). Previously, it only contained
configured member namespaces, which caused namespaces to be removed
from the list every time the SMCP was updated (until the namespace
was configured according to the new SMCP version).

Kiali does not require the control plane namespace to be listed
in accessible_namespaces, as it adds it automatically. Kiali also
does not need the member namespaces to be configured in any way,
so it's safe to include all member namespaces.

### DIFF
--- a/pkg/controller/servicemesh/memberroll/controller.go
+++ b/pkg/controller/servicemesh/memberroll/controller.go
@@ -290,6 +290,7 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
 	}
 
 	// 5. check each ServiceMeshMember object to see if it's configured or terminating
+	allKnownMembers := sets.NewString(roll.Spec.Members...).Insert(getNamespaces(members)...).Delete(meshNamespace)
 	var configuredMembers = sets.NewString()
 	var terminatingMembers = sets.NewString()
 	for _, member := range members.Items {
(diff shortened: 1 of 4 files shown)


pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -290,6 +290,7 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
 	}
 
 	// 5. check each ServiceMeshMember object to see if it's configured or terminating
+	allKnownMembers := sets.NewString(roll.Spec.Members...).Insert(getNamespaces(members)...).Delete(meshNamespace)
```
rcernich, line 293: I think we should also filter out namespaces that do not exist.  Personally, I don't think we should allow configuration of members that don't exist.  Maybe we can leave this as is, and move the discussion to member roll validation (I think there's an existing issue).

rcernich (approved):

pkg/controller/servicemesh/memberroll/controller.go
dgn: can't we leave that validation to kiali?

dgn (approved):

pkg/controller/servicemesh/memberroll/controller.go
luksa: Based on my tests, Kiali seems to handle missing namespaces by itself. 

I do agree that we should probably not allow users to add nonexistent namespaces to the SMMR. We don't even need to add any kind of validation to the SMMR. We can simply remove the `.spec.members` field. Actually, adding validation wouldn't even work properly, as users could delete the namespace and we'd end up with an SMMR pointing to a nonexistent namespace again. 

Mesh admins can create a member namespace, create a ServiceMeshMember resource inside it, and then grant admin access to the namespace to a particular mesh user. This way, there's no way of somebody else gaining access to the mesh simply by creating the namespace that they know is specified in the SMMR.

PR #709, opened 2021-05-05
MAISTRA-2302: Bump openshift dependency

jwendell:
This is needed because maistra/api Status object depends
on Operator's version package. This needs to be fixed as
part of https://issues.redhat.com/browse/MAISTRA-2331.

Without this we get this go mod error:
```
github.com/openshift/api@v3.9.1-0.20190924102528-32369d4db2ad+incompatible: invalid pseudo-version: preceding tag (v3.9.0) not found
```


rcernich:
Shouldn't we do this when we migrate to using maistra/api?  I don't understand the _need_ for this at this time.  I don't think maistra/api should depend on maistra/istio-operator at all, as it will create a circular dependency between api and operator.

jwendell:
> I don't think maistra/api should depend on maistra/istio-operator at all, as it will create a circular dependency between api and operator.

Yeah, me too. This is described in MAISTRA-2331. I believe this should be fixed as part of that ticket.

jwendell:
Ok, I've removed that circular dependency for now, we will need to move it from status API to operator code base. It's kind of a static method that's only relevant to the operator, should stay in operator codebase: https://github.com/maistra/istio-operator/blob/4ee387063d7835a4dd533a61b5542a5e46e8fd52/pkg/apis/maistra/status/status.go#L190

So, we are safe to close this PR. `maistra/api` doesn't depend on operator anymore.

PR #716, opened 2021-05-17
Run make gen to sync charts

jwendell:


rcernich:
You need to either drop the csv files or update the REPLACES_*_CSV variables in Makefile (which should probably be done).  You'll also have to update the unit tests.

Makefile
```diff
@@ -16,8 +16,8 @@
 
 MAISTRA_VERSION        ?= 2.1.0
 MAISTRA_BRANCH         ?= maistra-2.1
-REPLACES_PRODUCT_CSV   ?= 2.0.3
-REPLACES_COMMUNITY_CSV ?= 2.0.3
+REPLACES_PRODUCT_CSV   ?= 2.0.4
```
knrc: Shouldn't this be 2.0.5?

Makefile
```diff
@@ -16,8 +16,8 @@
...
+REPLACES_COMMUNITY_CSV ?= 2.0.4
```
knrc: I believe this should also be 2.0.5

rcernich (approved):

knrc (approved):

PR #718, opened 2021-05-20
MAISTRA-2378 Fix reconciliation of SMMR with large number of members when ovs-multitenant is used

luksa:


knrc:
@luksa I've tested this on a cluster and it works nicely, I updated the JIRA issue with the results

knrc (approved):

jwendell (approved):

PR #725, opened 2021-05-26
Fix broken release notes link used in CSV

shamsher31:


luksa (approved):

shamsher31:
Thnaks @jwendell , this really helped.

```
$ make clean gen restore-manifest-dates
```

brian-avery:
Would it be useful to have something like htmlproofer in this and other repos to make sure links stay correct?

brian-avery (approved):

shamsher31:
> Would it be useful to have something like htmlproofer in this and other repos to make sure links stay correct?

That will be good to add.

But those links are in CSV yaml so not sure if it will work.

jwendell (approved):

PR #713, opened 2021-05-11
MAISTRA-2295 chart updates for ServiceImport resource support

rcernich:
Signed-off-by: rcernich <rcernich@redhat.com>

PR #726, opened 2021-05-26
Run make gen

luksa:


jwendell:
Fixed in https://github.com/maistra/istio-operator/pull/713

PR #727, opened 2021-05-26
MAISTRA-2322 Move Kiali affinity config to SMCP.spec.runtime.components

luksa:


pkg/apis/maistra/v2/runtime.go
```diff
@@ -199,13 +199,19 @@ type CommonPodRuntimeConfig struct {
 // XXX: istio does not support full corev1.Affinity settings, hence the special
 // types here.
 type Affinity struct {
+	// +optional
+	NodeAffinity *corev1.NodeAffinity `json:"nodeAffinity,omitempty"`
```
rcernich, line 203: If we add this, will we need to modify the rest of the charts to add support for this (and PodAffinity below)?

luksa: No, I've added validation to the webhook to prevent users from setting these fields for any other component than Kiali.

jwendell (commented):
This looks like a big change. Do we really need this for 2.0? Or it's safer to have this on 2.1 only? (Jira doesn't say anything about priority or customer case).

knrc (commented):
I second Jonh's concerns, this is a big change which needs careful review especially as it appears to be changing configuration parameters which are already supported.

Do the old parameters continue to work?

luksa:
I initially added Kiali's affinity config in the wrong place. This PR adds the config to where it belongs, but still allows users to use the old config. The idea was to introduce this fix ASAP to ensure that most customers use this new config, so we can deprecate the old one and remove it from the documentation.

luksa:
Closing this. Will only implement in 2.1.

PR #733, opened 2021-06-05
MAISTRA-2404: Register SME conversion webhook only when CRD is present

jwendell:
Because SME is a control plane CRD, it's only installed after the first
reconcile. We should make sure the conversion webhook is registered only
when the CRD is present, which could be on initialization, or after the
first reconcile.

### DIFF
--- a/pkg/controller/servicemesh/webhooks/hacks.go
+++ b/pkg/controller/servicemesh/webhooks/hacks.go
@@ -10,13 +10,10 @@ import (
 	admissionv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
 	apixv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	clientapixv1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	pttypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
-	"k8s.io/apimachinery/pkg/util/json"
 	clientcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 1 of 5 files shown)


pkg/controller/servicemesh/webhooks/server.go
```diff
@@ -78,5 +86,46 @@ func Add(mgr manager.Manager) error {
...
+	return nil
+}
+
+func RegisterControlPlaneWebhooks() error {
+	if params == nil {
+		log.Info("webhooks.Add() was not called or have failed. Ignoring.")
+		return nil
+	}
+
+	if params.cpwhRegistered {
```
dgn: should we rather look at the CRD to check if we already registered a webhook?
also, does this need a mutex? we shouldn't be running >1 reconciler thread but I have some memory of @luksa working on something to that end..?

dgn (approved):

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -200,6 +201,14 @@ func (r *controlPlaneInstanceReconciler) Reconcile(ctx context.Context) (result
 			return
 		}
 
+		// Register webhooks for control plane CRD's
```
rcernich: Would it be better to invoke this from the bootstrap/crds.go, as opposed to being part of the reconcile loop?

pkg/controller/servicemesh/webhooks/server.go
```diff
@@ -78,5 +86,46 @@ func Add(mgr manager.Manager) error {
...
+		common.GetOperatorNamespace(),
+		&smeConverterServicePath,
+		webhookca.ServiceMeshExtensionCRDName); err != nil {
+		return err
+	}
+
+	log.Info("Adding Maistra ServiceMeshExtension conversion handler")
+	wh := &conversion.Webhook{}
+	wh.InjectScheme(params.mgr.GetScheme())
+	params.mgr.GetWebhookServer().Register(smeConverterServicePath, wh)
```
rcernich: Why wouldn't we continue to register to the server piece at startup?

dgn: +1

pkg/controller/servicemesh/webhooks/hacks.go
```diff
@@ -172,49 +170,41 @@ func createWebhookResources(ctx context.Context, mgr manager.Manager, log logr.L
...
 	path *string,
 	crdName string) error {
 
 	log.Info(fmt.Sprintf("Adding conversion webhook to %s CRD", crdName))
-	apixclient, err := clientapixv1.NewForConfig(mgr.GetConfig())
-	if err != nil {
+
+	crd := &apixv1.CustomResourceDefinition{}
+	if err := cl.Get(ctx, client.ObjectKey{Name: crdName}, crd); err != nil {
+		if errors.IsNotFound(err) {
```
rcernich: Do you think this would be better handled when we try registering the sme crd?  I mention this because registration only occurs for smcp at startup and if it's not there, we should probably fail.

rcernich (commented):
lgtm.  just one minor comment.

pkg/controller/servicemesh/webhooks/hacks.go
```diff
@@ -122,10 +119,11 @@ func createWebhookResources(ctx context.Context, mgr manager.Manager, log logr.L
 		return err
 	}
 
-	if err := registerConversionWebhook(ctx, mgr, log, operatorNamespace, &smcpConverterServicePath, webhookca.ServiceMeshControlPlaneCRDName); err != nil {
+	if err := RegisterConversionWebhook(ctx, cl, log, operatorNamespace, &smcpConverterServicePath, webhookca.ServiceMeshControlPlaneCRDName, true); err != nil {
 		return err
 	}
-	if err := registerConversionWebhook(ctx, mgr, log, operatorNamespace, &smeConverterServicePath, webhookca.ServiceMeshExtensionCRDName); err != nil {
+
+	if err := RegisterConversionWebhook(ctx, cl, log, operatorNamespace, &SmeConverterServicePath, webhookca.ServiceMeshExtensionCRDName, false); err != nil {
```
rcernich, line 126: do we need to check for `IsNotFound` error here, to prevent returning an error if the sme crd is not installed?

rcernich (approved):

PR #751, opened 2021-06-28
MAISTRA-2459 Don't treat NotFound as an error when deleting a resource

luksa:


pkg/controller/servicemesh/memberroll/namespace_reconciler.go
```diff
@@ -167,7 +167,7 @@ func (r *namespaceReconciler) removeNamespaceFromMesh(ctx context.Context, names
 		for _, rb := range rbList.Items {
 			logger.Info("deleting RoleBinding for mesh ServiceAccount", "RoleBinding", rb.GetName())
 			err = r.Client.Delete(ctx, &rb)
-			if err != nil {
+			if err != nil && !(apierrors.IsNotFound(err) || apierrors.IsGone(err)) {
```
rcernich, line 171: Do we need to reset `err` to `nil` here?  I ask because it looks like it's declared out of this scope and might be read elsewhere.

pkg/controller/servicemesh/memberroll/ovs_networkpolicy.go
```diff
@@ -134,7 +134,7 @@ func (s *networkPolicyStrategy) removeNamespaceFromMesh(ctx context.Context, nam
 		for _, np := range npList.Items {
 			logger.Info("deleting NetworkPolicy for mesh", "NetworkPolicy", np.GetName())
 			err = s.Client.Delete(ctx, &np)
-			if err != nil {
+			if err != nil && !(errors.IsNotFound(err) || errors.IsGone(err)) {
```
rcernich, line 137: Same as above, do we need to reset `err` to `nil`?

rcernich (approved):

shamsher31 (approved):

pkg/controller/servicemesh/memberroll/namespace_reconciler.go
luksa: No, as it gets overwritten in the next statement.

pkg/controller/servicemesh/memberroll/ovs_networkpolicy.go
luksa: Nope. It isn't used anywhere else.

knrc:
Not okay to merge, this needs to be rolled back

knrc:
We will need to reapply this once the 2.0.6.2 release is tagged, the operator is back at the built version

PR #754, opened 2021-06-30
Update comment for workingDir hack

shamsher31:
> XXX: hacks - remove before 2.0 release

I noticed this when working on the 1.9 rebase. Not sure why this was added?

### DIFF
--- a/build/patch-charts.sh
+++ b/build/patch-charts.sh
@@ -509,7 +509,6 @@ function copyGlobalValues() {
 }
 
 function hacks() {
-  echo "XXXXXXXX HACKS THAT NEED TO BE RESOLVED BEFORE 2.0 RELEASE XXXXXXXX"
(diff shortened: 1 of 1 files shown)


rcernich:
@shamsher31, did you test this change with our images?  This was put in place because the workingDir was not set to `/` in the maistra images.

shamsher31:
> @shamsher31, did you test this change with our images? This was put in place because the workingDir was not set to `/` in the maistra images.

No, I didn't get a chance to test it. Some issues with my machine. Can we update the image instead of providing a hack?

rcernich:
That was the original intent, but until the images are updated, this cannot be removed.  We can probably remove the comment though, in either case.

luksa (approved):

rcernich (approved):

PR #761, opened 2021-07-07
MAISTRA-2462 Cache NetworkAttachmentDefinition objects

luksa:


luksa:
gen-check will pass once #762 is merged.

rcernich (approved):

jwendell (commented):
Is this actually caching anything? Shouldn't you use the NAD-provided client/informers in order to make use of the cache benefit?

rcernich:
The client provided by controller-runtime has an internal cache of its own, but only when using typed objects (i.e. not unstructured).

jwendell (approved):
Thanks.

PR #771, opened 2021-07-14
MAISTRA-2511: Set minimum supported version as defined in the spec file

knrc:


build/build.sh
```diff
@@ -35,7 +35,7 @@ fi
 LD_EXTRAFLAGS+=" -X ${REPO_PATH}/pkg/version.buildStatus=${GITSTATUS}"
 
 : ${GITTAG:=$(git describe 2> /dev/null || echo "unknown")}
-LD_EXTRAFLAGS+=" -X ${REPO_PATH}/pkg/version.buildTag=${GITTAG}"
+LD_EXTRAFLAGS+=" -X ${REPO_PATH}/pkg/version.buildTag=${GITTAG} -X ${REPO_PATH}/pkg/version.minimumSupportedVersion=${MINIMUM_SUPPORTED_VERSION}"
```
jwendell, line 38: This variable doesn't seem to exist. This will break standalone/developer builds.

knrc: Not sure what you mean, this is just pulling in the existing changes from 2.0 which are already in use

jwendell: Sorry, it doesn't break the build. But `minimumSupportedVersion` doesn't exist in 2.1 branch of pkg/version.

knrc: Good catch, I'll go back and double check then resubmit this PR

knrc: Hmm, these changes are from January but it looks as if the other PR is also missing from this codebase

knrc: I'll close this for now, there's much more work to be done

PR #773, opened 2021-07-14
Enable WASM extensions by default

jwendell:
While working on enabling WASM by default (MAISTRA-2418), an issue was found (MAISTRA-2516).
This PR addresses both issues in separate commits.

Complement of https://github.com/maistra/istio-operator/pull/747

### DIFF
--- a/pkg/controller/versions/strategy_v2_1.go
+++ b/pkg/controller/versions/strategy_v2_1.go
@@ -323,6 +323,9 @@ func (v *versionStrategyV2_1) Render(ctx context.Context, cr *common.ControllerR
 		log.V(5).Info(fmt.Sprintf("rendering values:\n%s", string(rawValues)))
 	}
 
+	// Update spec.Istio back with the previous content merged with global.yaml
+	spec.Istio = v1.NewHelmValues(values)
+
 	// Validate the final AppliedSpec
 	err = v.ValidateV2Full(ctx, cr.Client, &smcp.ObjectMeta, &smcp.Status.AppliedSpec)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


rcernich:
Does this need to include the fix for `isComponentEnabled()`?

jwendell:
> Does this need to include the fix for `isComponentEnabled()`?

I don't think so. It's a separate issue and deserve its own Jira ticket.

rcernich:
Maybe we change the issue to be a little broader in scope, e.g.: enable wasm extensions by default

jwendell:
The issue is already "enable wasm extensions by default". I just think that the issue we found yesterday should be handled separated as strictly speaking it's another issue that was just discovered because of this wasm issue.

rcernich:
it may be another issue, but without the fix, wasm extensions won't be enabled by default. :/

jwendell:
Correct, then we could hold this one and only merge it when the other issue is fixed

pkg/controller/versions/strategy_v2_1.go
```diff
@@ -323,6 +323,9 @@ func (v *versionStrategyV2_1) Render(ctx context.Context, cr *common.ControllerR
 		log.V(5).Info(fmt.Sprintf("rendering values:\n%s", string(rawValues)))
 	}
 
+	// Update spec.Istio back with the previous content merged with global.yaml
+	spec.Istio = v1.NewHelmValues(values)
```
rcernich, line 327: We shouldn't be updating the spec.  We should be updating appliedValues or just using the values as it was before.

rcernich: Ignore what I said.  It looks like there were more issues than just the isComponentEnabled() invocation (which is correct with this change), with part of the code using the merged `values` and other parts using the user specified `spec.Istio`.  Also, I verified that `spec` is not the actual resource, but the results after applying profiles.

rcernich (approved):

luksa (approved):

PR #774, opened 2021-07-14
MAISTRA-2511: Set minimum supported version as defined in the spec file

knrc:


build/build.sh
```diff
@@ -35,7 +35,7 @@ fi
 LD_EXTRAFLAGS+=" -X ${REPO_PATH}/pkg/version.buildStatus=${GITSTATUS}"
 
 : ${GITTAG:=$(git describe 2> /dev/null || echo "unknown")}
-LD_EXTRAFLAGS+=" -X ${REPO_PATH}/pkg/version.buildTag=${GITTAG}"
+LD_EXTRAFLAGS+=" -X ${REPO_PATH}/pkg/version.buildTag=${GITTAG} -X ${REPO_PATH}/pkg/version.minimumSupportedVersion=${MINIMUM_SUPPORTED_VERSION}"
```
jwendell, line 38: Please initialize this variable to a reasonable default value if not provided externally, since not all builds are done via RPM build.

knrc: This is a copy of the 2.0 codebase which is already working for product builds

knrc: Note the additional code does default to all versions if this is not configured

knrc: I've confirmed that 'make compile' works without issue

jwendell: yeah, it compiles fine

pkg/version/version.go
```diff
@@ -13,19 +13,23 @@ var (
 	buildStatus      = "unknown"
 	buildTag         = "unknown"
 
+	// Minimum supported mesh version (nil (all), "v1_0", "v1_1", "v2_0", "v2_1" etc)
+	minimumSupportedVersion = ""
```
jwendell, line 17: empty means all :)

build/build.sh
jwendell: We will have to add a default value anyway in the future (even if empty), when working on MAISTRA-2107.

jwendell (approved):

build/build.sh
knrc: Sounds good

pkg/version/version.go
knrc: ah yes, not nil.  That comment is also like that in the 2.0 version

pkg/bootstrap/cni.go
```diff
@@ -37,19 +46,27 @@ func internalInstallCNI(ctx context.Context, cl client.Client, config cni.Config
...
-		return err
+	var releases []string
+	for _, version := range supportedVersions {
+		releases = append(releases, version.String())
 	}
+	values["supportedReleases"] = releases
+
+	// always install the latest version of the CNI image
+	renderings, _, err = helm.RenderChart(path.Join(versions.DefaultVersion.GetChartsDir(), "istio_cni"), operatorNamespace, values)
+	return
```
luksa, line 57: Ideally, this function should return `renderings["istio_cni"]` instead of the entire `renderings` map.

luksa (approved):
LGTM, but I'd clean up the `"istio_cni"` map key leakage from `internalRenderCNI()`.

pkg/bootstrap/cni.go
knrc: That behaviour was already present in the code, it has not changed with this PR

knrc: Sorry, when I expanded this I see you are talking about the new function

knrc: I can see what you mean, keep the knowledge about the key ("istio_cni") in one function instead of reference it in two.  I can submit a change for this later and apply it to both 2.0 and 2.1 branches.

luksa: Yes, that's what I meant.

PR #776, opened 2021-07-19
MAISTRA-2459 Don't treat NotFound as an error when deleting a resource

luksa:


jwendell (approved):

dgn (approved):

jwendell:
@luksa ok to merge?

luksa:
I don't know. This targets 2.0 and I have no idea whether we are free to merge things or not. :)

jwendell:
cc @knrc

knrc:
@luksa @jwendell +1, it was to be resubmitted after the release was tagged and that was done.

dgn:
@luksa should this also be done in 2.1? I think there are a few spots where we treat NotFound as error still there

PR #777, opened 2021-07-20
MAISTRA-2475 Fix initialization and update of Kiali.spec.deployment.accessible_namespaces

luksa:
This ensures that the .spec.deployment.accessible_namespaces field is initialized when the Kiali resource is first deployed (by the smcp controller), but is then updated only by the smmr controller. The smcp controller should never overwrite the value, as this would cause the issue described in MAISTRA-1956. Previously, we simply removed the field from the Helm chart, which resulted in the smcp controller never updating the field (it was thus only ever updated by the smmr controller), but this resulted in a different issue, where Kiali starts logging errors, because it thinks it has access to all namespaces.

### DIFF
--- a/resources/helm/overlays/istio-telemetry/kiali/templates/kiali-cr.yaml
+++ b/resources/helm/overlays/istio-telemetry/kiali/templates/kiali-cr.yaml
@@ -13,6 +13,7 @@ spec:
     strategy: "openshift"
 
   deployment:
+    accessible_namespaces: []
 {{- if and .Values.kiali.hub .Values.kiali.image }}
     image_name: "{{ .Values.kiali.hub }}/{{ .Values.kiali.image }}"
 {{- end }}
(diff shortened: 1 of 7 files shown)


resources/helm/overlays/istio-telemetry/kiali/templates/kiali-cr.yaml
```diff
@@ -13,6 +13,7 @@ spec:
     strategy: "openshift"
 
   deployment:
+    accessible_namespaces: []
```
rcernich, line 16: Does this need to be initialized to `{{ .Release.Namespace}}`?

luksa: No. Kiali adds the mesh namespace automatically.

jwendell (approved):

luksa:
Forced push to fix the commit message (should be smmr, not smm controller)

rcernich (approved):

PR #780, opened 2021-07-22
MAISTRA-2525: Remove hostNetwork permission from CNI

jwendell:
This permission is not necessary anyway.
We need to be explicit about using elevated privileges, by
adding `privileged:true` in the containers.

### DIFF
--- a/resources/helm/overlays/istio_cni/templates/daemonset.yaml
+++ b/resources/helm/overlays/istio_cni/templates/daemonset.yaml
@@ -35,7 +35,6 @@ spec:
     spec:
       nodeSelector:
         beta.kubernetes.io/os: linux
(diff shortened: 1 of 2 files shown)


rcernich (approved):

dgn:
Strange. So did adding `hostNetwork: true` just make the container privileged so that this worked?

dgn (approved):

jwendell:
@dgn Yeah, that's how it looks like. The pod runs with privileged scc already, but it seems the `privileged` field is not necessary if you have `hostNetwork`. With the latter gone, the former is required.

There's a thread `hostNetwork:true == privileged:true?` on aos-devel about it.

PR #783, opened 2021-07-27
MAISTRA-2536 Fix global proxy env injection

luksa:


luksa:
I'm not sure if fixing the issue in this manner is correct. Would it be better to modify the chart so that the existing values in `global.proxy.env` are injected?

luksa:
@dgn and I talked about this and we've come to the conclusion that setting `meshConfig.defaultConfig.proxyMetadata` is correct.

dgn (approved):

jwendell (approved):

PR #787, opened 2021-07-28
MAISTRA-2545: Disable option for operator webhook management

VikingPingvin:
Config field added, and neccesary checks are placed in the code.

To disable the (default) webhook management by the operator use the following annotation in the operator deployment manifest:
`olm.webhookManagementDisabled: "true"`

After this, it is possible to apply custom `ValidatingWebhookConfigurations` and `MutatingWebhookConfigurations` from manifests.


pkg/controller/servicemesh/webhookca/controller.go
```diff
@@ -60,6 +60,11 @@ var autoRegistrationMap = map[string]CABundleSource{
 // Add creates a new Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
+	if common.Config.OLM.WebhookManagementDisabled == true {
+		createLogger().Info("Webhook Config Management is disabled via olm configuration, not adding webhookca to watch list...")
```
dgn: nit: `not adding webhookca controller to Manager` makes more sense I think

dgn (approved):
lgtm in general. Adding a hold so @rcernich has a chance to comment.

I'm guessing you don't need to update any certs because the webhook endpoints you expose to customers use a long-lived cert?

pkg/controller/servicemesh/webhookca/controller.go
libesz: Yes. The initial cert management and the eventual renewal is made by external automation in our case.

libesz:
> lgtm in general. Adding a hold so @rcernich has a chance to comment.
> 
> I'm guessing you don't need to update any certs because the webhook endpoints you expose to customers use a long-lived cert?

Yes. The initial cert management and the eventual renewal is made by external automation in our case.

pkg/controller/servicemesh/webhookca/controller.go
VikingPingvin: Renamed. Makes sense.

pkg/controller/servicemesh/webhookca/controller.go
```diff
@@ -60,6 +60,11 @@ var autoRegistrationMap = map[string]CABundleSource{
 // Add creates a new Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
+	if common.Config.OLM.WebhookManagementDisabled == true {
```
jwendell: you don't need `== true`

pkg/controller/servicemesh/webhooks/server.go
```diff
@@ -34,8 +34,13 @@ func Add(mgr manager.Manager) error {
 	ctx := common.NewContextWithLog(common.NewContext(), log)
 	log.Info("Configuring Maistra webhooks")
 
-	if err := createWebhookResources(ctx, mgr, log, common.GetOperatorNamespace()); err != nil {
-		return err
+	if common.Config.OLM.WebhookManagementDisabled == true {
```
jwendell: ditto

jwendell:
Please create a ticket at https://issues.redhat.com/projects/MAISTRA/issues and update the commit/PR title including the ticket number, like `MAISTRA-1234: Reworked....`

pkg/controller/common/config.go
```diff
@@ -21,7 +21,8 @@ type config struct {
 
 // OLM is intermediate struct for serialization
 type olm struct {
-	Images images `json:"relatedImage,omitempty"`
+	Images                    images `json:"relatedImage,omitempty"`
+	WebhookManagementDisabled bool   `json:"webhookManagementDisabled,omitempty"`
```
luksa: Any reason why this is `webhookManagementDisabled` and not `webhookManagementEnabled` that defaults to `true`?

libesz: I think we went this direction as this will be BW compatible even when the annotation is completely missing (i.e. operator is deployed with old manifest).

VikingPingvin: > 
> 
> I think we went this direction as this will be BW compatible even when the annotation is completely missing (i.e. operator is deployed with old manifest).

Yes, that was exactly the reason.

dgn: I agree that `webhookManagementEnabled` makes more sense. You can default it to true by using `viper.SetDefault()` afaik

rcernich: I think this should probably be moved into `config.Controller`.

pkg/controller/servicemesh/webhookca/controller.go
```diff
@@ -60,7 +60,7 @@ var autoRegistrationMap = map[string]CABundleSource{
 // Add creates a new Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
-	if common.Config.OLM.WebhookManagementDisabled {
+	if common.Config.Controller.WebhookManagementEnabled {
```
dgn: missing a `!`?

pkg/controller/servicemesh/webhooks/server.go
```diff
@@ -34,7 +34,7 @@ func Add(mgr manager.Manager) error {
 	ctx := common.NewContextWithLog(common.NewContext(), log)
 	log.Info("Configuring Maistra webhooks")
 
-	if common.Config.OLM.WebhookManagementDisabled {
+	if common.Config.Controller.WebhookManagementEnabled {
```
dgn: same here

pkg/controller/servicemesh/webhookca/controller.go
VikingPingvin: Fixed in commit 13f4.

VikingPingvin:
I don't think the Viper default values will do any good for the unit tests eg. maistra2040_test.go in the case of webhookca, as it does not read in configs.

A possiblity would be to extend the unit tests by including the controller package and configuring the common structs in-place in any of the unit tests it may be needed, but that might not be the ideal solution.

Would there be a better way to solve this?

libesz:
UT is fixed now. Do we need anything else for this one?

jwendell (approved):
LGTM, Any objection on merging this?

luksa (approved):

jwendell:
@VikingPingvin Can you resolve the conflicts so that we can merge this? Thanks.

VikingPingvin:
> @VikingPingvin Can you resolve the conflicts so that we can merge this? Thanks.

Maybe I'm missing something @jwendell, but I don't see where are the conflicts you mentioned.
I can merge just fine into `maistra-2.1`.
```bash
git pull libeszfork disable-webhookmanagement
From github.com:libesz/istio-operator
 * branch              disable-webhookmanagement -> FETCH_HEAD
Auto-merging pkg/controller/servicemesh/webhookca/controller.go
Auto-merging pkg/controller/common/config.go
Auto-merging cmd/manager/main.go
Merge made by the 'recursive' strategy.
 cmd/manager/main.go                                | 6 +++---
 pkg/controller/common/config.go                    | 8 ++++++++
 pkg/controller/servicemesh/webhookca/controller.go | 5 +++++
 pkg/controller/servicemesh/webhooks/server.go      | 9 +++++++--
 4 files changed, 23 insertions(+), 5 deletions(-)
 ```
 
 Also GH says there is no conflicts.
 I think only a `tide` label is needed for this PR?

jwendell:
@VikingPingvin Sorry for the noise. GitHub was saying it could not rebase due to conflicts, that's why I said that. Thanks for the PR :)

VikingPingvin:
No worries :)
Ty also

PR #790, opened 2021-07-29
MAISTRA-2544 Improve SMCP status condition message on operator upgrade

luksa:


pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -466,14 +466,13 @@ func (r *controlPlaneInstanceReconciler) initializeReconcileStatus() {
 	var conditionReason status.ConditionReason
 	if r.isUpdating() {
 		if r.Status.ObservedGeneration == r.Instance.GetGeneration() {
-			fromVersion := r.Status.OperatorVersion
-			toVersion := buildinfo.Info.Version
-			readyMessage = fmt.Sprintf("Upgrading mesh from version %s to version %s", fromVersion[strings.LastIndex(fromVersion, "-")+1:], toVersion[strings.LastIndex(toVersion, "-")+1:])
+			readyMessage = fmt.Sprintf("Updating mesh due to operator version change (%s to %s)", r.Status.OperatorVersion, buildinfo.Info.Version)
```
dgn, line 469: Shouldn't we check that they really are different? Couldn't the re-reconcile also be due to the operator being restarted, without changing versions?

luksa: In theory, the reconciliation shouldn't be triggered in that case. And if it is, I don't know what we're supposed to do here in that case. I mean what should the reason be if we can't determine the reason why a full reconcile was triggered? With this change, we at least now log the full version, so if it ever happens that we upgrade when there's no change in the operator version and no change in the SMCP generation, at least we'll have more information than we do now.

luksa: Oh, the operator restart itself doesn't trigger a full reconcile. The smcp controller is kinda weird in that it doesn't perform the full reconciliation when a reconcile is triggered. It only performs full reconciliation when the SMCP.spec is updated or the operator version changes. In all other cases, it simply updates the readiness information. We eventually need to fix this and ensure that a full reconcile happens every time the reconcile is triggered. This will ensure that the actual state matches the desired state specified in the SMCP. But for now, this wouldn't work because it would take too much time and would also override changes that users make to the child resources.

dgn: ah okay, maybe in my case it was something else then- I thought I remembered deploying a custom operator and it would re-reconile my SMCPs. But I do also remember having to edit the SMCP to trigger it sometimes

dgn (approved):

jwendell (approved):

shamsher31 (approved):

PR #792, opened 2021-08-02
MAISTRA-2548 Reject SMCP if default gateway targets a non-member namespace

luksa:


pkg/controller/versions/util.go
```diff
@@ -138,39 +143,54 @@ func validateGateways(ctx context.Context, meta *metav1.ObjectMeta, spec *v2.Con
...
 	}
+
 	if reservedGatewayNames.Has(name) {
 		*allErrors = append(*allErrors, fmt.Errorf("cannot define additional gateway named %q", name))
 	}
 }
 
+func validateGatewayNamespace(name string, gateway *v2.GatewayConfig, meshNamespaces sets.String, allErrors *[]error) {
+	if (gateway.Enabled == nil || *gateway.Enabled) && gateway.Namespace != "" && !meshNamespaces.Has(gateway.Namespace) {
+		*allErrors = append(*allErrors, fmt.Errorf("namespace %q for ingress gateway %q must be part of the mesh", gateway.Namespace, name))
```
jwendell: What about:
`namespace is not part of the mesh for the gateway %s/%s`

jwendell (approved):
lgtm, just a minor suggestion

pkg/controller/versions/util.go
luksa: This was actually an existing error message, but you're right that it could be improved.

dgn (approved):
lgtm, but could this be problematic for customers that already have an SMCP with a default gateway in a non-system namespace? ie should we only verify this for 2.1 control planes?

luksa:
Not problematic because the reconcile already fails in this case, as it checks whether the charts try to create any resource outside the mesh.

dgn:
> Not problematic because the reconcile already fails in this case, as it checks whether the charts try to create any resource outside the mesh.

Nevermind I was misreading the intention here- I thought this would limit to control plane namespace, not non-member namespace

PR #795, opened 2021-08-02
MAISTRA-2532 Gather readiness information for gateways outside control plane namespace

luksa:


pkg/controller/servicemesh/controlplane/readiness_test.go
```diff
@@ -0,0 +1,371 @@
...
+
+}
+
+func newDeployment(name, namespace, owner, component string, ready bool) *appsv1.Deployment {
+	var readyReplicas int32
+	if ready {
+		readyReplicas = 1
+	} else {
+		readyReplicas = 0
+	}
```
shamsher31: nit: Zero value for `int32` is by default `0`. Anyways these are tests, so not a big deal.
```suggestion
	var readyReplicas int32
	if ready {
		readyReplicas = 1
	}
```

shamsher31: Same for [this](https://github.com/maistra/istio-operator/pull/795/files#diff-db0700473d726d545874acf32f5a10643272d68e6df319fc276b2e5c84f9e043R325)  and [this](https://github.com/maistra/istio-operator/pull/795/files#diff-db0700473d726d545874acf32f5a10643272d68e6df319fc276b2e5c84f9e043R350).

pkg/controller/servicemesh/controlplane/readiness.go
```diff
@@ -175,8 +177,16 @@ func (r *controlPlaneInstanceReconciler) calculateComponentReadinessMap(ctx cont
 			},
 		},
 	}
+
+	namespaces, err := r.getNamespacesToCheck()
+	if err != nil {
+		return nil, err
+	}
+
+	log.V(2).Info("Calculating readiness", "namespaces", namespaces)
```
shamsher31, line 197: Can we add a component name here?

pkg/controller/servicemesh/controlplane/readiness.go
```diff
@@ -209,36 +221,76 @@ func (r *controlPlaneInstanceReconciler) isCNIReady(ctx context.Context) (bool,
...
+			metaObject, err := meta.Accessor(obj)
+			if err != nil {
+				return err
+			}
+			if component, ok := metaObject.GetLabels()[common.KubernetesAppComponentKey]; ok {
+				ready, exists := readinessMap[component]
+				readinessMap[component] = (ready || !exists) && isReady(obj)
+			} else {
+				// how do we have an owned resource with no component label?
+				log.Error(nil, "skipping resource for readiness check: resource has no component label", obj.GetObjectKind(), metaObject.GetName())
```
shamsher31: This should be an error or just log info?

pkg/controller/servicemesh/controlplane/readiness_test.go
luksa: Of course. Thanks!

pkg/controller/servicemesh/controlplane/readiness.go
luksa: What do you mean by "component"?

pkg/controller/servicemesh/controlplane/readiness.go
luksa: Good catch. This should be logged as V(3).Info, as it's definitely not an error that whoever is overseeing the operator needs to know about, since it can be caused by the user creating the resource.

pkg/controller/servicemesh/controlplane/readiness.go
shamsher31: I mean Deployment, StatefulSet, and DaemonSet. Move this log to the `typesToCheck` loop and just print for which type it is calculating readiness. Not sure if that is going to be super useful.

shamsher31 (approved):

pkg/controller/servicemesh/controlplane/readiness.go
luksa: I don't think we need to log at the lower abstraction layer than we currently are, because it doesn't really matter what the resource type is. I actually added this line because I needed to see which namespaces it was scanning. Also, we do log each resource that we find (at V(3)), so if someone needs to see additional details of the scan, they can just turn that on.

shamsher31: Sounds good.

pkg/controller/servicemesh/controlplane/readiness.go
```diff
@@ -219,36 +232,76 @@ func (r *controlPlaneInstanceReconciler) isCNIReady(ctx context.Context) (bool,
...
+		for _, gw := range gateways.EgressGateways {
+			if gw.Namespace != "" {
+				namespaces.Insert(gw.Namespace)
+			}
+		}
+	}
+
+	// ensure we only check namespaces that are part of the mesh
+	smmr := &maistrav1.ServiceMeshMemberRoll{}
+	if err := r.Client.Get(context.TODO(), client.ObjectKey{Namespace: r.Instance.Namespace, Name: common.MemberRollName}, smmr); err != nil {
```
jwendell, line 295: do we have an informer/cache to use instead of making an api call?

jwendell (approved):
lgtm, just a minor comment

pkg/controller/servicemesh/controlplane/readiness.go
rcernich: Internally, I think the controller-runtime client delegates to a cache, as long as you're not using unstructured object types.

luksa: Yes, this hits the cache.

PR #796, opened 2021-08-02
MAISTRA-2546 Ensure gateways outside control plane namespace trigger SMCP reconciliation

luksa:


jwendell (approved):

pkg/controller/common/helm/manifestprocessing.go
```diff
@@ -204,11 +206,12 @@ func (p *ManifestProcessor) addMetadata(obj *unstructured.Unstructured, componen
...
 		common.KubernetesAppVersionKey:   p.appVersion,
 		common.KubernetesAppComponentKey: component,
-		common.KubernetesAppPartOfKey:    "istio",
-		common.KubernetesAppManagedByKey: "maistra-istio-operator",
+		common.KubernetesAppPartOfKey:    common.KubernetesAppPartOfValue,
+		common.KubernetesAppManagedByKey: common.KubernetesAppManagedByValue,
 		// legacy
 		// add owner label
-		common.OwnerKey: p.owner,
+		common.OwnerKey: p.owner.Namespace,
```
dgn, line 213: ~if we don't pass real values here anymore (we're calling this with `types.NamespacedName{}`), do we still need to set them?~

Nevermind we do at the call for normal manifests, just not for CNI

pkg/controller/servicemesh/controlplane/controller.go
```diff
@@ -84,28 +83,13 @@ func add(mgr manager.Manager, r *ControlPlaneReconciler) error {
...
+	if err = c.Watch(&source.Kind{Type: &appsv1.StatefulSet{}}, enqueueRequestForSMCP, ownedResourcePredicates); err != nil {
 		return err
 	}
-	if err = c.Watch(&source.Kind{Type: &appsv1.DaemonSet{}},
-		&handler.EnqueueRequestForOwner{
-			IsController: true,
-			OwnerType:    &v2.ServiceMeshControlPlane{},
-		},
-		ownedResourcePredicates); err != nil {
+	if err = c.Watch(&source.Kind{Type: &appsv1.DaemonSet{}}, enqueueRequestForSMCP, ownedResourcePredicates); err != nil {
```
dgn, line 92: is this going to do anything now? CNI daemonset isn't tied to any SMCP

pkg/controller/servicemesh/controlplane/controller.go
```diff
@@ -84,28 +83,13 @@ func add(mgr manager.Manager, r *ControlPlaneReconciler) error {
...
+	if err = c.Watch(&source.Kind{Type: &appsv1.Deployment{}}, enqueueRequestForSMCP, ownedResourcePredicates); err != nil {
 		return err
 	}
-	if err = c.Watch(&source.Kind{Type: &appsv1.StatefulSet{}},
-		&handler.EnqueueRequestForOwner{
-			IsController: true,
-			OwnerType:    &v2.ServiceMeshControlPlane{},
-		},
-		ownedResourcePredicates); err != nil {
+	if err = c.Watch(&source.Kind{Type: &appsv1.StatefulSet{}}, enqueueRequestForSMCP, ownedResourcePredicates); err != nil {
```
dgn, line 89: nit: not sure why we have this even. maybe we should file an issue to remove it?

pkg/controller/common/helm/manifestprocessing.go
```diff
@@ -204,11 +206,12 @@ func (p *ManifestProcessor) addMetadata(obj *unstructured.Unstructured, componen
 		common.KubernetesAppInstanceKey:  p.appInstance,
 		common.KubernetesAppVersionKey:   p.appVersion,
 		common.KubernetesAppComponentKey: component,
-		common.KubernetesAppPartOfKey:    "istio",
-		common.KubernetesAppManagedByKey: "maistra-istio-operator",
+		common.KubernetesAppPartOfKey:    common.KubernetesAppPartOfValue,
+		common.KubernetesAppManagedByKey: common.KubernetesAppManagedByValue,
 		// legacy
```
dgn, line 211: not sure how legacy this is now that we're using it. let's remove the comment

pkg/controller/servicemesh/controlplane/controller.go
dgn: tbh I'm not sure this did ever work...?

rcernich: I think CNI status is calculated elsewhere, as it's not part of the actual control plane installation.  This code would only check daemon sets that were installed through the charts.  As there are none, It can probably be safely removed at some point in the future.  For now, it's just an extra watch.

pkg/controller/servicemesh/controlplane/controller.go
rcernich: Agreed.  This is only necessary if the control plane includes stateful sets.

rcernich (approved):

dgn (approved):

PR #799, opened 2021-08-03
MAISTRA-2557 add support for processing v1 CRDs

rcernich:

### DIFF
--- a/pkg/bootstrap/crds.go
+++ b/pkg/bootstrap/crds.go
@@ -14,6 +14,8 @@ import (
 
 	"github.com/Masterminds/semver"
 	"github.com/ghodss/yaml"
+	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
@@ -52,7 +54,7 @@ func InstallCRDs(ctx context.Context, cl client.Client, chartsDir string) error 
 	crdPath := path.Join(chartsDir, "istio-init", "files")
 	crdDir, err := os.Stat(crdPath)
 	if err != nil || !crdDir.IsDir() {
-		return fmt.Errorf("Cannot locate any CRD files in %s", crdPath)
+		return fmt.Errorf("cannot locate any CRD files in %s", crdPath)
 	}
 	err = filepath.Walk(crdPath, func(path string, info os.FileInfo, err error) error {
 		if err != nil || info.IsDir() {
@@ -167,7 +169,7 @@ func processCRDFile(ctx context.Context, cl client.Client, fileName string) erro
 	}
 
 	allErrors := []error{}
-	for index, raw := range releaseutil.SplitManifests(string(buf.Bytes())) {
+	for index, raw := range releaseutil.SplitManifests(buf.String()) {
 		crd, err := decodeCRD(common.NewContextWithLog(ctx, log.WithValues("file", fileName, "index", index)), raw)
 		if err != nil {
 			allErrors = append(allErrors, err)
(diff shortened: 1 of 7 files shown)


pkg/bootstrap/crds.go
```diff
@@ -181,35 +182,52 @@ func processCRDFile(ctx context.Context, cl client.Client, fileName string) erro
...
 	}
-	obj := &apiextensionsv1beta1.CustomResourceDefinition{}
-	_, _, err = unstructured.UnstructuredJSONScheme.Decode(rawJSON, nil, obj)
+	v1beta1obj := &apiextensionsv1beta1.CustomResourceDefinition{}
+	_, _, err = unstructured.UnstructuredJSONScheme.Decode(rawJSON, nil, v1beta1obj)
 	if err != nil {
-		log.Error(err, "unable to decode object into Unstructured")
-		return nil, err
+		v1obj := &apiextensionsv1.CustomResourceDefinition{}
+		_, _, err = unstructured.UnstructuredJSONScheme.Decode(rawJSON, nil, v1beta1obj)
```
jwendell: should be `v1obj`, no?

pkg/bootstrap/crds.go
```diff
@@ -181,35 +182,52 @@ func processCRDFile(ctx context.Context, cl client.Client, fileName string) erro
...
+		if v1obj.GroupVersionKind().GroupKind().String() == "CustomResourceDefinition.apiextensions.k8s.io" {
+			return v1obj, nil
+		}
+		return nil, nil
 	}
-	if obj.GroupVersionKind().GroupKind().String() == "CustomResourceDefinition.apiextensions.k8s.io" {
-		return obj, nil
-	} else {
+	if v1beta1obj.GroupVersionKind().GroupKind().String() != "CustomResourceDefinition.apiextensions.k8s.io" {
 		return nil, nil
```
jwendell, line 210: should return an error?

pkg/controller/common/test/api.go
```diff
@@ -148,26 +150,33 @@ func (a *AbstractActionFilter) Handles(action clienttesting.Action) bool {
...
-    return a
+	resourceAndSub := strings.SplitN(resource, "/", 2)
+	a.Resource = resourceAndSub[0]
+	if len(resourceAndSub) > 1 {
+		a.Subresource = resourceAndSub[1]
+	}
+	return a
+}
+
+// In initializes the version whithin which the filter should apply.  Use "*"
```
jwendell: s/In/Version/

pkg/controller/common/test/assertions.go
```diff
@@ -34,22 +34,29 @@ type ActionAssertionFactory struct {
 // between resource an subresource, e.g. deployments/status.  Use "*" to match
 // all resources.
 func (f *ActionAssertionFactory) On(resource string) *ActionAssertionFactory {
-    f.AbstractActionFilter.On(resource)
-    return f
+	f.AbstractActionFilter.On(resource)
+	return f
+}
+
+// In initializes the version whithin which the created assertion should apply.
```
jwendell: s/In/Version/

pkg/controller/common/test/verifiers.go
```diff
@@ -40,6 +40,13 @@ func (f *ActionVerifierFactory) On(resource string) *ActionVerifierFactory {
 	return f
 }
 
+// In initializes the version whithin which the created verifier should apply.
```
jwendell: s/In/Version/

pkg/bootstrap/crds.go
rcernich: That would be changing the original behavior.  If we return an error, then CRD processing fails as will all reconciles from this point on.  I believe all this does is silently fails to keep reconcile going, without breaking everything, as the failure is unrecoverable (i.e. no user action can fix it).  I'll add a log message though, so there's something visible.

pkg/bootstrap/crds.go
rcernich: good catch!

dgn (approved):

jwendell (approved):

rcernich:
I've been able to verify that existing installs for v1.1, v2.0, and v2.1 all succeed.

PR #800, opened 2021-08-04
MAISTRA-2395: Implement Rate Limiting

jwendell:
Under `techPreview`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,10 +98,6 @@ tags
 
 ### VisualStudioCode ###
 .vscode/*
-!.vscode/settings.json
-!.vscode/tasks.json
-!.vscode/launch.json
-!.vscode/extensions.json
 .history
 
 
--- a/resources/helm/overlays/rls/templates/configmap.yaml
+++ b/resources/helm/overlays/rls/templates/configmap.yaml
@@ -1,0 +1,23 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  labels:
+    app: rls
+    istio.io/rev: {{ .Values.revision | default "default" }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  name: rls-{{ .Values.revision | default "default" }}
+  namespace: {{ .Release.Namespace }}
+data:
+  config.yaml: |
+    #      domain: echo-ratelimit
+    #      descriptors:
+    #        - key: PATH
+    #          value: "/"
+    #          rate_limit:
+    #            unit: minute
+    #            requests_per_unit: 1
+    #        - key: PATH
+    #          rate_limit:
+    #            unit: minute
+    #            requests_per_unit: 100
--- a/resources/helm/overlays/rls/templates/configmap.yaml
+++ b/resources/helm/overlays/rls/templates/configmap.yaml
@@ -10,14 +10,6 @@ metadata:
   namespace: {{ .Release.Namespace }}
 data:
   config.yaml: |
-    #      domain: echo-ratelimit
-    #      descriptors:
-    #        - key: PATH
-    #          value: "/"
-    #          rate_limit:
-    #            unit: minute
-    #            requests_per_unit: 1
-    #        - key: PATH
-    #          rate_limit:
-    #            unit: minute
-    #            requests_per_unit: 100
+{{- if .Values.rateLimiting.rawRules }}
+{{ toYaml .Values.rateLimiting.rawRules | indent 4 }}
+{{- end }}
--- a/resources/helm/overlays/rls/templates/deployment.yaml
+++ b/resources/helm/overlays/rls/templates/deployment.yaml
@@ -59,6 +59,10 @@ spec:
           value: /data
         - name: RUNTIME_SUBDIRECTORY
           value: ratelimit
+        - name: RUNTIME_WATCH_ROOT
+          value: "false"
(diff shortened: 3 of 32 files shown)


resources/helm/overlays/rls/templates/deployment.yaml
```diff
@@ -0,0 +1,81 @@
...
+        - name: RUNTIME_SUBDIRECTORY
+          value: ratelimit
+        {{- if .Values.rateLimiting.rls.env }}
+        {{- range $key, $val := .Values.rateLimiting.rls.env }}
+        - name: {{ $key }}
+          value: "{{ $val }}"
+        {{- end }}
+        {{- end }}
+        ports:
+        - containerPort: 8080
```
jwendell, line 73: need to add 8081

resources/helm/v2.1/rls/templates/autoscale.yaml
```diff
@@ -0,0 +1,26 @@
+{{- if and .Values.rateLimiting.rls.autoscaleEnabled .Values.rateLimiting.rls.autoscaleMin .Values.rateLimiting.rls.autoscaleMax }}
+apiVersion: autoscaling/v2beta1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: rls-{{ .Values.revision | default "default" }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    maistra-version: "2.1.0"
+    maistra-version: "2.1.0"
```
jwendell: duplicated

resources/helm/overlays/rls/templates/configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+  labels:
+    app: rls
+    istio.io/rev: {{ .Values.revision | default "default" }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  name: rls-{{ .Values.revision | default "default" }}
+  namespace: {{ .Release.Namespace }}
+data:
+  config.yaml: |
+    #      domain: echo-ratelimit
```
shamsher31: these are intentionally commented?

shamsher31 (commented):
Can we provide Redis as a default `storageBackend` as part of the RLS configuration, Instead of relying completely on the user to configure it? If they wish they can configure other `storageBackend` like Memcache as mentioned in your PR.
Ref: https://github.com/maistra/istio/commit/64e752da3b7fd5340f9d48aa20bdb8216082e56a#diff-e752e7417eb33824010438a331dc9ad18c8279dfb5cc3606ef89b9883511a7b2R28

dgn:
> Can we provide Redis as a default `storageBackend` as part of the RLS configuration, Instead of relying completely on the user to configure it? If they wish they can configure other `storageBackend` like Memcache as mentioned in your PR.
> Ref: [maistra/istio@64e752d#diff-e752e7417eb33824010438a331dc9ad18c8279dfb5cc3606ef89b9883511a7b2R28](https://github.com/maistra/istio/commit/64e752da3b7fd5340f9d48aa20bdb8216082e56a#diff-e752e7417eb33824010438a331dc9ad18c8279dfb5cc3606ef89b9883511a7b2R28)

I think we explicitly don't want to do that because we don't want to support users running Redis. We'll point them to Redis documentation from our docs; it's really easy to spin up a single-pod Redis instance, but there are tons of options for more advanced deployments which we don't want to have to deal with.

resources/helm/overlays/rls/templates/configmap.yaml
dgn: maybe we can make this configurable, by rendering e.g. `values.rls.config` into YAML here - that way, users could apply free-form configuration. I'm afraid if we don't offer that, we'll just overwrite the users configuration with every reconcile and this chart will be of little use

jwendell: @shamsher31 yeah, my intention was to provide a simple example - the same found in upstream doc - commented.

@dgn yeah, that's right, this will be overwritten. I think we should use `techPreview.rateLimiting.rawRules` for this, where we get raw yaml from user and apply here. We don't want to wrap this in a API, that's what upstream is working on ATM.

resources/helm/overlays/rls/templates/deployment.yaml
```diff
@@ -0,0 +1,89 @@
...
+        - containerPort: 8081
+        - containerPort: 6070
```
shamsher31, line 75: What is this port is used for? This should also be added to the Service Port list?

shamsher31 (approved):

.gitignore
```diff
@@ -98,10 +98,6 @@ tags
 
 ### VisualStudioCode ###
 .vscode/*
-!.vscode/settings.json
```
dgn, line 101: ?

resources/helm/overlays/rls/templates/deployment.yaml
jwendell: This is a [debug port](https://github.com/envoyproxy/ratelimit/#debug-port). Not sure if it's worth to be exposed on svc...

dgn (commented):
This is missing the default imageNames for the new component (*-rhel8 and *-ubi8) in `resources/smcp-templates/v2.1` - it's required for when you use a custom hub/tag

jwendell:
@dgn done, thanks

dgn (approved):

PR #801, opened 2021-08-06
MAISTRA-2556 Default gateway label 'app' to gateway (service) name

dgn:
This is the most elegant solution I could come up with: if the user did not set the `app` label in the SMCP, we set it to the gateway name. For default gateways, this doesn't change anything. But additional gateways with custom names will now have the label `app: <GATEWAY_NAME>`. If users want the old behaviour, they have to set the `app` label to `istio-ingressgateway` or `istio-egressgateway` themselves in the SMCP.

Initially, I had tried removing all the default labels when users supplied custom labels in the SMCP, but that didn't work - turns out you need the `istio` and `app` labels for ingress gateways to work. Not sure why -- I couldn't find a reference anywhere. I'm asking on Istio Slack now.


rcernich (approved):
I think this is a great solution, simple, effective.

jwendell (approved):

brian-avery (approved):

PR #806, opened 2021-08-10
MAISTRA-2576 Updated generated CRD to v1; MAISTRA-2575 propagate preserve unknown fields to nested properties

rcernich:
MAISTRA-2575 fixes the issue with proxy not coming up because of an invalid filter configuration.


dgn (approved):
interesting. do we know how this ended up causing the xDS problems?

jwendell (approved):
Confirmed with this commit, bookinfo deploys fine.

rcernich:
@dgn, the x-kubernetes-preserve-unknown-fields does not propagate to nested schema.  The `value` field in the EnvoyFilter had no schema associated with it, so the entire value field was pruned to `{}`, which caused an empty filter to be merged into the filter chain.  The filters in question were the internal filters supporting metrics (e.g. prometheus, metadata exchange, etc.).  The fix here was to walk the schema and add the x-kubernetes-preserve-unknown-fields to any field that was an object/array type that didn't have a schema defined.

PR #807, opened 2021-08-11
OSSM-316 Enable persistent volume for prometheus deployed with servicemesh

shamsher31:


rcernich:
Could you add a unit test to verify the generated resource is correct, when volumes are configured in smcp?

shamsher31:
@rcernich Let me know if my understanding is correct, Prometheus can be configured using runtime components? 
After looking into the codebase, I found that currently runtime components don't have support for setting `volumes` and `volumeMounts`, which means we need to add these 2 options in order to make it configurable? 

```yaml
spec:
  runtime:
    components:
      prometheus:
        container:
          volumeMounts:
          - mountPath: /prometheus
            name: prometheus-k8s-db
        pod:
          volumes:
          - name: prometheus-k8s-db
            persistentVolumeClaim:
              claimName: prometheus-db-pv-claim
```

pkg/apis/maistra/v2/runtime.go
```diff
@@ -193,6 +193,19 @@ type CommonPodRuntimeConfig struct {
...
+}
+
+// Restrict Volumes to only support PVC 
+type Volumes struct {
+	// Volume's name. Must be a DNS_LABEL and unique within the pod
+	Name string `json:"name,omitempty"`
+
+	// PersistentVolumeClaim references the user's PVC in the same namespace.
+	// This volume finds the bound PV and mounts that volume for the pod
+	PersistentVolumeClaim corev1.PersistentVolumeClaimVolumeSource `json:"persistentVolumeClaim,omitempty"`
```
shamsher31, line 208: @rcernich Do we also want to support more common volume types eg: `configMap`, `hostPath` and `emptyDir`?

jwendell:
Closing as we need to rework the solution.

PR #810, opened 2021-08-11
Add tests

jwendell:


manifests-maistra/2.1.0/maistraoperator.v2.1.0.clusterserviceversion.yaml
```diff
@@ -603,6 +603,8 @@ spec:
                     - istio-operator
                     - --config
                     - /etc/operator/olm/config.properties
+                    - --zap-devel
+                    - --zap-level=2
```
rcernich, line 607: I don't think we want these in our production code

PR #819, opened 2021-08-20
Run make gen

luksa:


jwendell (approved):

shamsher31 (approved):

luksa:
@jwendell, the automator does this automatically when charts are updated in 2.1. But it looks like it doesn't update 2.1 when the 2.0 charts are updated.

dgn (approved):

PR #821, opened 2021-08-24
MAISTRA-2624 Improve operator startup by reducing number of API discovery calls

luksa:


go.mod
```diff
@@ -71,3 +72,5 @@ replace github.com/evanphx/json-patch => github.com/evanphx/json-patch v4.9.0+in
 
 // pkg disappeard from bitbucket
 replace bitbucket.org/ww/goautoneg => github.com/munnerz/goautoneg v0.0.0-20120707110453-a547fc61f48d
+
+replace github.com/operator-framework/operator-sdk => github.com/luksa/operator-sdk v0.0.0-20210824135520-3b25565223d4
```
jwendell: I'd rather have this submitted/reviewed/pushed upstream so we don't need maintain a fork.

luksa: We're using a _very_ old version. I don't think this code is even in operator-sdk anymore. This is just a quick fix until we upgrade.

jwendell: Yeah, my opinion is that we submit this upstream, get it merged there and we will get it on the next rebase, in 2.2.

luksa: I'm saying we can't submit this upstream, because the code that I've fixed has been removed completely. They won't accept PRs for the ancient version of the library that we're using. In 2.2 we'll use an up-to-date version and there won't be any more unnecessary API discovery calls. But since the changes I've made are a considerable improvement (and they are simple enough to not introduce any new issues) I think we should get them into 2.1.

jwendell: I got it Marko. I understand all you are saying. My point is: Do we really need this for 2.1, especially at this point where we are working on stabilizing? If the team agrees that we do need this, then at least we should fork the repo under maistra org, open a pr there and use this fork here.

Or, we could wait for 2.2 when we can bump to the latest sdk (my vote).

luksa: The existing code performs full API discovery for each of the operator's custom resources and also a full API discovery simply to determine whether one specific resource type exists. This is such a bad implementation that I'd almost consider it a bug. If the fix wasn't as simple as it is, I would be fine with waiting for 2.2, but since the fix is really simple (perform a single API discovery call and then use the retrieved data when processing all of the custom resources), I don't see why we wouldn't include it in 2.1 The work is finished and there's no additional work that QE needs to do. 

Sure I could fork the repo under maistra, but since we're vendoring dependencies anyway (and we already do vendor in third party repos that are owned by individuals and not organizations) I didn't think it was that important to fork it.

rcernich:
For the stuff from k8sutil, would it make sense to just move this into the operator.  For the other changes, is the resource list dynamic?  I ask because the first time the operator starts, there won't be any istio.io resources installed on the cluster and I assume those won't be in the initial list used to initialize the client.

luksa:
The resource list is only used at startup, when some metrics stuff gets initialized. Basically, my changes was just the following:

From:
```
foreach cr in operatorCRs {
  resourceList := getResourceList()
  doSomethingWith(resourceList, cr)
}
```

to:
```
resourceList := getResourceList()
foreach cr in operatorCRs {
  doSomethingWith(resourceList, cr)
}
```

That's one of the changes. The other is in `isNamespaced()`, which previously performed another partial API Discovery for each CR, but now simply uses the data it already read when it performed the full API discovery request.

In all cases, `resourceList` is then discarded. If it's later needed again, it is loaded from the API server again (for example, when you create the client).

rcernich:
Thanks @luksa. So this means that we were only ever reporting metrics for CRs whose definitions were installed as part of the operator, which means this change doesn't affect existing metrics.  And, even if it did register metrics for istio types, the operator still would have required a restart, as the metrics were only initialized at operator startup.  Do I have this right?

luksa:
We specify the list of CRs (see #1), so restarting won't have any effect. We only register the maistra CRs. And I've just noticed that we register maistra/v1 twice instead of registering v2.

#1 https://github.com/Maistra/istio-operator/blob/maistra-2.1/cmd/manager/main.go#L234-L242

rcernich (approved):

jwendell:
I really think we should fork the sdk into maistra and use from there, where anybody in the team has access to.

luksa:
Okay, I've now forked operator-sdk to Maistra/operator-sdk.

go.mod
```diff
@@ -71,3 +72,5 @@ replace github.com/evanphx/json-patch => github.com/evanphx/json-patch v4.9.0+in
 
 // pkg disappeard from bitbucket
 replace bitbucket.org/ww/goautoneg => github.com/munnerz/goautoneg v0.0.0-20120707110453-a547fc61f48d
+
+replace github.com/operator-framework/operator-sdk => github.com/Maistra/operator-sdk v0.0.0-20210824135520-3b25565223d4
```
jwendell: nit: Maistra was renamed to maistra

go.sum
```diff
@@ -651,6 +653,8 @@ github.com/lightstep/lightstep-tracer-common/golang/gogo v0.0.0-20190605223551-b
 github.com/lightstep/lightstep-tracer-go v0.18.0/go.mod h1:jlF1pusYV4pidLvZ+XD0UBX0ZE6WURAspgAczcDHrL4=
 github.com/lithammer/dedent v1.1.0/go.mod h1:jrXYCQtgg0nJiN+StA2KgR7w6CiQNv9Fd/Z9BP0jIOc=
 github.com/lovoo/gcloud-opentracing v0.3.0/go.mod h1:ZFqk2y38kMDDikZPAK7ynTTGuyt17nSPdS3K5e+ZTBY=
+github.com/luksa/operator-sdk v0.0.0-20210824135520-3b25565223d4 h1:XHjawxilNjGhqQbLu+oeFx2u94edMxcEEeGr+FNPOJQ=
```
jwendell: missing a go mod tidy?

jwendell (approved):

go.sum
luksa: fixed. thanks

go.mod
luksa: fixed

PR #824, opened 2021-08-25
MAISTRA-2627 Remove namespace field from istiod-internal ClusterRoleBinding definition

luksa:


dgn (approved):

jwendell (approved):

resources/helm/v2.0/istio-control/istio-discovery/templates/meshclusterrolebinding.yaml
```diff
@@ -2,7 +2,6 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: istiod-internal-{{ .Values.revision | default "default" }}-{{ .Release.Namespace }}
-  namespace: {{ .Release.Namespace }}
```
dgn, line 5: This is going to be overwritten until the 2.0 PR goes in I guess

dgn: Shoud we just not make the change here?

luksa: That's why I'll wait for the 2.0 PR to be merged before merging this. If I remove this change and merge this 2.1 PR and then the 2.0 PR, we'll then get gen-check errors in all other 2.1 PRs until we run `make gen`. I thought just adding this change here was simpler.

PR #826, opened 2021-08-25
MAISTRA-2589 Do not allow 2.1 control planes with Mixer enabled

dgn:
This will make sure people cannot upgrade to 2.1 when they have Mixer enabled.

### DIFF
--- a/pkg/controller/versions/strategy_v2_1.go
+++ b/pkg/controller/versions/strategy_v2_1.go
@@ -125,6 +125,7 @@ func (v *versionStrategyV2_1) ValidateV2(ctx context.Context, cl client.Client, 
 	allErrors = validatePolicyType(ctx, meta, spec, v.version, allErrors)
 	allErrors = validateTelemetryType(ctx, meta, spec, v.version, allErrors)
 	allErrors = v.validateProtocolDetection(ctx, meta, spec, allErrors)
+	allErrors = v.validateMixerDisabled(spec, allErrors)
 	return NewValidationError(allErrors...)
 }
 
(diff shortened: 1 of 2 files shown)


jwendell (approved):

pkg/controller/versions/strategy_v2_1.go
```diff
@@ -142,6 +143,16 @@ func (v *versionStrategyV2_1) validateProtocolDetection(ctx context.Context, met
 	return allErrors
 }
 
+func (v *versionStrategyV2_1) validateMixerDisabled(spec *v2.ControlPlaneSpec, allErrors []error) []error {
+	if spec.Policy != nil && (spec.Policy.Type == v2.PolicyTypeMixer || spec.Policy.Mixer != nil) {
+		allErrors = append(allErrors, fmt.Errorf("support for policy.type '%s' and policy.Mixer options been removed in v2.1, please use another alternative", v2.PolicyTypeMixer))
```
shamsher31: nit:
```suggestion
		allErrors = append(allErrors, fmt.Errorf("support for policy.type %q and policy.Mixer options have been removed in v2.1, please use another alternative", v2.PolicyTypeMixer))
```

pkg/controller/versions/strategy_v2_1.go
```diff
@@ -142,6 +143,16 @@ func (v *versionStrategyV2_1) validateProtocolDetection(ctx context.Context, met
...
+	}
+	if spec.Telemetry != nil && (spec.Telemetry.Type == v2.TelemetryTypeMixer || spec.Telemetry.Mixer != nil) {
+		allErrors = append(allErrors, fmt.Errorf("support for telemetry.type '%s' and telemetry.Mixer options have been removed in v2.1, please use another alternative", v2.TelemetryTypeMixer))
```
shamsher31:
```suggestion
		allErrors = append(allErrors, fmt.Errorf("support for telemetry.type %q and telemetry.Mixer options have been removed in v2.1, please use another alternative", v2.TelemetryTypeMixer))
```

shamsher31 (approved):

rcernich (approved):
Should we also scan for config.istio.io resources, or is it good enough to ensure mixer is disabled?  I assume, if mixer is disabled, then any handler/instance resources will simply be ignored if mixer isn't present, but I'm not sure.  The goal is to ensure that existing sidecars won't be affected when disabling mixer (i.e. if mixer goes away, do they need to be reinjected?).  That said, if this is an issue, then it probably exists in 2.0 too, in which case we can create a follow-up issue to deal with that.  (Maybe it's good enough to "fail open" when mixer is removed.)

dgn:
> Should we also scan for config.istio.io resources, or is it good enough to ensure mixer is disabled? I assume, if mixer is disabled, then any handler/instance resources will simply be ignored if mixer isn't present, but I'm not sure. The goal is to ensure that existing sidecars won't be affected when disabling mixer (i.e. if mixer goes away, do they need to be reinjected?). That said, if this is an issue, then it probably exists in 2.0 too, in which case we can create a follow-up issue to deal with that. (Maybe it's good enough to "fail open" when mixer is removed.)

I think generally we should try to avoid looking at resources outside the SMCP when validating it, simply for performance reasons. If customers have handler/instance resources around, but Mixer disabled, they cannot expect them to work anyway

PR #827, opened 2021-08-30
MAISTRA-2606 Add missing commits from 2.0

luksa:


pkg/controller/servicemesh/controlplane/profiles_test.go
```diff
@@ -160,6 +160,37 @@ func TestProfiles(t *testing.T) {
 				smcp.Spec.TechPreview.RemoveField("wasmExtensions")
 			},
 		},
+		{
+			name: "MAISTRA-2409",
+			v:    versions.V2_0,
```
jwendell, line 165: do we need to duplicate this test with a 2.1 version?

luksa: Not really, as the code is the same for both 2.0 and 2.1. If anything, we'd have to extend all the tests to check for both 2.0 and 2.1, but that's a whole new PR.

dgn: I recently filed an issue to cover that: https://issues.redhat.com/browse/MAISTRA-2574

dgn (approved):

pkg/controller/servicemesh/controlplane/profiles_test.go
jwendell: The code is the same, but the charts not.

jwendell (approved):

PR #829, opened 2021-09-01
Instruct istio-iptables to capture DNS

jwendell:
So that federation works.

iptables rules that redirect DNS traffic are only applied if this
variable is true.


luksa:
Are we sure this won't affect existing users that don't care about federation at all? It seems like we need to enable this only when federation is enabled.

jwendell:
This is already done in https://github.com/maistra/istio/commit/25628477a92e25fe180513860d3904d70c3c6981
Closing.

PR #833, opened 2021-09-03
MAISTRA-2637 Disable ingress controller in istiod

luksa:


rcernich (approved):

dgn (approved):
I see you're also adding it to the 2.0 strategy - might that break anyone?

luksa:
I'm in the process of making sure 2.0 works on OpenShift 4.9+ and I found that even 2.1 doesn't work there. This needs to go into 2.0, too. Not sure if anyone needs this. If they do, maybe we need to only disable it in 4.9+ clusters.

PR #834, opened 2021-09-03
MAISTRA-2641 Prune non-deprecated versions of Ingress and WebhookConfigurations

luksa:


rcernich (approved):

pkg/controller/servicemesh/controlplane/pruner.go
```diff
@@ -42,8 +42,8 @@ var (
 		gvk("networking.k8s.io", "v1", "NetworkPolicy"),
 		gvk("rbac.authorization.k8s.io", "v1", "RoleBinding"),
 		gvk("rbac.authorization.k8s.io", "v1", "Role"),
-		gvk("admissionregistration.k8s.io", "v1beta1", "MutatingWebhookConfiguration"),
-		gvk("admissionregistration.k8s.io", "v1beta1", "ValidatingWebhookConfiguration"),
+		gvk("admissionregistration.k8s.io", "v1", "MutatingWebhookConfiguration"),
```
jwendell, line 45: should we remove beta resources from this list? or just add the v1 to it so both end up pruned?

pkg/controller/servicemesh/controlplane/pruner.go
```diff
@@ -31,7 +31,7 @@ var (
 		gvk("apps", "v1", "Deployment"),
 		gvk("apps", "v1", "DaemonSet"),
 		gvk("apps", "v1", "StatefulSet"),
-		gvk("extensions", "v1beta1", "Ingress"),
+		gvk("networking.k8s.io", "v1", "Ingress"),
```
jwendell, line 34: should we remove beta resources from this list? or just add the v1 to it so both end up pruned?

pkg/controller/servicemesh/controlplane/pruner.go
rcernich: This is okay, so long as the clusters we're deploying to support v1.  The resource is the same, it's just whether you're looking at it through the eyes of v1 or v1beta1.  Just like smcp v1 vs v2.

jwendell (approved):

PR #835, opened 2021-09-03
MAISTRA-2644 Convert v1beta1 webhookconfigurations to v1 when deploying istio manifests

luksa:


rcernich (approved):
Looks good.  Just be aware, that I found some issues when writing the CRD converter, in that it didn't account for `preserveUnknownFields`.  If you haven't already, you might want to check the verification code to ensure it's not omitting validation of specific settings if the resource is being updated (e.g. you could still use a v1 with preserveUnknownFields if it was replacing an old v1beta1 resource).  jfyi

jwendell (approved):

PR #836, opened 2021-09-06
OSSM-569 Use default resources for prometheus sidecar

dgn:
There's currently no way for users to set resources for the istio-proxy that comes with prometheus. This can cause scheduling problems in kubernetes.

Adding a /hold as we'll need to decide whether we want to ship this in 2.0.8. I verified that it works locally.

PS: This needs a manual cherry-pick for 2.1 as the template for prometheus has become an overlay

### DIFF
--- a/build/patch-prometheus.sh
+++ b/build/patch-prometheus.sh
@@ -100,6 +100,14 @@ function prometheus_patch_deployment() {
     rollingUpdate:\
       maxSurge: 25%\
       maxUnavailable: 25%' $file
+  
+  sed_wrap -i -e '/        - name: istio-proxy/ a\
+          resources:\
+{{- if .Values.global.proxy.resources }}\
+{{ toYaml .Values.global.proxy.resources | indent 12 }}\
+{{- else }}\
+{{ toYaml .Values.global.defaultResources | indent 12 }}\
+{{- end }}' ${HELM_DIR}/istio-telemetry/prometheus/templates/deployment.yaml
 }
 
 function prometheus_patch_service() {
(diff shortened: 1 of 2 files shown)


build/patch-prometheus.sh
```diff
@@ -100,6 +100,14 @@ function prometheus_patch_deployment() {
     rollingUpdate:\
       maxSurge: 25%\
       maxUnavailable: 25%' $file
+  
+  sed_wrap -i -e '/        - name: istio-proxy/ a\
+          resources:\
+{{- if .Values.global.proxy.resources }}\
+{{ toYaml .Values.global.proxy.resources | indent 12 }}\
+{{- else }}\
+{{ toYaml .Values.global.defaultResources | indent 12 }}\
```
rcernich, line 109: Should we verify that defaultResources is not empty before using it?

dgn:
Removing hold as discussed on our call

build/patch-prometheus.sh
dgn: shouldn't be necessary (and I blatantly copy-pasted this from other places where it's used exactly like this). it's always gonna have [something set](https://github.com/maistra/istio-operator/blob/d6c8a08b2cd839bd655bd8b8ae847a0cedfea2c1/resources/helm/v2.0/global.yaml#L354-L356)

rcernich: My concern would be that we end up with an empty resources block, which I'm not sure would be valid yaml or is interpreted as a null value.

jwendell (approved):

rcernich (approved):

PR #841, opened 2021-09-08
MAISTRA-2322 Move Kiali affinity config to SMCP.spec.runtime.components

luksa:


shamsher31 (commented):
Based on the comment in the ticket, does it mean it is reverted in 2.0?

Also, the `.adoc` file is missing from this PR, is that optional?

luksa:
Unfortunately, we can't revert the old API change until we deprecated it. This fix was initially meant to go into 2.0, but will now only go into 2.1. We'll then deprecate the old API ASAP.

shamsher31 (approved):

rcernich (approved):

PR #847, opened 2021-09-10
MAISTRA-1399 default to NetworkPolicy if an unknown network type is in use

rcernich:
This simply removes the allowlist check and defaults to NetworkPolicy.

Signed-off-by: rcernich <rcernich@redhat.com>

PR #848, opened 2021-09-10
MAISTRA-1399 default to NetworkPolicy if an unknown network type is i…

rcernich:
…n use (#847)


jwendell (approved):
better create another jira for 2.1?

dgn:
> better create another jira for 2.1?

nah, I don't think it's necessary. When cherry-picking to unreleased versions we usually just recycle the old JIRA

dgn (approved):

PR #849, opened 2021-09-13
MAISTRA-2668 Set sideEffects on webhooks

luksa:


rcernich (approved):

jwendell (approved):
Why don't we fix this in the charts? Isn't it easier?

rcernich:
I suspect this will save us having to make changes in the older maistra charts.

luksa:
I used the same strategy as for the CRDs, which we also convert on the fly. I guess we could do these changes in the charts, but it would be more error-prone.

PR #852, opened 2021-09-14
MAISTRA-2669: Webhook controller should handle secret deletion

jwendell:
So that when the "cacerts" user-defined secret is deleted, it updates
the webhook with the old caBundle from the "istio-ca-secret" secret.


luksa (approved):

dgn (approved):
good catch!

dgn:
let's discus on our call if this can be merged still despite code freeze (I vote yes because of minimal risk, but it's not technically a blocker)

PR #855, opened 2021-09-14
MAISTRA-2671 Ensure Galley can start on OCP 4.9+

luksa:


dgn (approved):

dgn (commented):
Wait.. why do we also exclude Pod, Endpoints and Service?

dgn:
> Wait.. why do we also exclude Pod, Endpoints and Service?

Is it to circumvent the CRD presence check on startup? I see in the code that galley will watch builtin types anyway, so I guess that's why it's okay to add them to the list?

luksa:
> Wait.. why do we also exclude Pod, Endpoints and Service?

If you don't specify `--excludeResourceKinds`, it defaults to `Service,Endpoints,Pod,Node,Namespace` (see [1] and [2])

[1] https://github.com/Maistra/istio/blob/22bc71bcc46c9b6b4b4b3d5017a1b78d5efab327/galley/pkg/config/util/kuberesource/resources.go#L66
[2] https://github.com/Maistra/istio/blob/8dc359773107a150114e28c27ae85ae188c163ae/galley/pkg/source/kube/builtin/types.go#L46-L242

luksa:
@dgn @jwendell any thoughts on whether to fix this with this PR or with https://github.com/maistra/istio/pull/438? Why aren't we fixing everything by changing the charts in maistra/istio instead of making things unnecessary complex via patch-charts?

jwendell:
heh, this is a broader discussion we should have for 2.2 I think. JFYI, I agree with you, I'm against patching charts in operator.

dgn:
+1 I think in the future we should be making all changes directly in maistra/istio, it's just so much easier than writing sed scripts. For this issue, I don't mind either way

luksa:
OK, then let's go with this PR for now. I'll close the other one. @dgn can you approve this one?

dgn (approved):

luksa:
Oops, sorry.. @dgn already approved it. @jwendell ?

jwendell (approved):

PR #861, opened 2021-09-28
MAISTRA-2686 Clear configuredMembers when SMCP is deleted

luksa:


pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -242,6 +242,10 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
 		return reconcile.Result{}, err
 	} else if mesh == nil {
```
dgn, line 243: not your change, but: indentation looks strange here?

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -242,6 +242,10 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
...
 		log.Info("Skipping reconciliation of SMMR", "project", meshNamespace, "reason", reason, "message", message)
+		if reason == maistrav1.ConditionReasonSMCPMissing {
+			roll.Status.PendingMembers = sets.NewString(roll.Status.Members...).Difference(sets.NewString(roll.Status.TerminatingMembers...)).List()
+			roll.Status.ConfiguredMembers = []string{}
+		}
 		setReadyCondition(roll, false, reason, message)
 		// when a control plane is created/updated/deleted our watch will pick it up and issue a new reconcile event
 		return reconcile.Result{}, r.updateStatus(ctx, roll)
```
dgn, line 251: will we also have to delete any remaining ServiceMeshMembers? Or only when they are removed from `.members`?

luksa: The ServiceMeshMembers don't need to be removed. 

This is just a quick fix that I copied over from 2.0 (I removed it during the redesign of the SMMR/SMM controller as it didn't look right, but turns out that removing this code breaks the subsequent SMCP deployment). The correct fix is to have the SMM controller de-configure the member namespace whenever the SMCP is deleted. But this is too big of a change to squeeze into 2.1, so it needs to be fixed later.

pkg/controller/servicemesh/memberroll/controller.go
luksa: Why? Looks ok to me.

rcernich (approved):

pkg/controller/servicemesh/memberroll/controller.go
dgn: nevermind it's my github side-by-side that makes it look broken

dgn (approved):

PR #862, opened 2021-09-28
MAISTRA-2688 Ensure pruner never deletes RoleBindings in member namespaces

luksa:
These RoleBindings are created by the SMM controller (by copying the RoleBindings from the mesh namespace). Thus, they should only be deleted by the SMM controller.


rcernich:
Do we have similar issues with NetworkPolicy and NetworkAttachementDefinition resources?

dgn (approved):

luksa:
Ugh, @rcernich you're right. The same issue exists for NetworkPolicies, because their labels are also copied from the NetworkPolicy in the mesh namespace. For NetworkAttachmentDefinitions, there's no issue because we create the object from scratch. 

I've now updated the PR so that the NetworkPolicy labels are removed.

luksa:
@dgn Can you review the additional commit?

rcernich (approved):

PR #863, opened 2021-10-05
Bump the CSV `replaces` version

jwendell:


luksa:
Did we release 2.0.8 as 2.0.8-2? Does replaces version need to be 2.0.8-2 because of this?

rcernich (approved):

jwendell:
> Did we release 2.0.8 as 2.0.8-2? Does replaces version need to be 2.0.8-2 because of this?

I don't think so. We do this when we release things like 2.0.7.1 (dot, not dash).

rcernich (approved):

luksa:
We definitely released something as 2.0.8-2 (https://catalog.redhat.com/software/containers/openshift-service-mesh/istio-rhel8-operator/5d6ed9285a13461f5f020c15), but it looks like the operator version in OperatorHub is 2.0.8-0.

luksa (approved):

PR #865, opened 2021-11-10
MAISTRA-2172 Return error if 3scale is enabled in 2.1

luksa:


luksa:
Can't add unit test, because that would introduce an import cycle. We first need to reorganize the packages. The main problem is that `Version` returns the strategy used for validation and a bunch of other stuff. This needs to be uncoupled from the version.

pkg/controller/versions/strategy_v2_1.go
```diff
@@ -181,6 +182,17 @@ func (v *versionStrategyV2_1) validateMixerDisabled(spec *v2.ControlPlaneSpec, a
 	return allErrors
 }
 
+func (v *versionStrategyV2_1) validateAddons(spec *v2.ControlPlaneSpec, allErrors []error) []error {
+	if spec.Addons == nil {
+		return allErrors
+	}
+
+	if spec.Addons.ThreeScale != nil && spec.Addons.ThreeScale.Enabled != nil && *spec.Addons.ThreeScale.Enabled {
+		allErrors = append(allErrors, fmt.Errorf("support for 3scale has been removed in v2.1; please configure the 3scale WebAssembly adapter using a ServiceMeshExtension resource"))
```
jwendell: Should we also explicitly instruct the user to remove the 3scale block?

dgn: +1

luksa: Done.

dgn (approved):

jwendell (approved):

PR #866, opened 2021-11-12
OSSM-734: Remove completely initContainers from injection template

jwendell:
Because we *always* use CNI, we end up with a section like this
```yaml
  initContainers: null
```

And this causes a merge with a spec that has initContainers to assume we
want to remove the block entirely.

Removing this section entirely - we do not use it anyway - fixes the
problem.


jwendell:
I'll create a test for it in the maistra-test-suite repo.

jwendell:
https://github.com/maistra/maistra-test-tool/pull/263

luksa (approved):

dgn (approved):

PR #869, opened 2021-12-02
MAISTRA-2736: Option control for CNI installation

VikingPingvin:
This is part of the external control plane story, documented here (see last update): https://issues.redhat.com/projects/MAISTRA/issues/MAISTRA-2488

As per the recent agreements, we manage the control-plane and data-plane cluster components with different operator instances, to avoid a single operator talk to multiple K8S APIs. That is, on the control-plane side we only install discovery chart, meaning we do not need i.e. any CNI plugin installation on that cluster.

Right now the CNI installation is unconditionally happens on SMCP operations regardless of any static/dinamic configuration. The plan is to add a new annotation to the operator deployment to control this behavior.

### DIFF
--- a/pkg/controller/common/cni/cni.go
+++ b/pkg/controller/common/cni/cni.go
@@ -24,7 +24,7 @@ type Config struct {
 func InitConfig(m manager.Manager) (Config, error) {
 	config := Config{}
 
-	if common.Config.OLM.CNIDisabled {
+	if !common.Config.OLM.CNIEnabled {
 		config.Enabled = false
 		return config, nil
 	}
--- a/pkg/controller/common/cni/cni.go
+++ b/pkg/controller/common/cni/cni.go
@@ -24,9 +24,14 @@ type Config struct {
 func InitConfig(m manager.Manager) (Config, error) {
 	config := Config{}
 
+	log := logf.Log.WithName("controller_init")
+
 	if !common.Config.OLM.CNIEnabled {
 		config.Enabled = false
+		log.Info(fmt.Sprintf("CNI is disabled for this installation: %v", config.Enabled))
 		return config, nil
+	} else {
+		log.Info(fmt.Sprintf("CNI is enabled for this installation: %v", config.Enabled))
 	}
 	_, err := m.GetRESTMapper().ResourcesFor(schema.GroupVersionResource{
 		Group:    "k8s.cni.cncf.io",
@@ -58,9 +63,6 @@ func InitConfig(m manager.Manager) (Config, error) {
 	} else if !meta.IsNoMatchError(err) {
 		return config, err
 	}
-
-	log := logf.Log.WithName("controller_init")
-	log.Info(fmt.Sprintf("CNI is enabled for this installation: %v", config.Enabled))
 
 	return config, nil
 }
--- a/pkg/controller/common/cni/cni.go
+++ b/pkg/controller/common/cni/cni.go
@@ -24,6 +24,10 @@ type Config struct {
 func InitConfig(m manager.Manager) (Config, error) {
 	config := Config{}
 
+	if common.Config.OLM.CNIDisabled {
+		config.Enabled = false
+		return config, nil
+	}
 	_, err := m.GetRESTMapper().ResourcesFor(schema.GroupVersionResource{
 		Group:    "k8s.cni.cncf.io",
 		Version:  "v1",
--- a/pkg/controller/common/cni/cni_test.go
+++ b/pkg/controller/common/cni/cni_test.go
@@ -16,6 +16,8 @@ func TestInitConfig_disablingCNI(t *testing.T) {
 	assert.Equals(err, nil, "", t)
 	assert.Equals(config.Enabled, false, "", t)
 
+	// Quick test cleanup
+	common.Config.OLM.CNIEnabled = true
 }
 
 func InitializeGlobals(operatorNamespace string) func() {
--- a/pkg/controller/common/cni/cni_test.go
+++ b/pkg/controller/common/cni/cni_test.go
@@ -1,0 +1,26 @@
+package cni
+
+import (
+	"testing"
+
+	"github.com/maistra/istio-operator/pkg/controller/common"
+	"github.com/maistra/istio-operator/pkg/controller/common/test/assert"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+func TestInitConfig_disablingCNI(t *testing.T) {
+	operatorNamespace := "istio-operator"
+	InitializeGlobals(operatorNamespace)()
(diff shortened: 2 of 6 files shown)


pkg/controller/common/cni/cni.go
```diff
@@ -24,6 +24,10 @@ type Config struct {
 func InitConfig(m manager.Manager) (Config, error) {
 	config := Config{}
 
+	if !common.Config.OLM.CNIEnabled {
+		config.Enabled = false
```
luksa, line 30: Please ensure this code logs the fact that CNI is disabled.

pkg/controller/common/cni/cni_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"testing"
+
+	"github.com/maistra/istio-operator/pkg/controller/common"
+	"github.com/maistra/istio-operator/pkg/controller/common/test/assert"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+func TestInitConfig_disablingCNI(t *testing.T) {
```
rcernich, line 11: Can you add a test that ensures this is enabled by default?  I'm concerned that if this is not specified in the config that it will default to false.  @luksa, would you be okay if this was renamed `CNIDisabled`?

VikingPingvin:
I am not sure what to do regarding the failing unittests. Can you give me some pointers?

pkg/controller/common/cni/cni_test.go
luksa: I'd prefer to keep `CNIEnabled` to prevent double negatives, but I see your point. Is there a way we can make it default to `true`, like we can do with cmd-line arguments?

luksa:
@VikingPingvin You need to run `make gen` to fix the gen-check. The unit test fails because `TestInitConfig_disablingCNI` runs first. It sets `CNIEnabled` to `false`, but never resets it back to the previous value.

That being set, we should remove the global `Config` variable and find a better place for it.

pkg/controller/common/cni/cni_test.go
VikingPingvin: @luksa @rcernich 
Currently, I have defaulted it to `true` in an `init()` function, same way as for the `webhookConfiguration` switch.
https://github.com/maistra/istio-operator/pull/869/files#diff-2e3990f5984ac30e31d8e7167db0157324bd18330d8502da656e93cc620355fdL14-L16

libesz:
@luksa it is okay to leave the global `Config` as is now? Or you meant we should refactor this within the scope of this CNI related change?
Is there anything else for this PR?

VikingPingvin:
@luksa as @libesz said, should we consider the refactor, or could this PR be good as-is?

luksa:
@VikingPingvin @libesz We can fix the `Config` in another PR. 

I've added a few more comments.

pkg/controller/common/cni/cni_test.go
```diff
@@ -0,0 +1,32 @@
...
+	var m manager.Manager
+	config, err := InitConfig(m)
+	assert.Equals(err, nil, "", t)
+	assert.Equals(config.Enabled, false, "", t)
+
+	// Quick test cleanup
+	common.Config.OLM.CNIEnabled = true
+}
+
+func InitializeGlobals(operatorNamespace string) func() {
```
luksa: Why does this return a function instead of just doing the work?

pkg/controller/common/cni/cni_test.go
```diff
@@ -0,0 +1,32 @@
...
+func TestInitConfig_disablingCNI(t *testing.T) {
+	operatorNamespace := "istio-operator"
+	InitializeGlobals(operatorNamespace)()
+	var m manager.Manager
+	config, err := InitConfig(m)
+	assert.Equals(err, nil, "", t)
+	assert.Equals(config.Enabled, false, "", t)
+
+	// Quick test cleanup
+	common.Config.OLM.CNIEnabled = true
```
luksa: It would be safer to store the value of `CNIEnabled` in a local variable and reset it to that stored value instead of setting it to `true`. This way this test never has any side-effects.

pkg/controller/common/cni/cni_test.go
VikingPingvin: True, fixed.

pkg/controller/common/cni/cni_test.go
VikingPingvin: I don't think I completely understand.

The point of this test function would be to read the config value into the `common` struct, decoupling it into a local variable does not seem like a logical step to me.

Care to elaborate on this?

luksa: In this line here, you clean things up by setting `CNIEnabled` to `true`. But how do you know it had that value when you started the test? Right now, it's true that the value is `true`, but this could change in the future. If we ever change this to default to `false`, running this test will change that value and possibly affect any tests that run after this test. 

I know this is all hypothetical and it's just me nit-picking, but I've seen these sorts of things happen too often and they always require too much time to find the problem.

VikingPingvin: Gotcha, thanks for explaining that side. I totally agree. Fixing.

VikingPingvin:
@luksa With this latest commit, would your concers be resolved? Or you had a different idea there?

luksa (approved):

libesz:
@rcernich Can we merge this?

rcernich:
We'll need to hold off on merging until we tag the branch.  We just finished a release.

dgn:
FYI: the default branch has been updated to `maistra-2.2`. please make sure that this change lands on both `maistra-2.1` and `maistra-2.2`. you can use the `/cherry-pick <branch>` command

VikingPingvin:
Added as separate PR https://github.com/maistra/istio-operator/pull/875 @dgn

VikingPingvin:
Hi @dgn @rcernich .

Will these 2 branches do, or shall we do anything else?

VikingPingvin:
Hi @dgn @rcernich @luksa again, sorry for the repeated ping.
It would be great for us to know if we need to work on anything else regarding this 2 PRs, or what your general schedule is with this.

dgn (approved):

luksa:
@jwendell This PR targets the maistra-2.1 branch, but the builder image used for gen-check is 2.2. Any idea why?

jwendell:
> @jwendell This PR targets the maistra-2.1 branch, but the builder image used for gen-check is 2.2. Any idea why?

I **think** prow got confused when we changed the default branch... All jobs ran with the 2.2 image.

Maybe it's easier to just close this PR and open a brand new targeting 2.1 again? cc @VikingPingvin

jwendell:
It should be good now

PR #870, opened 2021-12-16
MAISTRA-2236: Make NetworkPolicy optional

jwendell:
This adds the boolean flag `spec.security.spec.security.manageNetworkPolicy` to
the SMCP, with the default value of `true`.

If it is true, current behavior is preserved, i.e., the operator creates
and manages the `NetworkPolicies` that allow communication between the
control plane and namespaces that are members of the mesh.

If it is false, the operator doesn't create such resources. The user is
entirely responsible for managing the `NetworkPolicies` and troubleshoot
any issue that might occur as a result of opting out of this feature.

### DIFF
--- a/manifests-maistra/2.1.0/servicemeshcontrolplanes.crd.yaml
+++ b/manifests-maistra/2.1.0/servicemeshcontrolplanes.crd.yaml
@@ -4984,6 +4984,8 @@ spec:
                       type:
                         type: string
                     type: object
+                  manageNetworkPolicy:
+                    type: boolean
                   trust:
                     properties:
                       additionalDomains:
@@ -9844,6 +9846,8 @@ spec:
                           type:
                             type: string
                         type: object
+                      manageNetworkPolicy:
+                        type: boolean
                       trust:
                         properties:
                           additionalDomains:
--- a/deploy/maistra-operator.yaml
+++ b/deploy/maistra-operator.yaml
@@ -4985,6 +4985,8 @@ spec:
                       type:
                         type: string
                     type: object
+                  manageNetworkPolicy:
+                    type: boolean
                   trust:
                     properties:
                       additionalDomains:
@@ -9845,6 +9847,8 @@ spec:
                           type:
                             type: string
                         type: object
+                      manageNetworkPolicy:
+                        type: boolean
                       trust:
                         properties:
                           additionalDomains:
--- a/deploy/servicemesh-operator.yaml
+++ b/deploy/servicemesh-operator.yaml
@@ -4985,6 +4985,8 @@ spec:
                       type:
                         type: string
                     type: object
+                  manageNetworkPolicy:
+                    type: boolean
                   trust:
                     properties:
                       additionalDomains:
@@ -9845,6 +9847,8 @@ spec:
                           type:
                             type: string
                         type: object
+                      manageNetworkPolicy:
+                        type: boolean
                       trust:
                         properties:
                           additionalDomains:
--- a/deploy/src/crd.yaml
+++ b/deploy/src/crd.yaml
@@ -4984,6 +4984,8 @@ spec:
                       type:
                         type: string
                     type: object
+                  manageNetworkPolicy:
+                    type: boolean
                   trust:
                     properties:
                       additionalDomains:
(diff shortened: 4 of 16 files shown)


manifests-maistra/2.1.0/servicemeshcontrolplanes.crd.yaml
```diff
@@ -4967,6 +4967,8 @@ spec:
                     type: object
                   dataPlane:
                     properties:
+                      automaticNetworkPolicy:
```
rcernich: nit, maybe `configureMeshNetworkPolicyRules` or enable, or install, or manage?

rcernich (commented):
This looks pretty good.  I wonder if you could check the new setting in reconcile and just skip creation of NetworkPolicy resources?  It would require no chart changes and would apply to all versions.  wdyt?  (Inspired by the short circuit code in the member roll controller.)

manifests-maistra/2.1.0/servicemeshcontrolplanes.crd.yaml
jwendell: Thanks. Naming things is hard.
@dgn @luksa Any [other] suggestion here?

Also, I've put this under `spec.security.dataPlane`. Does that sound good?

jwendell:
@rcernich Not sure I get it. The NetworkPolicy must be created in the control plane so that member namespaces can communicate with the control plane ns. This PR prevents this NP to be copied to the member namespaces upon user request. But I think the original NP in the control plane ns must not be touched.

About chart changes... Are you saying we don't need to make changes in globals and shouldn't care about conversion v1 <-> v2? I'm a bit confused :)

rcernich:
Hey @jwendell, I meant in the code where we process the objects produced by the charts, e.g. [here](https://github.com/maistra/istio-operator/blob/maistra-2.1/pkg/controller/common/helm/manifestprocessing.go#L133), you might do a check of the setting in smcp and skip processing if the object is of kind NetworkPolicy.

manifests-maistra/2.1.0/servicemeshcontrolplanes.crd.yaml
rcernich: That seems like the correct place to me.

dgn:
didn't we decide on a global on-off switch? shouldn't that mean we don't create any NetworkPolicy resources? I think that's clearer tbh

that would then also mean it should probably live straight under `spec.security` because it affects both control plane and data plane

jwendell:
> didn't we decide on a global on-off switch? shouldn't that mean we don't create any NetworkPolicy resources? I think that's clearer tbh
> 
> 
> 
> that would then also mean it should probably live straight under `spec.security` because it affects both control plane and data plane

I updated the patch. Please take a look. It is a global switch.

dgn:
> > didn't we decide on a global on-off switch? shouldn't that mean we don't create any NetworkPolicy resources? I think that's clearer tbh
> > that would then also mean it should probably live straight under `spec.security` because it affects both control plane and data plane
> 
> I updated the patch. Please take a look. It is a global switch.

Sorry, I didn't see you moved it to .spec.security already. Re: naming: I think `spec.security.manageNetworkPolicy` is good

This is missing a few places where we create NetworkPolicy resources:

```
$ rg NetworkPolicy resources/helm/v2.1/
resources/helm/v2.1/gateways/istio-ingress/templates/networkpolicy.yaml
3:kind: NetworkPolicy

resources/helm/v2.1/istio-control/istio-discovery/templates/networkpolicy.yaml
2:kind: NetworkPolicy

resources/helm/v2.1/istio-telemetry/kiali/templates/networkpolicy.yaml
3:kind: NetworkPolicy

resources/helm/v2.1/istio-telemetry/grafana/templates/networkpolicy.yaml
3:kind: NetworkPolicy

resources/helm/v2.1/istio-telemetry/tracing/templates/networkpolicy.yaml
4:kind: NetworkPolicy

resources/helm/v2.1/mesh-config/templates/networkpolicy.yaml
2:kind: NetworkPolicy
21:kind: NetworkPolicy

resources/helm/v2.1/istio-telemetry/prometheus/templates/networkpolicy.yaml
3:kind: NetworkPolicy
```

I think @rcernich's approach of filtering them in the operator before applying might make more sense, because it'll allow us to do this without adding to `patch-charts.sh`

rcernich (approved):
lgtm

dgn (approved):
lgtm, we just need to also apply this change in maistra/api (we really need to migrate operator to use that repository)

PR #876, opened 2022-01-26
MAISTRA-2742: added istiod-remote chart, more optional charts, split mode annotation

mtturcsik:
Currently this is not a final solution just a concept.

The PR contains:
- Readded the istiod-remote chart
- There are several charts which can not be disabled in the current version of Maistra. In a split configuration we need the ability to disable these charts too. I have created new SMCP options as a solution.
- There can be a situation in a split configuration when we need the same namespace name for an SMCP as the operator's but in a different cluster. Currently this is not possible because of the validation webhooks. In this PR we use an optional `SplitModeEnabled` annotation for this purpose. 

This PR should not change the previous default behaviour of Maistra


build/patch-charts.sh
```diff
@@ -546,3 +545,5 @@ moveEnvoyFiltersToMeshConfigChart
 copyGlobalValues
 # TODO: remove this hack once the image is updated to include workingDir
 hacks
+cp ${HELM_DIR}/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml resources/helm/v2.1/istiod-remote/templates/
```
dgn: could you encapsulate these two `cp`s in their own function with a comment?

pkg/controller/versions/strategy_v2_1.go
```diff
@@ -266,6 +271,46 @@ func (v *versionStrategyV2_1) Render(ctx context.Context, cr *common.ControllerR
...
+		} else {
+			err = spec.Istio.SetField("gateways.enabled", true)
+		}
+	}
+
+	if smcp.Spec.MeshConfig != nil {
+		if smcp.Spec.MeshConfig.Enabled != nil {
+			err = spec.Istio.SetField("meshConfig.enabled", *smcp.Spec.MeshConfig.Enabled)
+		} else {
+			err = spec.Istio.SetField("meshConfig.enabled", true)
```
dgn: all these assignments to `err` are missing the corresponding `if err != nil` checks

pkg/controller/versions/strategy_v2_1.go
```diff
@@ -74,7 +74,7 @@ var (
 		},
 		MeshConfigChart: {
 			path:         "mesh-config",
-			enabledField: "",
+			enabledField: "meshConfig",
```
dgn: hmm. this is going to be weird, as we're writing out the contents of the `meshConfig` dict into the actual meshconfig configmap. so it will now contain an `enabled: true/false`. we should avoid that IMO

pkg/controller/versions/strategy_v2_1.go
```diff
@@ -266,6 +271,46 @@ func (v *versionStrategyV2_1) Render(ctx context.Context, cr *common.ControllerR
...
+
+	if smcp.Spec.Telemetry != nil {
+		if smcp.Spec.Telemetry.Enabled != nil {
+			err = spec.Istio.SetField("telemetry.common.enabled", *smcp.Spec.Telemetry.Enabled)
+		} else {
+			err = spec.Istio.SetField("telemetry.common.enabled", true)
+		}
+	}
+
+	if smcp.Spec.RemoteMode != nil {
```
dgn, line 302: won't you need to set `remotePilotAddress` in this case?

pkg/controller/versions/strategy_v2_1.go
```diff
@@ -30,7 +30,7 @@ var (
 	v2_1ChartMapping = map[string]chartRenderingDetails{
 		DiscoveryChart: {
 			path:         "istio-control/istio-discovery",
-			enabledField: "",
+			enabledField: "istioDiscovery",
```
dgn, line 33: why not `pilot.enabled`? isn't that the field upstream uses?

pkg/controller/servicemesh/webhooks/validation/controlplane.go
```diff
@@ -46,7 +46,7 @@ func (v *ControlPlaneValidator) Handle(ctx context.Context, req admission.Reques
 		return admission.Allowed("")
 	}
 
-	if req.Namespace == common.GetOperatorNamespace() {
+	if !common.Config.OLM.SplitModeEnabled && req.Namespace == common.GetOperatorNamespace() {
```
dgn, line 49: I'm not sure about the `SplitModeEnabled` field. Can't we detect whether a request is coming from a remote cluster and decide based on that?

dgn (commented):
First-pass review, added a few comments. What I'd like to discuss further are the API changes (to SMCP). I think it shouldn't be separate fields (`controlPlaneMode.enabled` vs `remoteMode.enabled`) but rather one field (e.g. `spec.topology = SINGLE_CLUSTER/DATA_PLANE_CLUSTER/CONTROL_PLANE_CLUSTER`). @luksa @rcernich wdyt?

pkg/controller/versions/strategy_v2_1.go
mtturcsik: We didn't need that for our setup but from the chart install enablement perspective it is not relevant

pkg/controller/versions/strategy_v2_1.go
mtturcsik: We can use that if it seems more convenient to others. I chose this because it is the same as the chart's name. I wanted to express that this is for the whole chart and not just for the pilot deployment

pkg/controller/servicemesh/webhooks/validation/controlplane.go
mtturcsik: I think we can use the `!ControlPlaneMode` here because that should be the only case when we want this to be turned off
If it is okay to you, I will change this to that parameter

pkg/controller/versions/strategy_v2_1.go
mtturcsik: I have removed the option to turn of this chart as it is not necessary to turn this off in different setups

pkg/controller/versions/strategy_v2_1.go
mtturcsik: I have fixed it

pkg/controller/servicemesh/webhooks/validation/controlplane.go
mtturcsik: > I think we can use the `!ControlPlaneMode` here because that should be the only case when we want this to be turned off If it is okay to you, I will change this to that parameter

I have tried it but it won't work because the `ControlPlaneMode` setting is on SMCP level.
The easiest way to adjust this setting would be a configuration annotation like the current approach

manifests-maistra/2.1.1/maistraoperator.v2.1.1.clusterserviceversion.yaml
```diff
@@ -15,7 +15,7 @@ metadata:
       The Maistra Operator enables you to install, configure, and manage an instance of Maistra service mesh. Maistra is based on the open source Istio project.
 
     containerImage: quay.io/maistra/istio-ubi8-operator:2.1.1
-    createdAt: 2022-01-03T10:28:03EST
```
dgn: nit: let's not commit this change

manifests-servicemesh/2.1.1/servicemeshoperator.v2.1.1.clusterserviceversion.yaml
```diff
@@ -17,7 +17,7 @@ metadata:
       The OpenShift Service Mesh Operator enables you to install, configure, and manage an instance of Red Hat OpenShift Service Mesh. OpenShift Service Mesh is based on the open source Istio project.
 
     containerImage: registry.redhat.io/openshift-service-mesh/istio-rhel8-operator:2.1.1
-    createdAt: 2022-01-03T10:28:05EST
```
dgn: nit: same here

pkg/controller/versions/strategy_v2_1.go
```diff
@@ -266,6 +271,62 @@ func (v *versionStrategyV2_1) Render(ctx context.Context, cr *common.ControllerR
...
+		} else {
+			err = spec.Istio.SetField("istioDiscovery.enabled", true)
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	if smcp.Spec.Telemetry != nil {
+		if smcp.Spec.Telemetry.Enabled != nil {
```
dgn, line 289: why add the Enabled field to Telemetry? can't you just set `telemetry.common.enabled` to the value of  `smcp.Spec.ControlPlaneMode.Enabled` like above for istioDiscovery?

mtturcsik: We wanted the possibility to switch off all telemetry resources and this was not necessarily a split control plane related situation
Everything can be disabled from the telemetry stack except this chart. Maistra only needs it when at least one component is enabled from the telemetry addons.

libesz:
@dgn @luksa could you please advice on the failure of gen-check tests? Last time it was related to the build system if I remember correctly.

luksa:
@libesz Just run `make gen` and commit the changes. However, be aware that you'll have to do this again when #888 is merged.

luksa:
Actually, you may only need to rebase.

mtturcsik:
I ran the `make gen` again but the only diff was the `createdAt` field in the manifests. I didn't commit them because of an earlier comment

dgn:
> I ran the `make gen` again but the only diff was the `createdAt` field in the manifests. I didn't commit them because of an earlier comment

try `make clean gen`

mtturcsik:
> try make clean gen

Now I got some other files too but I think there is some bug in the script because it puts the results to `repoRootFolder/github.com/maistra/istio-operator` and `repoRootFolder/istio-operator`

libesz:
@dgn @luksa tests seem to be green now. Could we proceed with the merge please?

pkg/apis/maistra/v2/telemetry.go
```diff
@@ -2,6 +2,8 @@ package v2
 
 // TelemetryConfig for the mesh
 type TelemetryConfig struct {
+	Enablement `json:",inline"`
```
rcernich, line 5: This should be unnecessary.  To disable telemetry, set type to None.

pkg/apis/maistra/v2/servicemeshcontrolplane_types.go
```diff
@@ -155,6 +154,11 @@ type ControlPlaneSpec struct {
 	// Runtime configuration for pilot (and galley, etc., pre 2.0)
 	// +optional
 	Runtime *ControlPlaneRuntimeConfig `json:"runtime,omitempty"`
+
+	ControlPlaneMode *ControlPlaneModeConfig `json:"controlPlaneMode,omitempty""`
+
+	RemoteMode *RemoteModeConfig `json:"remoteMode,omitempty"`
+
```
rcernich, line 159: I don't understand why we need two different settings.  Wouldn't this be easier to configure using an enum?  Is it possible to have both modes enabled?  Also, this might be better placed under the `General` section.

pkg/controller/versions/strategy_v2_1.go
rcernich: See my comment above.  You should be able to disable telemetry by setting telemetry.type=None

resources/smcp-templates/v2.1/external
```diff
@@ -0,0 +1,107 @@
+apiVersion: maistra.io/v2
+kind: ServiceMeshControlPlane
+metadata:
+  name: external
+spec:
+  version: v2.1
+  gateways:
```
rcernich, line 7: Unless you are changing default behavior, the ideal thing to do here would be to reuse the other templates and apply your changes here.  You can refer to the servicemesh and maistra templates for examples.  For example:
```yaml
  profiles:
  - servicemesh
  gateways:
    enabled: false # I don't think anything else is required here, as this disables all gateway charts
  general:
    controlPlaneMode: External
  addons:
    kiali:
      enabled: false
    grafana:
      enabled: false
```

Also, generally speaking, if there are requirements for field settings based on the controlPlaneMode, then I would manage that logic in the render code.  This will provide a single place to maintain derived settings, and you won't have to worry about users overriding them by accident.

resources/smcp-templates/v2.1/external
```diff
@@ -0,0 +1,107 @@
...
+    enabled: false
+    type: Istiod
+
+  controlPlaneMode:
+    enabled: true
+
+  remoteMode:
+    enabled: false
+
+  cni:
```
rcernich, line 50: I don't believe this field exists in smcp.spec.

build/patch-charts.sh
```diff
@@ -531,6 +530,13 @@ function hacks() {
     }' ${HELM_DIR}/istio-control/istio-discovery/templates/deployment.yaml
 }
 
+# Copying the correct ValidatingWebhookConfiguration and MutatingWebhookConfiguration from the istio-discovery chart
+# to istiod-remote because of the differences from the upstream Istio
+function copyIstioDiscoveryWebhooksToRemoteChart() {
+  cp ${HELM_DIR}/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml resources/helm/v2.1/istiod-remote/templates/
+  cp ${HELM_DIR}/istio-control/istio-discovery/templates/mutatingwebhook.yaml resources/helm/v2.1/istiod-remote/templates/
```
luksa, line 458: Should this use dir `v2.2` now that the PR targets 2.2?

pkg/apis/maistra/v2/servicemeshcontrolplane_types.go
luksa: +1

Also: please add a comment documenting what each of these fields is. The comments added here will end up being the documentation shown when users run `oc explain smcp.spec`, so their importance is beyond the scope of the code itself.

pkg/controller/servicemesh/webhooks/validation/controlplane.go
luksa: So if split mode is enabled, you can deploy the service mesh in the operator namespace?

> I have tried it but it won't work because the ControlPlaneMode setting is on SMCP level. The easiest way to adjust this setting would be a configuration annotation like the current approach

But you can get the SMCP from the request, right?

pkg/controller/servicemesh/webhooks/validation/member.go
```diff
@@ -53,7 +53,8 @@ func (v *MemberValidator) Handle(ctx context.Context, req admission.Request) adm
 		return admission.Errored(http.StatusBadRequest, fmt.Errorf("ServiceMeshMember must be named %q", common.MemberName))
 	}
 
-	if smm.Namespace == common.GetOperatorNamespace() {
+	// TODO: Need switch for external/split control plane detection (Dataplane is in a different clustern than the operator)
+	if !common.Config.OLM.SplitModeEnabled && smm.Namespace == common.GetOperatorNamespace() {
```
luksa, line 57: Ah, this is where you can't/shouldn't get the SMCP and replace SplitModeEnabled with !SMCP.ControlPlaneMode.

pkg/controller/versions/strategy_v2_1.go
mtturcsik: The `telemetry-common` chart will be installed always:
https://github.com/maistra/istio-operator/blob/fdcaa2af68e1c1950391f747e5f5e6c519eeb038/pkg/controller/versions/strategy_v2_1.go#L43-L45

pkg/apis/maistra/v2/telemetry.go
mtturcsik: See my comment above, the `telemetry-common` chart will be installed always.

pkg/controller/servicemesh/webhooks/validation/member.go
mtturcsik: Yes, because this webhook is responsible for validating the SMCP

resources/smcp-templates/v2.1/external
mtturcsik: DId you mean only the cni parameter?

build/patch-charts.sh
mtturcsik: I think we need this in `v2.1` as well because we are already using it this way

resources/smcp-templates/v2.1/external
mtturcsik: I will remove this, thank you

resources/smcp-templates/v2.1/external
mtturcsik: Thank you, I think I will try to rewrite the template to this format.
In our case the user shouldn't reach the SMCP settings of the control plane side (when controlPlaneMode is enabled). However in the data plane logic we need to consider this

pkg/apis/maistra/v2/servicemeshcontrolplane_types.go
mtturcsik: I think we can use one SMCP parameter to switch between these two modes. I thought this would be more transparent  when somebody tries to find out what enables these specific charts. Also, it adds the possibility of adding more parameters in the future (in case if its needed).
It can be placed under the `General` section, and I will add the comments too.

pkg/controller/servicemesh/webhooks/validation/controlplane.go
mtturcsik: Yes, I think we can add a validator function for that and move the original and the requested logic there.

pkg/apis/maistra/v2/servicemeshcontrolplane_types.go
libesz: @rcernich I think control and data plane parts will never want to co-exist in a single cluster (maybe in testing scenarios, but it is not a use-case for us), because that is the original deployment model which makes this change set completely unnecessary.
So do you suggest to implement a 3 way switch for this (`None`, meaning the original combined deployment, `ControlPlane` and `Remote`) similarly to how Telemetry works?

pkg/controller/versions/strategy_v2_1.go
libesz: @dgn @rcernich our insight is that the charts which has empty `enabledField` setting will always be rendered and deployed at least partially (based on how other values are set). Ref: https://github.com/mtturcsik/istio-operator/blob/7faff8fc6c1d005f35639ed4f38bc5f13ab26061/pkg/controller/versions/strategy_v2_1.go#L471

Isn't this the intended behavior?

pkg/apis/maistra/v2/servicemeshcontrolplane_types.go
rcernich: Basically, yes.  @luksa?

libesz: @luksa any thoughts or concerns please?

dgn:
I spoke with @luksa and we agreed that we'd like to see the following changes implemented:

- remove the SMCP API changes, only add fields to helm charts (accessible through techPreview)
- "telemetry.type: NONE" should disable the chart (by setting telemetry.enabled to false in helm values) instead of the SMCP spec change
- "pilot.enabled" should be the enabledField for the istio-discovery chart
- only make changes to the 2.2 charts (the current PR makes changes to 2.1 charts)

Once these are in, we can merge this ASAP. Sorry for the delay!

libesz:
The user project is not active anymore and so we are not able to test and maintain this change on our end.
We can close this for now.

PR #873, opened 2022-01-10
OSSM-796 Remove patching rules for resources from maistra.io and route.openshift.io

jewertow:
Patching rules for resources from maistra.io and route.openshift.io will no longer be needed once pull request [OSSM-796 add SMMR integration test](https://github.com/maistra/istio/pull/458) is merged.

PR #882, opened 2022-02-10
Remove patching clusterrole file with maistra specific roles

jewertow:
It's a cherry pick of #873.


jewertow:
I closed it, because we have to wait for maistra/istio 2.2 to be ready. I will review roles once rebase is completed.

PR #884, opened 2022-02-17
OSSM-303 | Make creating OpenShift Route for istio-ingressgateway optional

jewertow:
This change enables users to control creation of OpenShift Route for istio-ingressgateway deployment by setting the following config in the SMCP spec:
```
apiVersion: maistra.io/v2
kind: ServiceMeshControlPlane
metadata:
  name: basic
  namespace: istio-system
spec:
  version: v2.1
  gateways:
    ingress:
      routeConfig:
        enabled: true # or false
```
Setting `gateways.ingress.routeConfig` is optional, because creation of OpenShift Route is enabled by default.

Important:
- this change affects SMCP **v2.1**
- `routeConfig` **does not apply** to egress gateway nor additional gateways.

### DIFF
--- a/build/patch-charts.sh
+++ b/build/patch-charts.sh
@@ -243,14 +243,6 @@ function patchGateways() {
   # do not expose port 15012, support for mesh expansion will automatically add this port
   sed_wrap -i -e '/port: 15012/,+3d' ${HELM_DIR}/gateways/istio-ingress/values.yaml
 
-  # add route config
-  routeConfig='\
-\
-    # specifies whether to create a Route resource (route.openshift.io/v1) for the istio-ingressgateway deployment\
-    routeConfig:\
-      enabled: true'
-  sed_wrap -i -e "/runAsRoot: false/a $routeConfig" ${HELM_DIR}/gateways/istio-ingress/values.yaml
-
   # add tracer config
   tracerConfig='\
   # Configuration for each of the supported tracers\
--- a/build/patch-charts.sh
+++ b/build/patch-charts.sh
@@ -243,6 +243,13 @@ function patchGateways() {
   # do not expose port 15012, support for mesh expansion will automatically add this port
   sed_wrap -i -e '/port: 15012/,+3d' ${HELM_DIR}/gateways/istio-ingress/values.yaml
 
+  # add route config
+  routeConfig='\
+\
+    # specifies whether to create a Route resource (route.openshift.io/v1) for the istio-ingressgateway deployment\
+    route: true'
+  sed_wrap -i -e "/runAsRoot: false/a $routeConfig" ${HELM_DIR}/gateways/istio-ingress/values.yaml
+
   # add tracer config
   tracerConfig='\
   # Configuration for each of the supported tracers\
--- a/build/patch-charts.sh
+++ b/build/patch-charts.sh
@@ -247,7 +247,8 @@ function patchGateways() {
   routeConfig='\
 \
     # specifies whether to create a Route resource (route.openshift.io/v1) for the istio-ingressgateway deployment\
-    route: true'
+    routeConfig:\
+      enabled: true'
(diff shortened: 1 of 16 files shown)


pkg/apis/maistra/v2/gateways.go
```diff
@@ -112,6 +112,9 @@ type ClusterIngressGatewayConfig struct {
 	// MeshExpansionPorts define the port set used with multi-cluster/mesh expansion
 	// +optional
 	MeshExpansionPorts []corev1.ServicePort `json:"meshExpansionPorts,omitempty"`
+	// RouteEnabled specifies whether to create a Route (route.openshift.io/v1) for the istio-ingressgateway component
+	// +optional
+	RouteEnabled *bool `json:"route,omitempty"`
```
rcernich: I think a more complete solution might be to add an `Ingress` field of type `ComponentIngressConfig`.

jewertow: It's a good idea, but `ClusterIngressGatewayConfig` already has a field ```IngressEnabled *bool `json:"ingress,omitempty"```, so it would be confusing and we couldn't name it `ingress`.

rcernich: Agreed, and `ingress.ingress` is odd.  Perhaps then,
```go
    RouteConfig *ClusterIngressGatewayConfig `json:"routeConfig,omitempty"`
```

jewertow: Don't you mean ```RouteConfig *ComponentIngressConfig `json:"routeConfig,omitempty"` ```?

jewertow: I assumed that you suggest to use type `ComponentIngressConfig`, but to handle only field `enabled` for now. Am I right?

rcernich: Yes, and that type includes an `enabled` field.

rcernich: For reference, https://github.com/maistra/istio-operator/blob/maistra-2.2/pkg/apis/maistra/v2/runtime.go#L312

dgn: @rcernich I don't understand, why do we want to add this whole struct? The issue was about adding a bool enable/disable field, now we're adding a bunch of fields with no functionality?

jewertow: This approach enables us to make other route properties customizable in the future, e.g. configure TLS.

build/patch-charts.sh
```diff
@@ -243,6 +243,14 @@ function patchGateways() {
...
   sed_wrap -i -e '/port: 15012/,+3d' ${HELM_DIR}/gateways/istio-ingress/values.yaml
 
+  # add route config
+  routeConfig='\
+\
+    # specifies whether to create a Route resource (route.openshift.io/v1) for the istio-ingressgateway deployment\
+    routeConfig:\
+      enabled: true'
+  sed_wrap -i -e "/runAsRoot: false/a $routeConfig" ${HELM_DIR}/gateways/istio-ingress/values.yaml
+
```
jewertow: Once this [pull request](https://github.com/maistra/istio/pull/469) is merged, we can revert this patch.

pkg/apis/maistra/v2/gateways.go
dgn: sure, I get why we would use a struct with `Enablement`, but why reuse an existing one and not support the fields? Can't we just have an empty struct, with just a bool and add fields if/when we actually implement them?

dgn (approved):

pkg/apis/maistra/v2/smcp_new.yaml
```diff
@@ -302,6 +302,8 @@ spec:
           port: 443
           targetPort: 8443
       meshExpansionPorts: [] # additional expansion ports. default expansion ports will be added automatically if multiCluster is configured
+      routeConfig:
```
jwendell: nit: add the field comment here as well

jwendell (approved):

PR #888, opened 2022-03-03
OSSM-1147 Add support for OSSM 2.2.0

luksa:


build/patch-charts.sh
```diff
@@ -169,20 +177,20 @@ function patchGalley() {
 \1  containerPort: 15017/' $deployment
 
   # always istiod
-  sed_wrap -i -e '/{{- if eq .Values.revision ""}}/,/{{- end }}/ { /istiod/! d }' $deployment
+#  sed_wrap -i -e '/{{- if eq .Values.revision ""}}/,/{{- end }}/ { /istiod/! d }' $deployment
 
   # multitenant
-  echo '
```
dgn, line 175: didn't we remove this as part of https://github.com/maistra/istio-operator/pull/873 ? 

actually, it looks like that change was never cherry-picked

rcernich (approved):
lgtm

build/patch-charts.sh
```diff
@@ -38,42 +38,47 @@ function patchTemplates() {
...
+#  sed_wrap -i -e '/^---/,$ d' \
+#      -e 's/name: istiod-{{.*$/name: istiod-{{ .Values.revision | default "default" }}-{{ .Release.Namespace }}/' \
+#      ${HELM_DIR}/istio-control/istio-discovery/templates/clusterrole.yaml
+#  mv ${HELM_DIR}/base/templates/clusterrole.yaml ${HELM_DIR}/istio-control/istio-discovery/templates/clusterrole.yaml
 
   # - move istiod specific templates to istio-discovery
-  mv ${HELM_DIR}/base/templates/endpoints.yaml ${HELM_DIR}/base/templates/services.yaml ${HELM_DIR}/base/templates/validatingwebhookconfiguration.yaml ${HELM_DIR}/istio-control/istio-discovery/templates/
+#  mv ${HELM_DIR}/base/templates/endpoints.yaml ${HELM_DIR}/base/templates/services.yaml ${HELM_DIR}/base/templates/validatingwebhookconfiguration.yaml ${HELM_DIR}/istio-control/istio-discovery/templates/
+
+  sed_wrap -i -e 's/"cluster.local"/{{ .Values.global.trustDomain | default "cluster.local" }}/g' ${HELM_DIR}/istio-control/istio-discovery/templates/configmap.yaml
```
dgn: nit: a comment would be good to understand why we need this (I'm guessing we should rather write the trustDomain directly into the meshconfig field instead of going through values.global in the future?)

build/patch-charts.sh
```diff
@@ -38,42 +38,47 @@ function patchTemplates() {
...
-      ${HELM_DIR}/base/templates/rolebinding.yaml \
-      ${HELM_DIR}/base/templates/clusterrolebinding.yaml \
-      ${HELM_DIR}/istio-control/istio-discovery/templates/deployment.yaml
-
-  sed_wrap -i -e '1,/^---/ d' \
-    ${HELM_DIR}/base/templates/serviceaccount.yaml \
-    ${HELM_DIR}/base/templates/clusterrolebinding.yaml
+#  sed_wrap -i -e 's/\(: istiod-service-account\)/\1-{{ .Values.revision | default "default" }}/' \
+#      ${HELM_DIR}/base/templates/clusterrolebinding.yaml
+#  sed_wrap -i -e 's/\(: istiod-service-account\)/\1-{{ .Values.revision | default "default" }}/' \
```
dgn: it looks like none of these changes will be required anymore - should we just remove them entirely instead of commenting?

dgn (approved):
lgtm, but I think all the commented out `sed` calls should just be removed

build/patch-charts.sh
luksa: Yeah, I Told @jewertow not to cherry-pick it, because I knew this would be in a different place in 2.2.

build/patch-charts.sh
luksa: Done.

build/patch-charts.sh
luksa: Done.

PR #889, opened 2022-03-03
Run make gen

luksa:


jwendell:
https://github.com/maistra/istio-operator/pull/886 - we need to fix the auto-merge thing

PR #903, opened 2022-03-25
OSSM-1279 Run Maistra on Kubernetes

luksa:


deploy/examples/maistra_v2_servicemeshcontrolplane_cr_minimal_kubernetes.yaml
```diff
@@ -0,0 +1,44 @@
...
+spec:
+  techPreview:
+    gateways:
+      # To allow gateways to run in Kubernetes, we must set the UID/GUID
+      # instead of leaving it blank (in OpenShift, these values are initialized
+      # automatically; in Kubernetes, leaving them blank results in runAsUer
+      # being 0, which causes Pod creation to fail because the containers are
+      # configured with runAsNonRoot: true).
+      istio-ingressgateway:
+        runAsUser: 1337
```
rcernich: Is this something we could automate?  For example, by either modifying the default user in the image or by setting explicitly, similar to multus config?

rcernich: Or, just something in the operator's install config: platform=[k8s, ocp].

luksa: Yeah, setting the UID/GID in the image seems like the best solution. I've tested it and it works.

luksa: BTW: I've set the UID/GID to 1337 in the proxyv2 Dockerfile.

dgn: I guess we should move this to globals.yaml instead of defining it here, right?

dgn: I guess we should move this to globals.yaml instead of defining it here, right?

rcernich: Can we push the Dockerfile change upstream too?

dgn: ah nevermind- OpenShift *initializes* it, I understood OpenShift *overwrites* it. In that case-- can we make the distinction in the operator and set the value there?

luksa: If the proxyv2 Dockerfile change is merged, then this custom config is no longer necessary. Note: I only submitted the Dockerfile change in the product image. Should I also add the same change to Maistra/istio? As for upstream, I don't know if they'll accept the change, since they hardcode the UID/GID to 1337 in the Helm charts.

rcernich: I would submit the change to both places.  It doesn't seem like setting it in the container would affect the charts one way or the other.

dgn: so.. if our community images set this in the Dockerfile, we're good and don't need these changes anymore? I think I'd prefer that over having to use a special profile

luksa: Updated Dockerfile in https://github.com/maistra/istio/pull/504. The product proxyv2 Dockerfile change PR is also waiting to be reviewed. 

I'll also submit this change upstream.

I'll remove this custom SMCP config. With the proxyv2 changes, any SMCP should be deployable on Kubernetes with no k8s-specific config.

luksa: The Dockerfile change is no longer required, as the operator now checks if the UID will be autoassigned. If not, it sets the UID to 1337 in the Helm chart..

rcernich (approved):

PR #909, opened 2022-04-08
OSSM-1228 Reduce differences between Maistra and upstream CNI

luksa:


luksa:
/hold

This includes the commit from https://github.com/maistra/istio-operator/pull/903, which must be merged first.

resources/helm/overlays/istio_cni/templates/clusterrolebinding.yaml
```diff
@@ -1,4 +1,4 @@
-{{ if .Values.enabled }}
+{{ if .Values.cni.enabled }}
```
jwendell, line 1: Do we still need an overlay file? Can't we just use upstream's from istio repo?

luksa: We do. We run multiple containers in the istio-cni-node Pods (one for each CNI version).

pkg/controller/common/cni/cni.go
```diff
@@ -33,15 +36,17 @@ func InitConfig(m manager.Manager) (Config, error) {
 	} else {
 		log.Info(fmt.Sprintf("CNI is enabled for this installation: %v", config.Enabled))
 	}
+
+	config.Enabled = true
```
jwendell: do we want to enable CNI even if multus is not in use?

luksa: Yes. Multus simply allows us to support multiple Istio CNI versions at the same time (and fixes the race conditions).

rcernich (approved):

pkg/controller/servicemesh/controlplane/integration_test.go
```diff
@@ -75,7 +75,7 @@ func TestBootstrapping(t *testing.T) {
 		operatorNamespace     = "istio-operator"
 		controlPlaneNamespace = "test"
 		smcpName              = "test"
-		cniDaemonSetName      = "istio-node"
+		cniDaemonSetName      = "istio-cni-node"
```
jwendell, line 78: that needs docs update here: https://docs.openshift.com/container-platform/4.10/service_mesh/v2x/removing-ossm.html

luksa: Thanks. Created https://issues.redhat.com/browse/OSSMDOC-574

PR #910, opened 2022-04-11
OSSM-218: requeue reconcilation when Jaeger or Grafana routes are empty

jewertow:

#### Background context
Users sometimes encounter that tracing is disabled in Kiali CR when SMCP has configured `Elastic` as storage type for Jaeger.

The cause of this bug is function `PatchAddons` which disables Grafana or Jaeger in Kiali CR if their routes are empty.
It doesn't handle the case where Grafana or Jaeger routes are temporarily unavailable. In such a case tracing will be disabled until the next reconciliation of SMCP. The solution is to requeue reconciliation of SMCP in case of getting empty routes.

#### TODO:
- [x] unit tests
- [x] exponential backoff

### DIFF
--- a/pkg/controller/common/reconcile_result.go
+++ b/pkg/controller/common/reconcile_result.go
@@ -1,0 +1,19 @@
+package common
+
+import (
+	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func Reconciled() (reconcile.Result, error) {
+	return reconcile.Result{}, nil
+}
+
+func RequeueWithError(err error) (reconcile.Result, error) {
+	return reconcile.Result{}, err
+}
+
+func RequeueAfter(time time.Duration) (reconcile.Result, error) {
+	return reconcile.Result{RequeueAfter: time}, nil
+}
--- a/pkg/controller/servicemesh/controlplane/addons.go
+++ b/pkg/controller/servicemesh/controlplane/addons.go
@@ -11,22 +11,23 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/maistra/istio-operator/pkg/apis/external"
 	kialiv1alpha1 "github.com/maistra/istio-operator/pkg/apis/external/kiali/v1alpha1"
-	v1 "github.com/maistra/istio-operator/pkg/apis/maistra/v1"
-	v2 "github.com/maistra/istio-operator/pkg/apis/maistra/v2"
+	"github.com/maistra/istio-operator/pkg/apis/maistra/v1"
+	"github.com/maistra/istio-operator/pkg/apis/maistra/v2"
 	"github.com/maistra/istio-operator/pkg/controller/common"
 )
 
-func (r *controlPlaneInstanceReconciler) PatchAddons(ctx context.Context) error {
+func (r *controlPlaneInstanceReconciler) PatchAddons(ctx context.Context, spec *v2.ControlPlaneSpec) (reconcile.Result, error) {
 	// so far, only need to patch kiali
-	return r.patchKiali(ctx)
+	return r.patchKiali(ctx, spec.IsGrafanaEnabled(), spec.IsJaegerEnabled())
 }
 
-func (r *controlPlaneInstanceReconciler) patchKiali(ctx context.Context) error {
+func (r *controlPlaneInstanceReconciler) patchKiali(ctx context.Context, grafanaEnabled, jaegerEnabled bool) (reconcile.Result, error) {
 	if r.Instance == nil || !r.Instance.Status.AppliedSpec.IsKialiEnabled() {
-		return nil
+		return common.Reconciled()
 	}
 
 	log := common.LogFromContext(ctx)
@@ -38,9 +39,9 @@ func (r *controlPlaneInstanceReconciler) patchKiali(ctx context.Context) error {
 	if err := r.Client.Get(ctx, types.NamespacedName{Name: kialiConfig.ResourceName(), Namespace: r.Instance.Namespace}, kiali); err != nil {
 		if errors.IsNotFound(err) || errors.IsGone(err) {
 			log.Error(nil, fmt.Sprintf("could not patch kiali CR, %s/%s does not exist", r.Instance.Namespace, kialiConfig.ResourceName()))
-			return nil
+			return common.Reconciled()
 		}
-		return err
+		return common.RequeueWithError(err)
 	}
 	log.Info("patching kiali CR", kiali.Kind, kiali.GetName())
 
@@ -61,40 +62,38 @@ func (r *controlPlaneInstanceReconciler) patchKiali(ctx context.Context) error {
 	// grafana
 	grafanaURL, err := r.grafanaURL(ctx, log)
 	if err != nil {
-		return err
+		return common.RequeueWithError(err)
 	} else if grafanaURL == "" {
 		// disable grafana
 		if err := updatedKiali.Spec.SetField("external_services.grafana.enabled", false); err != nil {
-			return fmt.Errorf("could not set external_services.grafana.enabled in kiali CR: %s", err)
+			return common.RequeueWithError(errorOnSettingValueInKialiCR("external_services.grafana.enabled", err))
 		}
 	} else {
 		// enable grafana and set URL
 		// XXX: should we also configure the in_cluster_url
 		if err := updatedKiali.Spec.SetField("external_services.grafana.url", grafanaURL); err != nil {
-			return fmt.Errorf("could not set external_services.grafana.url in kiali CR: %s", err)
+			return common.RequeueWithError(errorOnSettingValueInKialiCR("external_services.grafana.url", err))
 		}
 		if err := updatedKiali.Spec.SetField("external_services.grafana.enabled", true); err != nil {
-			return fmt.Errorf("could not set external_services.grafana.enabled in kiali CR: %s", err)
+			return common.RequeueWithError(errorOnSettingValueInKialiCR("external_services.grafana.enabled", err))
 		}
 	}
 
 	// jaeger
-	jaegerURL, err := r.jaegerURL(ctx, log)
-	if err != nil {
-		return nil
-	} else if jaegerURL == "" {
+	jaegerURL := r.jaegerURL(ctx, log)
+	if jaegerURL == "" {
 		// disable jaeger
 		if err := updatedKiali.Spec.SetField("external_services.tracing.enabled", false); err != nil {
-			return fmt.Errorf("could not set external_services.tracing.enabled in kiali CR: %s", err)
+			return common.RequeueWithError(errorOnSettingValueInKialiCR("external_services.tracing.enabled", err))
 		}
 	} else {
 		// enable jaeger and set URL
 		// XXX: should we also configure the in_cluster_url
 		if err := updatedKiali.Spec.SetField("external_services.tracing.url", jaegerURL); err != nil {
-			return fmt.Errorf("could not set external_services.tracing.url in kiali CR: %s", err)
+			return common.RequeueWithError(errorOnSettingValueInKialiCR("external_services.tracing.url", err))
 		}
 		if err := updatedKiali.Spec.SetField("external_services.tracing.enabled", true); err != nil {
-			return fmt.Errorf("could not set external_services.tracing.enabled in kiali CR: %s", err)
+			return common.RequeueWithError(errorOnSettingValueInKialiCR("external_services.tracing.enabled", err))
 		}
 	}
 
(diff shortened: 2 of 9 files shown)


pkg/controller/servicemesh/controlplane/addons.go
```diff
@@ -103,29 +105,34 @@ func (r *controlPlaneInstanceReconciler) patchKiali(ctx context.Context) error {
...
+			return reconcile.Result{}, nil
 		}
-		return err
+		return reconcile.Result{}, err
 	}
 
-	return nil
+	// SMCP should be reconciled if Grafana or Jaeger are enabled, but there are no routes for these services
+	if (grafanaEnabled && grafanaURL == "") || (jaegerEnabled && jaegerURL == "") {
+		return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
```
luksa: I'm not a fan of this approach. Have you tried watching Routes and triggering the reconcile of the relevant SMCP(s) instead? That would be the proper fix, but I understand that it could be less robust.

If we stick to the approach used here, I'd really like to see the fixed requeue time replaced with an exponential backoff, so that it takes less than 5s seconds to retry the reconcile (and takes much more time if it fails repeatedly).

jewertow: I didn't consider your approach, but it seems to be better, so I can change my PR.
To be sure I correctly understand your idea: do you suggest to add the following watches to control plane reconciler?
```
c.Watch(&source.Kind{Type: &routev1.Route{}}, enqueueRequestForSMCP, filterJaegerRoutePredicate)
c.Watch(&source.Kind{Type: &routev1.Route{}}, enqueueRequestForSMCP, filterGrafanaRoutePredicate)
```
The only downside I can see is that additional watches increase load on kube-apiserver, but on the other hand the approach with watches is consistent with the current implementation and seems to be clearer.

jewertow: I figured out that the approach with watches will not work, because Jaeger routes don't have labels `maistra.io/owner` and `maistra.io/owner-name`. Without these labels we can't determine which name and namespace to set in a reconciliation request.

jewertow: Could we add these labels to Jaeger resources? And if we can, could we add them to the previous versions as well?

jewertow:
I tested this solution manually and it works.

pkg/controller/servicemesh/controlplane/addons.go
luksa: Ah yes, because we're not the ones creating those resources. How about watching the Jaeger resource? Does it contain the information we need?

jewertow: I don't know, but I will take a look at this resource once again. Maybe it's updated on route creation.

jewertow: I looked at the Jaeger resource and it doesn't contain any useful information that we could rely on. What's more, this resource is not updated when services or routes are changed, so it doesn't make sense to watch it.

jewertow: I added exponential backoff to reconciliation results.

go.mod
```diff
@@ -40,6 +40,7 @@ require (
 	sigs.k8s.io/controller-runtime v0.6.3
 	sigs.k8s.io/controller-tools v0.4.1
 	sigs.k8s.io/yaml v1.2.0
+	github.com/cloudflare/backoff v0.0.0-20161212185259-647f3cdfc87a
```
jewertow: This library is available under [BSD-2-Clause License](https://github.com/cloudflare/backoff/blob/master/LICENSE). As far as I understand this licence, we can use it, but could also anyone else take a look at it?

luksa: You can use the `Backoff` in `k8s.io/apimachinery/pkg/util/wait`.

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -42,12 +44,17 @@ type controlPlaneInstanceReconciler struct {
...
 }
 
 // ensure controlPlaneInstanceReconciler implements ControlPlaneInstanceReconciler
 var _ ControlPlaneInstanceReconciler = &controlPlaneInstanceReconciler{}
 
 const (
+	// Reconciliation backoff settings
+	backoffMaxDuration        = 10 * time.Minute
+	backoffIntervalMultiplier = 5
+	backoffInterval           = backoffIntervalMultiplier * time.Second
```
luksa: You should probably use a much shorter initial interval like 10ms, and a maximum of 3 or 5 minutes (I forget what's typically used in k8s).

jewertow: I don't see a need to reconcile SMCP so frequently. It's not critical for SMCP to reconcile its addons within milliseconds and  in fact it wouldn't change user experience effectively.

jewertow: I reduced interval to 1 second and max duration to 3 minutes.

luksa: The exp. backoff applies only when an error or unexpected event occurs. These things are typically a result of a race condition, which is resolved within milliseconds. That's why most backoffs used in Kubernetes controllers use an initial delay of 10 or so millis. This ensures that the system is brought to the desired state as soon as possible. The exponential doubling of this delay ensures that if there's some other reason for the failure, the retries don't overload the system. 

By using a 1 second initial delay, you're unnecessarily prolonging the time it takes for the system to be configured properly. In some use-cases, the difference between several milliseconds and a second could be important. That's why I suggest using a shorter initial delay.

jewertow: But this PR aims to solve a different problem. We just have to wait for Jaeger route to be ready and it may take a long time, because Jaeger components (query and collector) wait for Elasticsearch which takes a long time to start up.
The problem was observed only for production template which deploys 3 replicas where each one needs 16Gi RAM.

go.mod
jewertow: I considered this package, but the `Backoff` returns a fixed interval. There is also the `BackoffManger` interface which supports exponential backoff, but with different semantics. It returns the following channel `<-chan time.Time`, so it's not useful in our case.

luksa: If you set the `Steps` and `Factor` fields properly, the interval increases every time:
```
	backoff := wait.Backoff{
		Duration: 10 * time.Millisecond,
		Factor:   2,
		Steps:    math.MaxInt32,
		Cap:      3 * time.Minute,
	}
	for i := 0; i < 20; i++ {
		fmt.Println(backoff.Step())
	}
```

The code above prints the following:
```
10ms
20ms
40ms
80ms
160ms
320ms
640ms
1.28s
2.56s
5.12s
10.24s
20.48s
40.96s
1m21.92s
2m43.84s
3m0s
3m0s
3m0s
```

pkg/controller/servicemesh/controlplane/reconciler.go
luksa: Hmm, if you're sure that it always takes a long time (with absolutely no exceptions), then I agree that using sub-second intervals makes no sense. However, using intervals over one or two minutes also doesn't sound good. We don't want the users to wait more than necessary. 

The correct approach would be to watch the resources and do the reconcile as soon as the routes are updated, with the exponential backoff serving as backup if the watch events are missed. Is there really no way to watch the routes? If there isn't, then I suggest we found an appropriate backoff config that ensures that the reconciliation happens well within a minute. Maybe start with a 1s interval and allow it to go up to 10 or 15 seconds?

go.mod
jewertow: It seems that I misunderstood it, thanks. I will remove imported library and replace with it.

pkg/controller/servicemesh/controlplane/reconciler.go
jewertow: > if you're sure that it always takes a long time (with absolutely no exceptions)

No, I'm not sure, because we have no precise data. As I already said, the problem was observed only when Jaeger was configured to use Elasticsearch from production template. I tested this configuration and I observed that it takes some time for the Elasticsearch to be ready and then also Jaeger services must become ready to create Jaeger route. You have to take into account that images must be downloaded and deployments must become healthy and ready and it definitely doesn't take milliseconds. So I just don't think that it makes sense to reconcile SMCP so frequently.

jewertow: > Is there really no way to watch the routes?

As I said, I haven't found a way to do that.

jewertow: @luksa I reduced backoff interval as you suggested.

go.mod
jewertow: Done.

pkg/controller/servicemesh/controlplane/addons_test.go
```diff
@@ -359,3 +374,290 @@ func VerifyKialiUpdate(jaegerName, domain string, values *maistrav1.HelmValues)
...
+			objects: []runtime.Object{
+				kiali,
+				htpasswd,
+				grafanaRoute,
+				jaegerRoute,
+			},
+			expectedReconciliationResult: reconcile.Result{},
+		},
+		{
+			name:           "reconciliation should succeed when jaeger is disabled and there is no its route",
```
dgn: nit `and its route does not exist`

pkg/controller/servicemesh/controlplane/addons_test.go
```diff
@@ -359,3 +374,290 @@ func VerifyKialiUpdate(jaegerName, domain string, values *maistrav1.HelmValues)
...
+			jaegerEnabled:  false,
+			objects: []runtime.Object{
+				kiali,
+				htpasswd,
+				grafanaRoute,
+			},
+			expectedReconciliationResult: reconcile.Result{},
+		},
+		{
+			name:           "reconciliation should succeed when grafana is disabled and there is no its route",
```
dgn: same as above

pkg/controller/common/backoff.go
```diff
@@ -0,0 +1,43 @@
+package common
```
dgn: not 100% sure about this file being needed. If we want to keep it, I think it should be used everywhere we use `wait.Backoff` (I found two other places while quickly grepping through the code), and then it should probably use the same values everywhere, except where there are good reasons to deviate from those defaults. if it's only used in this PR, let's throw it out because it's really just a very thin wrapper around `wait.Backoff`

dgn (approved):
LGTM, only real comment is about the new `backoff.go`- I think if we want to keep that, we should use it everywhere as a sort of "wait.backoff with istio-operator defaults".

Very happy about the extensive tests! Good job Jacek

rcernich:
If I understand correctly, there is a single Backoff for the reconciler, which means that this will apply to all resources, i.e. the backoff is not resource specific.  Given that, would it be simpler to just always return a static value that's maybe long enough to ensure the resources exists and also long enough to not cause excessive load, while being short enough such that the fields are set correctly within a reasonable amount of time?  For example, what if we just returned 5min always?  We could also make it configurable, with the default being 5min so that it's easier to test, and users could modify the setting if they absolutely needed to.  Opinions?

jewertow:
Alternative solutions are:
- don't change anything, because SMCP is reconciled every 5 minutes, so we can just wait for the next reconciliation;
- reconcile SMCP immediately returning `reconcile.Result{Requeue: true}`;
- reconcile SMCP after static timeout, but much shorter than 5 minutes, e.g. 1m or 30s.

rcernich:
After reviewing the controller-runtime code, it appears that the default refresh interval is actually 10hr, not 5min.

@dgn, @luksa, what is your opinion regarding no back off and simply using a fixed value, something between 1-5min?

I worry we're over thinking this.

pkg/controller/common/backoff.go
jewertow: You're right that we use `wait.Backoff` in the code base, but in a little bit different context.

The backoff use cases you are talking about are used just as arguments for function `wait.ExponentialBackoff` which uses this object under the hood and manage its state. On the other hand, the purpose of this wrapper is to calculate and keep backoff value that is not used in the scope of a single function call. I intentionally moved it out from `controlplane` package to hide its implementation details and exported only its essential functions (I wanted to hide locking). Additionally `common.Backoff` is concurrent-safe and might be used by multiple goroutines, and it's not relevant for the rest of uses cases, because those are used in the scope of `wait.ExponentialBackoff`.

pkg/controller/servicemesh/controlplane/addons_test.go
jewertow: Thanks, done.

pkg/controller/servicemesh/controlplane/addons_test.go
jewertow: Done.

pkg/controller/servicemesh/controlplane/addons.go
```diff
@@ -142,12 +150,11 @@ func (r *controlPlaneInstanceReconciler) grafanaURL(ctx context.Context, log log
 	return getURLForRoute(grafanaRoute), nil
 }
 
-func (r *controlPlaneInstanceReconciler) jaegerURL(ctx context.Context, log logr.Logger) (string, error) {
```
jewertow, line 145: This function was always returning nil error.

jewertow:
> If I understand correctly, there is a single Backoff for the reconciler, which means that this will apply to all resources, i.e. the backoff is not resource specific.

You're right, but it's not a problem in my opinion, because Grafana is much more reliable and it should almost always be ready immediately.

pkg/controller/common/reconcile_result.go
```diff
@@ -0,0 +1,19 @@
...
+	return reconcile.Result{}, nil
+}
+
+func RequeueWithError(err error) (reconcile.Result, error) {
+	return reconcile.Result{}, err
+}
+
+func RequeueWithTimeout(timeout time.Duration) (reconcile.Result, error) {
+	return reconcile.Result{RequeueAfter: timeout}, nil
+}
```
jewertow, line 19: Inspired by: https://cloud.redhat.com/blog/7-best-practices-for-writing-kubernetes-operators-an-sre-perspective

I didn't change existing code which returns `reconcile.Result{}` directly, because the PR would be huge and I didn't want to increase its scope.
If you don't like this, because of inconsistent code style I can revert it. I just thought that we can replace old code gradually along other changes.

jewertow:
I noticed that the backoff is not growing, so don't merge it yet.

Edit:
I think that the backoff is not growing, because it's always refreshed due to [this](https://github.com/maistra/istio-operator/blob/maistra-2.2/pkg/controller/servicemesh/controlplane/controller.go#L220):
```
key, reconciler := r.getOrCreateReconciler(instance)
defer r.deleteReconcilerIfFinished(key, reconciler)
```

I think it is not worth to fix this problem, because of the complexity. I will just remove backoff and return always 1 minute timeout.

jewertow:
Ok, I tested this change without backoff and now it works fine.

pkg/controller/common/reconcile_result.go
luksa: No, this is a great improvement. Thanks!

luksa (approved):

jewertow:
This PR must wait for 2.2 to be tagged.

PR #913, opened 2022-04-14
OSSM-1377: Add v2.2 to the list of versions

jwendell:


dgn (changes requested):
Until [OSSM-1148](https://issues.redhat.com/browse/OSSM-1148) is resolved, this change will have to be made in the containers-midstream repo

jwendell:
Closing if favor of a more robust fix

PR #915, opened 2022-04-22
OSSM-1400 Remove support for installing v1.0

dgn:
One of the more impactful changes here is that we no longer default to v1.0 when no version is set. Especially in tests this can lead to new behaviour, but also in conversion- e.g. we were using the same conversion algorithm for status.AppliedSpec conversion, which often had no version (yet), and so would now fail with not defaulting to v1.0.

Leaving this as WIP as I still want to look into the behaviour of this when you have an existing v1.0 control plane and then update to 2.2.0. My plan was for the reconciler to simply log an error and skip reconciliation, but I'm not sure yet if that'll end up spinning the operator into an infinite loop (as the MemberController is also parsing version).


luksa:
Would it be better to throw an error when the version isn't set instead of defaulting to the latest version?

dgn:
> Would it be better to throw an error when the version isn't set instead of defaulting to the latest version?

That's what happens now. It's just in the conversion that we use DefaultVersion, but that's only to know how to fill some fields, and it only applies to the case where AppliedSpec is empty (and with it spec.Version and would thus fail conversion)

jewertow:
I found in [pkg/apis/maistra/v1/servicemeshcontrolplane_types.go](https://github.com/maistra/istio-operator/blob/maistra-2.2/pkg/apis/maistra/v1/servicemeshcontrolplane_types.go#L112) the following comment:
```
// Existing ServiceMeshControlPlanes with an empty version are treated as
// having the version set to "v1.0"
// +optional
Version string `json:"version,omitempty"`
```
Not very important, but would be nice to update it.

jewertow:
Another comment to update: https://github.com/maistra/istio-operator/blob/maistra-2.2/pkg/controller/servicemesh/webhooks/mutation/controlplane.go#L59.

jewertow:
What about function [ValidateDowngrade](https://github.com/maistra/istio-operator/blob/maistra-2.2/pkg/controller/versions/strategy_v1_1.go#L152) in versionStrategyV1_1? It looks to be no longer needed.

pkg/controller/versions/strategy_v1_1.go
```diff
@@ -215,21 +215,6 @@ func (v *versionStrategyV1_1) ValidateDowngrade(ctx context.Context, cl client.C
 		})
```
jewertow: Shouldn't whole this function be removed?

rcernich: I believe that's part of the interface, so it needs to be there.  It should return an error, as 1.0 is not supported (i.e. you can't downgrade).

dgn: tbh what I never understood about these func is that they're called on downgrade *to* that version and also on downgrade *from* it. So when I'm downgrading 2.1 to 2.0, ValidateDowngrade is called for both, which is really confusing

pkg/apis/maistra/conversion/policy.go
```diff
@@ -203,7 +203,7 @@ func populatePolicyConfig(in *v1.HelmValues, out *v2.ControlPlaneSpec, version v
 		policyType = version.Strategy().GetPolicyType(in, mixerPolicyEnabled, mixerPolicyEnabledSet, remoteEnabled)
 	}
 
-	istiod := !(version == versions.V1_0 || version == versions.V1_1)
+	istiod := !(version == versions.V1_1)
```
jwendell: nit: `istiod := version != versions.V1_1`

pkg/apis/maistra/conversion/policy.go
```diff
@@ -9,15 +9,15 @@ import (
...
+// between v1.1 and v2.0
 
 func populatePolicyValues(in *v2.ControlPlaneSpec, values map[string]interface{}) error {
 	// Cluster settings
 	if in.Policy == nil {
 		return nil
 	}
 
-	istiod := !(in.Version == "" || in.Version == versions.V1_0.String() || in.Version == versions.V1_1.String())
+	istiod := !(in.Version == "" || in.Version == versions.V1_1.String())
```
jwendell: nit: `istiod := in.Version != versions.V1_1.String()`

pkg/apis/maistra/conversion/security.go
```diff
@@ -61,7 +61,7 @@ func populateSecurityValues(in *v2.ControlPlaneSpec, values map[string]interface
 			if istiod == nil {
 				break
 			}
-			hasIstiod := !(in.Version == versions.V1_0.String() || in.Version == versions.V1_1.String())
+			hasIstiod := !(in.Version == versions.V1_1.String())
```
jwendell: ditto

pkg/apis/maistra/conversion/security.go
```diff
@@ -115,7 +115,7 @@ func populateSecurityValues(in *v2.ControlPlaneSpec, values map[string]interface
 					break
 				}
 				componentRoot := "pilot"
-				if in.Version == "" || in.Version == versions.V1_0.String() || in.Version == versions.V1_1.String() {
+				if in.Version == "" || in.Version == versions.V1_1.String() {
```
jwendell: ditto

pkg/apis/maistra/conversion/telemetry.go
```diff
@@ -14,7 +14,7 @@ func populateTelemetryValues(in *v2.ControlPlaneSpec, values map[string]interfac
 		return nil
 	}
 
-	istiod := !(in.Version == "" || in.Version == versions.V1_0.String() || in.Version == versions.V1_1.String())
+	istiod := !(in.Version == "" || in.Version == versions.V1_1.String())
```
jwendell: ditto

pkg/controller/servicemesh/controlplane/controller_test.go
```diff
@@ -202,7 +203,9 @@ func newControlPlane() *maistrav2.ServiceMeshControlPlane {
 			Generation: 1,
 			UID:        controlPlaneUID,
 		},
-		Spec:   maistrav2.ControlPlaneSpec{},
+		Spec: maistrav2.ControlPlaneSpec{
+			Version: versions.V2_2.String(),
```
jwendell: perhaps `versions.DefaultVersion`?

pkg/controller/servicemesh/member/controller_test.go
```diff
@@ -460,7 +429,7 @@ func TestReconciliationOfTerminatingNamespace(t *testing.T) {
 				tc.configureMember(member)
 			}
 
-			controlPlane := markControlPlaneReconciled(newControlPlane(""), operatorVersionDefault)
+			controlPlane := markControlPlaneReconciled(newControlPlane("v2.2"), operatorVersionDefault)
```
jwendell: perhaps `versions.DefaultVersion`?

pkg/apis/maistra/conversion/policy.go
dgn: :+1:

pkg/apis/maistra/conversion/policy.go
dgn: not sure about this one. I don't want to break existing 1.0 SMCPs-- just not reconcile them anymore

pkg/apis/maistra/conversion/security.go
dgn: :+1:

pkg/apis/maistra/conversion/security.go
dgn: see above

pkg/controller/servicemesh/controlplane/controller_test.go
dgn: :+1:

pkg/controller/servicemesh/member/controller_test.go
dgn: :+1:

pkg/apis/maistra/conversion/policy.go
dgn: actually, I guess the fact that in other places we don't have the check for empty string means that it's fine

pkg/controller/versions/strategy_v1_1.go
dgn: nevermind, I was mistaking a `<` for a `<=`. This is only called on the downgrade *to* a version.

dgn: func is throwing an error now

dgn:
All comments addressed. I just need to find some time to run the upgrade test against an existing v1.0 control plane to be sure

jwendell (approved):

jwendell (commented):
You should remove 1.0 from `build/manifest-templates/clusterserviceversion.yaml`

pkg/apis/maistra/conversion/gateways.go
```diff
@@ -626,8 +626,6 @@ func expansionPortsForVersion(version string) ([]corev1.ServicePort, error) {
 	switch version {
 	case "":
 		fallthrough
-	case versions.V1_0.String():
-		fallthrough
```
luksa, line 630: The "" case now falls through to V1_1. Shouldn't it go to the default version?

dgn: good catch. I think it shouldn't fall through at all but throw an error. will fix

luksa (approved):
Everything looks good except the fallthrough.

dgn:
> You should remove 1.0 from `build/manifest-templates/clusterserviceversion.yaml`

thanks! good catch

dgn:
okay nevermind the testing- it is practically impossible these days to install a v1.0 control plane with the 2.1.2 operator. missing CRDs, other errors

PR #916, opened 2022-04-25
OSSM-1395 Reconcile ServiceMeshMember even if SMCP isn't fully reconciled

luksa:
In some scenarios there was a deadlock between the SMCP and SMM reconciliation. For example, if istiod in a new version of OSSM requires additional privileges to be present in member namespaces, it won't start until those privileges are added by the SMM controller. However, since the SMM controller waits for the SMCP to be fully reconciled, and the SMCP controller waits for istiod to be ready, the two controllers are in a deadlock. By allowing the SMM controller to configure the namespace regardless of the status of the SMCP, the deadlock can't occur.

### DIFF
--- a/pkg/controller/servicemesh/member/controller.go
+++ b/pkg/controller/servicemesh/member/controller.go
@@ -199,8 +199,8 @@ func (r *MemberReconciler) reconcileObject(ctx context.Context, member *maistrav
 	member.Status.SetAnnotation(statusAnnotationControlPlaneRef, member.Spec.ControlPlaneRef.String())
 
 	// 1. Fetch the referenced SMCP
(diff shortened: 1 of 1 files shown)


luksa:
/hold

It's possible that this change won't be needed once the fixes for OSSM-1420 are merged. Holding this PR for now. But this should go in to 2.3 along with some other changes that would allow fully parallelized reconciliation of SMCP/SMMR/SMM.

luksa:
/hold cancel

This should go in as it's unrelated to OSSM-1420.

jwendell (approved):

dgn (approved):

PR #918, opened 2022-04-28
OSSM-1332 Use pruneIndividually for CRs

luksa:
Previously, the pruner pruned CRs using the pruneAll() function, which deletes resources using the DELETE-COLLECTION operation. Apparently, the newer versions of OpenShift return a NoMatchError in this case, even though the resource type actually exists. This PR modifies the pruner to always delete resources individually.

### DIFF
--- a/pkg/controller/servicemesh/controlplane/pruner.go
+++ b/pkg/controller/servicemesh/controlplane/pruner.go
@@ -110,7 +110,7 @@ func (r *controlPlaneInstanceReconciler) findResourcesToPrune(ctx context.Contex
 					Version: version,
 					Kind:    crd.Spec.Names.Kind,
 				},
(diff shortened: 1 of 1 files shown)


rcernich (approved):

rcernich:
If it's non-performant, could we List() then delete the list?

dgn (approved):

dgn:
This definitely needs an acceptance test

luksa:
> If it's non-performant, could we List() then delete the list?

I don't think that's possible. Fortunately, what was causing the pruner to be slow in the past were the NoMatchErrors that caused the API discovery data to be evicted from the cache requiring the client had to refetch it from the API server repeatedly. I don't expect the switch to individual deletion to noticeably slow down the pruner.

PR #919, opened 2022-04-28
OSSM-1332 Use pruneIndividually for CRs [maistra-2.1]

luksa:
Previously, the pruner pruned CRs using the pruneAll() function, which deletes resources using the DELETE-COLLECTION operation. Apparently, the newer versions of OpenShift return a NoMatchError in this case, even though the resource type actually exists. This PR modifies the pruner to always delete resources individually.


luksa:
This is fixed only in 2.2.

PR #920, opened 2022-05-05
OSSM-1420 Remove cluster-scoped privileges

luksa:


dgn (approved):
lgtm, but we should also delete the istiod-reader ClusterRole and ClusterRoleBinding (even though they're applied to a non-existant ServiceAccount right now, but that could be exploited)

luksa:
Created https://issues.redhat.com/browse/OSSM-1465 to remove the two, as the point of OSSM-1420 was only to remove the cluster-scoped privileges that were added during the rebase.

PR #922, opened 2022-05-05
OSSM-218: watch Jaeger resource to patch Kiali on Jaeger route creation

jewertow:

This is an alternative solution for PR #910. I investigated and tested Jaeger operator once again and it turned out that Jaeger resource is updated on successful children resource creation.

### DIFF
--- a/pkg/controller/servicemesh/controlplane/controller.go
+++ b/pkg/controller/servicemesh/controlplane/controller.go
@@ -23,6 +23,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	jaegerv1 "github.com/maistra/istio-operator/pkg/apis/external/jaeger/v1"
 	"github.com/maistra/istio-operator/pkg/apis/maistra/status"
 	v2 "github.com/maistra/istio-operator/pkg/apis/maistra/v2"
 	"github.com/maistra/istio-operator/pkg/controller/common"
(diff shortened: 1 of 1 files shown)


pkg/controller/servicemesh/controlplane/controller.go
```diff
@@ -92,6 +93,10 @@ func add(mgr manager.Manager, r *ControlPlaneReconciler) error {
 	if err = c.Watch(&source.Kind{Type: &appsv1.DaemonSet{}}, enqueueRequestForSMCP, ownedResourcePredicates); err != nil {
 		return err
 	}
+	// Jaeger must be watched to patch Kiali on Jaeger route creation
```
jwendell, line 96: I think this will log errors if Jaeger CRD's are not installed. They are not mandatory unless Jaeger is enabled in SMCP. I believe a check if Jaeger is enabled should be enough.

jewertow: Good point. Thanks

jewertow: You were right. I tested it and I got the following error:
```
{
  "level": "error",
  "ts": 1651836015.8846438,
  "logger": "controller-runtime.source",
  "msg": "if kind is a CRD, it should be installed before calling Start","kind":"Jaeger.jaegertracing.io","error":"no matches for kind \"Jaeger\" in version \"jaegertracing.io/v1\""
}
```

> I believe a check if Jaeger is enabled should be enough.

There are two reasons why we can't do that:
1. We can't check in `ControlPlaneReconciler` if Jaeger is enabled, because this information belongs to a particular `controlPlaneInstanceReconciler`.
2. We could check if `Jaeger` CRD is installed, but it wouldn't work if Jaeger is installed after the ServiceMesh operator is installed and started.

jewertow: So it seems that we can't rely on watching `Jaeger` resource.

rcernich: I thought we already determined that we watching the jaeger resource wouldn't help because it didn't update its status.  I thought that in the case where we can't find the jaeger route we were just going to requeue with a fixed time (30s, 1m, something like that).

jewertow: Yes, but I investigated the code base of jaeger-operator to be sure that I'm right and it turned out that I wasn't. Jaeger operator updates its `resourceVersion` on successful creation of children resources (deployments, services, routes...). So I tried to watch Jaeger, but as you can read in the comments above, Jaeger CRD may not exist and then the controller will fail on watching.

PR #924, opened 2022-05-06
OSSM-1480 Ensure resource patch failures are treated as errors

luksa:


luksa:
Note: This fixes a regression. The operator did treat these types of failures as errors, but in one of my changes a year ago, I accidentally shadowed the `err` variable, causing the error to be ignored.

jwendell (approved):

PR #926, opened 2022-05-11
OSSM-1491 Ensure Routes are created on OpenShift, but not on Kubernetes

luksa:


pkg/controller/common/helm/manifestprocessing.go
```diff
@@ -206,11 +217,25 @@ func (p *ManifestProcessor) processObject(ctx context.Context, obj *unstructured
 	if err == nil {
 		log.V(2).Info("resource reconciliation complete")
 	} else {
-		log.Error(err, "error occurred reconciling resource")
+		if meta.IsNoMatchError(err) && isOpenShiftSpecificResource(obj) {
+			log.Info("resource kind not supported by cluster; operator likely not running in OpenShift, but in vanilla Kubernetes")
```
dgn, line 221: I'm guessing this is for backwards-compatibility when deploying older control plane versions?

dgn (approved):
LGTM

pkg/controller/common/helm/manifestprocessing.go
luksa: No, the previous solution was to use Helm's `.Capabilities.APIVersions`, but we don't initialize it at all. My first instinct was to initialize it properly by performing API discovery calls, but then we'd have to figure out how to cache it (when to invalidate the cache). The alternative approach in this PR feels less error-prone.

PR #927, opened 2022-05-18
OSSM-1535 Remove objectSelector from validation webhook

luksa:


jwendell (approved):

jwendell:
Can we have a test to verify the validation webhook is always triggered?

luksa:
> Can we have a test to verify the validation webhook is always triggered?

No.

It's possible that this type of test is in the istio repo already. We need to move our chart changes into the istio repo and run the tests there (instead of adding new tests to istio-operator).

jwendell:
Even if these tests are done in upstream, given we change the charts I think we should add them. If this is not possible in unit tests, we should do them in e2e tests. Let me comment on the ticket.

luksa:
I meant maistra/istio, not upstream. Since we make changes to the charts, we should move those changes to maistra/istio and run the istio integration tests with our charts.

PR #928, opened 2022-05-19
WIP: backport istio csr backport 2 1

SpectralHiss:
As mentioned by Jamie and Rob in #890, the release process requires us to backport new improvements such as the new cert-manager CA implementation to the last version of Maistra.
I was not aware of this until a few weeks back, it would be good to see the documentation on this, especially if it might have something to do with the TechnologyPreview field or something like that?
please clarify about that @rcernich 

This implementation was already tested for 2.1 before the 2.2 release, we've simply backported our code and tests, we will conduct a further e2e hence the WIP mark on PR.

### DIFF
--- a/build/patch-charts.sh
+++ b/build/patch-charts.sh
@@ -513,6 +513,39 @@ function copyGlobalValues() {
   cp ${SOURCE_DIR}/resources/helm/overlays/global.yaml ${SOURCE_DIR}/resources/helm/v2.1/
 }
 
+# This hack is hopefully only needed for a few versions until this PR is merged: https://github.com/istio/istio/pull/37264
+# It essentially modifies the chart to have the exact same changes
+function patchPilotServingCert() {
+  # add extra values
+  sed_wrap -i -e '/traceSampling:/ a\
+  extraArgs: []\
+  extraVolumeMounts: []\
+  extraVolumes: []' ${HELM_DIR}/istio-control/istio-discovery/values.yaml
+
+  # add extra volume in deployments (prepend before end of file)
+  sed_wrap -i -e '/^---$/ i\
+{{- if .Values.pilot.extraVolumes }}\
+{{ toYaml .Values.pilot.extraVolumes | indent 6 }}\
+{{- end }}' ${HELM_DIR}/istio-control/istio-discovery/templates/deployment.yaml
+
+
+
+  # add extra volume mounts (by prepending to volumesMounts: block)
+  sed_wrap -i -e '/volumeMounts:/ a\
+{{- if .Values.pilot.extraVolumeMounts }}\
+{{ toYaml .Values.pilot.extraVolumeMounts | indent 10 }}\
+{{- end }}' ${HELM_DIR}/istio-control/istio-discovery/templates/deployment.yaml
+
+  # Add extraArgs (by appending after discovery argument)
+  sed_wrap -i -e '/- "discovery"/ a\
+{{- if .Values.pilot.extraArgs }}\
+  {{-  range .Values.pilot.extraArgs }}\
+          - {{ . | quote }}\
+  {{- end }}\
+{{- end }}'  ${HELM_DIR}/istio-control/istio-discovery/templates/deployment.yaml
+
+}
+
 function hacks() {
   sed_wrap -i -e '/containers:/,/name: discovery/ {
       /name: discovery/a\
@@ -528,10 +561,8 @@ patchTemplates
 patchGalley
 patchGateways
 patchSidecarInjector
-#patchMixer # no longer present in 2.1
-#patchKialiTemplate # no longer present upstream
-#patchKialiOpenShift
 moveEnvoyFiltersToMeshConfigChart
 copyGlobalValues
+patchPilotServingCert
 # TODO: remove this hack once the image is updated to include workingDir
 hacks
--- a/pkg/apis/maistra/conversion/security.go
+++ b/pkg/apis/maistra/conversion/security.go
@@ -148,6 +148,86 @@ func populateSecurityValues(in *v2.ControlPlaneSpec, values map[string]interface
 			if err := setHelmStringValue(values, "global.caAddress", custom.Address); err != nil {
 				return err
 			}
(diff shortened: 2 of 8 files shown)


pkg/apis/maistra/conversion/security.go
```diff
@@ -148,6 +148,86 @@ func populateSecurityValues(in *v2.ControlPlaneSpec, values map[string]interface
...
+						"secretName": CertManagerConf.PilotCertSecretName,
+					},
+				},
+			}
+
+			if suppliedRootConfigMap {
+				extraVolumes = append(extraVolumes, map[string]interface{}{
+					"name": "ca-root-cert",
+					"configMap": map[string]interface{}{
+						"name": "istio-ca-root-cert",
```
dgn, line 216: shouldn't this be `CertManagerConf.RootCAConfigMapName`?

pkg/apis/maistra/v2/security.go
```diff
@@ -168,6 +171,14 @@ type CustomCertificateAuthorityConfig struct {
 	Address string `json:"address,omitempty"`
 }
 
+type CertManagerCertificateAuthorityConfig struct {
+	// Address is the grpc address for an Istio compatible certificate authority endpoint.
+	// .Values.global.caAddress
+	Address             string `json:"address,omitempty"`
+	PilotCertSecretName string `json:"pilotSecretName,omitempty"`
```
dgn, line 178: I'd prefer this to be just `SecretName` to match the field name in the Istio Gateway config

build/patch-charts.sh
```diff
@@ -513,6 +513,39 @@ function copyGlobalValues() {
   cp ${SOURCE_DIR}/resources/helm/overlays/global.yaml ${SOURCE_DIR}/resources/helm/v2.1/
 }
 
+# This hack is hopefully only needed for a few versions until this PR is merged: https://github.com/istio/istio/pull/37264
```
dgn, line 516: I would prefer for these changes to be made in the istio charts directly, and then cherry-picked to maistra/istio

pkg/apis/maistra/conversion/security.go
```diff
@@ -148,6 +148,86 @@ func populateSecurityValues(in *v2.ControlPlaneSpec, values map[string]interface
...
+				"--tlsKeyFile=/etc/cert-manager/tls/tls.key",
+				"--caCertFile=/etc/cert-manager/tls/ca.crt",
+			}
+
+			suppliedRootConfigMap:= (CertManagerConf.RootCAConfigMapName != "")
+
+			if suppliedRootConfigMap {
+				extraArgs[2] = "--caCertFile=/etc/cert-manager/ca/root-cert.pem"
+			} else {
+				fmt.Printf(`Warning! using istiod-tls ca.crt key as root of trust for mesh,\
```
dgn, line 178: this should probably use the logging facility. If we really want to warn the user though, then this should be moved into the ValidatingWebhook.

dgn (commented):
This LGTM in general, I added a few comments. I'd prefer the chart changes to go into maistra/istio (ideally a cherry-pick of an already-merged upstream PR), because we're trying to move away from the chart patching.

Thank you very much for the contribution, and apologies that it took so long for us to review!

build/patch-charts.sh
SpectralHiss: It has been a struggle to get any progress on this PR, I am now getting core security contributors reviewing it, as soon as I get that merged I will come back and report here.

SpectralHiss: https://github.com/istio/istio/pull/38845 is the PR currently active

pkg/apis/maistra/conversion/security.go
SpectralHiss: yes totally , the trouble is , it's a valid configuration so i don't know if we should completely block the installation..

pkg/apis/maistra/conversion/security.go
SpectralHiss: correct, will fix :+1:

build/patch-charts.sh
SpectralHiss: it looks like I will have to change extraArgs a little but that should be a straightforward change..

SpectralHiss:
> This LGTM in general, I added a few comments. I'd prefer the chart changes to go into maistra/istio (ideally a cherry-pick of an already-merged upstream PR), because we're trying to move away from the chart patching.
> 
> Thank you very much for the contribution, and apologies that it took so long for us to review!

Sorry for not coming back to you earlier on this, there are many developments on the upstream istio project about this, what do you think is the best path forward now? https://github.com/istio/istio/pull/39375

To summarise what happened, the istio folks essentially asked me to implement implicitly loading the pilot certs with optional: true volumes, so istio-csr doesn't even need to set the volumemounts anymore!! 
it would still need to set ca adress helm value etc.. so i think this would completely remove the need for all of this work :cry: 

At least the user experience would be a lot better for istio-csr so there's that on the bright side!

dgn:
> Sorry for not coming back to you earlier on this, there are many developments on the upstream istio project about this, what do you think is the best path forward now? [istio/istio#39375](https://github.com/istio/istio/pull/39375)
> 
> To summarise what happened, the istio folks essentially asked me to implement implicitly loading the pilot certs with optional: true volumes, so istio-csr doesn't even need to set the volumemounts anymore!! it would still need to set ca adress helm value etc.. so i think this would completely remove the need for all of this work cry
> 
> At least the user experience would be a lot better for istio-csr so there's that on the bright side!

That does sound like a good development! I think for us it's okay to wait until those optional-volume-mounts are merged upstream and then cherry-pick them into [maistra/istio](https://github.com/maistra/istio). Then we can add this field to the API.

Thanks again for working on this!

PS: I realize we're discussing on the 2.1 branch- this would much rather go onto 2.2, not sure if we'll back-port it right away. Let's move the discussion to [your other PR](https://github.com/maistra/istio-operator/pull/890)

jewertow:
I'm closing this PR, because we don't backport new features to old versions.

PR #936, opened 2022-06-15
OSSM-1107 Add jwksResolverCA field

yxun:
- This PR is related to OSSM-1107 and Istio PR 17176
- Add a new 2.1 SMCP spec field as
  .spec.security.jwksResolverCA

Signed-off-by: Yuanlin Xu <yuanlin.xu@redhat.com>

### DIFF
--- a/deploy/maistra-operator.yaml
+++ b/deploy/maistra-operator.yaml
@@ -4990,6 +4990,8 @@ spec:
                       type:
                         type: string
                     type: object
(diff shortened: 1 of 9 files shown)


jwendell (approved):

jewertow (approved):

rcernich:
Will this break upgrades or existing users?  This change was merged into 2.1, so not having this feature would be a regression.

PR #939, opened 2022-06-20
OSSM-1148 Synchronise CSV template with what we have in dist-git

dgn:
This turns the production CSV (servicemeshoperator CSV) into a template for use in CPaaS. Community CSV stays untouched.

Note that
- it now has variables for container images so they can be filled in by CPaaS during a build
- the `relatedImages` section was removed, this is necessary for the OSBS pipeline to replace image references automatically

One negative side-effect is that you cannot directly `kubectl apply` the Deployment anymore because of the unreplaced variables. If we need this deployment model we'll need to find a way around that (e.g. generate another Deployment with the values filled in)

### DIFF
--- a/deploy/src/deployment-servicemesh.yaml
+++ b/deploy/src/deployment-servicemesh.yaml
@@ -13,56 +13,57 @@ spec:
       labels:
         name: istio-operator
       annotations:
-        olm.relatedImage.v1_0.3scale-istio-adapter: registry.redhat.io/openshift-service-mesh/3scale-istio-adapter-rhel8:1.0.0
-        olm.relatedImage.v1_0.citadel: registry.redhat.io/openshift-service-mesh/citadel-rhel8:1.0.11
-        olm.relatedImage.v1_0.cni: registry.redhat.io/openshift-service-mesh/istio-cni-rhel8:1.0.11
-        olm.relatedImage.v1_0.galley: registry.redhat.io/openshift-service-mesh/galley-rhel8:1.0.11
-        olm.relatedImage.v1_0.grafana: registry.redhat.io/openshift-service-mesh/grafana-rhel8:1.0.11
-        olm.relatedImage.v1_0.mixer: registry.redhat.io/openshift-service-mesh/mixer-rhel8:1.0.11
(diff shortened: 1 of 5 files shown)


deploy/src/deployment-servicemesh.yaml
```diff
@@ -13,56 +13,57 @@ spec:
...
+        olm.relatedImage.v1_0.galley: registry-proxy.engineering.redhat.com/rh-osbs/openshift-service-mesh-galley-rhel8:${OSSM_10}
+        olm.relatedImage.v1_0.grafana: registry-proxy.engineering.redhat.com/rh-osbs/openshift-service-mesh-grafana-rhel8:${OSSM_10}
+        olm.relatedImage.v1_0.mixer: registry-proxy.engineering.redhat.com/rh-osbs/openshift-service-mesh-mixer-rhel8:${OSSM_10}
+        olm.relatedImage.v1_0.pilot: registry-proxy.engineering.redhat.com/rh-osbs/openshift-service-mesh-pilot-rhel8:${OSSM_10}
+        olm.relatedImage.v1_0.prometheus: registry-proxy.engineering.redhat.com/rh-osbs/openshift-service-mesh-prometheus-rhel8:${OSSM_10}
+        olm.relatedImage.v1_0.proxy-init: registry-proxy.engineering.redhat.com/rh-osbs/openshift-service-mesh-proxy-init-rhel7:${OSSM_10}
+        olm.relatedImage.v1_0.proxyv2: registry-proxy.engineering.redhat.com/rh-osbs/openshift-service-mesh-proxyv2-rhel8:${OSSM_10}
+        olm.relatedImage.v1_0.sidecar-injector: registry-proxy.engineering.redhat.com/rh-osbs/openshift-service-mesh-sidecar-injector-rhel8:${OSSM_10}
```
jewertow, line 27: Shouldn't we remove images for v1.0, since we [removed support for installing v1.0](https://github.com/maistra/istio-operator/pull/915)?

dgn: I think you're right- in this PR I'm just synchronizing the CSVs though. I can create a follow-up that removes them

jewertow (approved):

PR #946, opened 2022-07-07
OSSM-1805: Requeue patching Kiali if Kiali resource is not available

jewertow:
While testing #910, I observed that the operator skips patching Kiali if this resource does not exist, but it may occur temporarily, so I think we should requeue reconciliation to wait for Kiali and patch it properly.


luksa (approved):

jewertow:
@jwendell I guess you should cherry-pick #956 to maistra-2.3.

PR #955, opened 2022-07-20
OSSM-1775: Fix 3scale version

jwendell:


rcernich:
It looks like the changes are being applied to the 2.2.1 metadata, not the 2.2.2 (but I think we need to roll the version there too).

jwendell:
@rcernich fixed.

luksa (approved):

PR #974, opened 2022-08-25
OSSM-1928 Remove erroneous maistra-version label from chart overlays [maistra-2.0]

luksa:
This label gets added when the chart is copied from the overlay dir to the final dir.


jwendell (approved):

luksa:
gen-check will pass when #982 is merged

PR #976, opened 2022-08-31
OSSM-1951 Fix use of deprecated policy/v1beta1 PodDisruptionBudget [maistra-2.2]

luksa:


rcernich:
Do we need to do this for 2.0 and 2.1 too?

luksa:
Yup. See #977 and #978

jewertow (approved):

PR #977, opened 2022-08-31
OSSM-1951 Fix use of deprecated policy/v1beta1 PodDisruptionBudget [maistra-2.1]

luksa:


jewertow (approved):

rcernich:
I'm a bit late here, but did we have issues using `Capabilities` in the past (I think it was related to detecting whether or not `Route` resources were available)?

luksa:
Yup, you're right. I forgot about that issue. I've created [OSSM-1975](https://issues.redhat.com/browse/OSSM-1975) to ensure we pass the correct server version when rendering the charts.

PR #986, opened 2022-09-06
OSSM-1958 Update Kiali CR spec.version so it can run the latest builds of Kiali

jmazzitelli:
Also updates the Kiali CR ingress setting since it has slightly changed - the old way is deprecated.
see: https://issues.redhat.com/browse/OSSM-1958

(this PR follows the pattern from the last time we updated the Kiali CR here: https://github.com/maistra/istio-operator/pull/911)

### DIFF
--- a/resources/helm/overlays/istio-telemetry/kiali/templates/kiali-cr.yaml
+++ b/resources/helm/overlays/istio-telemetry/kiali/templates/kiali-cr.yaml
@@ -5,7 +5,7 @@ metadata:
   name: {{ .Values.kiali.resourceName | default "kiali" }}
   namespace: {{ .Release.Namespace }}
 spec:
-  version: "v1.48"
+  version: "default"
   installation_tag: "Kiali [{{ .Release.Namespace }}]"
   istio_namespace: "{{ .Release.Namespace }}"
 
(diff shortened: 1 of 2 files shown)


jmazzitelli:
When 2.3 goes into code freeze and we know which version of Kiali will be released with it, we will want to change that `default` to the actual Kiali version (e.g. `v1.57`). That will be done in a separate PR when the time is appropriate.

resources/helm/v2.2/istio-telemetry/kiali/templates/kiali-cr.yaml
```diff
@@ -28,7 +28,8 @@ spec:
 {{- end }}
 {{- if .Values.kiali.ingress }}
 {{- if .Values.kiali.ingress.enabled }}
-    ingress_enabled: {{ .Values.kiali.ingress.enabled }}
+    ingress:
```
jwendell: Did you update this file manually?

jmazzitelli: Yes. Is this `resources/helm/v2.2/istio-telemetry/kiali/templates/kiali-cr.yaml` auto-generated somehow?

jwendell: Yes, they are auto-generated. You just need to touch the file in the `overlay` directory and run `make gen`.

If you need to change the 2.2 resources as well, for the ingress stuff, you need to do that in the `maistra-2.2` branch. Once it's merged there, it will be copied to the 2.3 branch.

jmazzitelli: OK.  I won't mess with the 2.2 one then. It isn't important. This will still work. Going forward, the new overlay will use the new ingress setting.

jmazzitelli:
OK, `make gen` shows no additional changes for this PR (other than the "createdAt" field in the maistra operator CSV, which we don't care about so I will not include that here).

jwendell (approved):

PR #997, opened 2022-09-08
OSSM-1969 Remove duplicate maistra-control-plane label [maistra-2.3]

luksa:


resources/helm/overlays/istio-telemetry/grafana/templates/deployment.yaml
```diff
@@ -19,7 +19,6 @@ spec:
   template:
     metadata:
       labels:
-        maistra-control-plane: {{ .Release.Namespace }}
```
jwendell, line 22: Is this correct? It's not a duplicate.

jewertow: This label can be removed from this template, because it's patched in patch-charts script.

jewertow (approved):

resources/helm/overlays/istio-telemetry/grafana/templates/deployment.yaml
luksa: Yup. When we copy templates into the {{overlay/}} dir, we need to remove labels that are added by the patch-charts script when the template is copied to the target dir. This is one such label.

PR #1003, opened 2022-09-09
OSSM-1872 Deprecate v1.1 from istio-operator 2.3.0

yxun:
Signed-off-by: Yuanlin <yuanlin.xu@redhat.com>


deploy/maistra-operator.yaml
```diff
@@ -10653,31 +10653,6 @@ spec:
...
-        olm.relatedImage.v1_0.citadel: docker.io/maistra/citadel-ubi8:1.0.8
-        olm.relatedImage.v1_0.cni: docker.io/maistra/istio-cni-ubi8:1.0.8
-        olm.relatedImage.v1_0.galley: docker.io/maistra/galley-ubi8:1.0.8
-        olm.relatedImage.v1_0.grafana: docker.io/maistra/grafana-ubi8:1.0.8
-        olm.relatedImage.v1_0.mixer: docker.io/maistra/mixer-ubi8:1.0.8
-        olm.relatedImage.v1_0.pilot: docker.io/maistra/pilot-ubi8:1.0.8
-        olm.relatedImage.v1_0.prometheus: docker.io/maistra/prometheus-ubi8:1.0.8
-        olm.relatedImage.v1_0.proxy-init: docker.io/maistra/proxy-init-centos7:1.0.8
-        olm.relatedImage.v1_0.proxyv2: docker.io/maistra/proxyv2-ubi8:1.0.8
-        olm.relatedImage.v1_0.sidecar-injector: docker.io/maistra/sidecar-injector-ubi8:1.0.8
```
jewertow, line 10684: As far as I remember correctly, we didn't remove these images for 1.0 intentionally, so I guess we shouldn't remove them and the images for 1.1 as well.

jewertow: @jwendell could you confirm?

luksa:
We should also remove the 1.0 container from the CNI daemonset.

luksa:
Oh, and the 1.0 strategy.

yxun:
This PR is WIP. I converted it back for testing the new OpenShift CI jobs

yxun:
> 

@luksa I removed the CNI v1.1 daemonset in both overlay and 2.3 templates. Is that correct or should I not remove the overlay one ?

yxun:
> Oh, and the 1.0 strategy.

We don't have 1.0 strategy in maistra-2.3 branch now. However, we have 1.1 strategy.
When I tried to remove that 1.1 strategies, I got some test failures and controller versions need additional changes to align with the strategy removal. 

Do we need to remove the v1_1 strategy in 2.3.0 ?

yxun:
I removed some v1.1 related test cases in this PR. Let me know if we still need to keep some v1.1 test cases or I missed any other tests. Thanks,

deploy/maistra-operator.yaml
yxun: @jewertow I haven't found a mail or doc about keeping those annotations. Do we still need them ?

dgn:
I wonder if we can remove the SMCPv1 API at some point

deploy/maistra-operator.yaml
dgn: not sure it was intentional. let's remove both

luksa:
You should also probably update the min version here: https://github.com/Maistra/istio-operator/blob/78ef39ec9034ff0139f4d04e51152ae0e36d6280/pkg/version/version.go#L17

jwendell (approved):

luksa:
There are a few more references to v1.1. For example:
- https://github.com/Maistra/istio-operator/blob/15f0a4411687264914209cc9c8bd23abb8ef306d/pkg/controller/versions/versions.go#L56
- https://github.com/Maistra/istio-operator/blob/15f0a4411687264914209cc9c8bd23abb8ef306d/pkg/controller/versions/strategy_v1_1.go
- https://github.com/Maistra/istio-operator/blob/95bca82d023e293a569c650f48fd530d9694e28d/Makefile#L81-L115
- https://github.com/Maistra/istio-operator/blob/36327f8a0e3f832d1c3a02145de16e4aa4cd268c/pkg/bootstrap/cni.go#L48
- ...

I guess we can remove these later.

yxun:
This PR removes 1.1 tests and I added do not merge label to avoid causing other open PR conflicts. When other 2.3 PRs are merged, I can rebase this one later. Thanks.

resources/helm/overlays/istio_cni/templates/daemonset.yaml
```diff
@@ -56,51 +56,6 @@ spec:
       containers:
         # This container installs the Istio CNI binaries
         # and CNI network config file on each node.
-{{- if and (.Values.cni.supportedReleases) (has "v1.1" .Values.cni.supportedReleases) }}
```
yxun, line 59: Should we keep the v1.1 cni container in the template ?

luksa: I think you removed the v1.1 images, right?

yxun: right, I was thinking what would happen in the existing cluster.
I am clear now.

luksa (approved):

yxun:
> There are a few more references to v1.1. For example:
> 
>     * https://github.com/Maistra/istio-operator/blob/15f0a4411687264914209cc9c8bd23abb8ef306d/pkg/controller/versions/versions.go#L56
> 
>     * https://github.com/Maistra/istio-operator/blob/15f0a4411687264914209cc9c8bd23abb8ef306d/pkg/controller/versions/strategy_v1_1.go
> 
>     * https://github.com/Maistra/istio-operator/blob/95bca82d023e293a569c650f48fd530d9694e28d/Makefile#L81-L115
> 
>     * https://github.com/Maistra/istio-operator/blob/36327f8a0e3f832d1c3a02145de16e4aa4cd268c/pkg/bootstrap/cni.go#L48
> 
>     * ...
> 
> 
> I guess we can remove these later.

I updated this PR. The cni v1.1 configmap setting is removed and Makefile v1.1 targets are removed.

I left the Strategy V1_1. I couldn't deprecate that one until cleaning all maistra apis v1 code.
I created a task for that remaining work. We can remove those later. 
https://issues.redhat.com/browse/OSSM-2106

luksa (approved):

PR #1004, opened 2022-09-22
Fix common.GetOperatorNamespace() to work in a OCP env

jwendell:
This fixes an issue with the tests when run in openshift CI, as the
`k8sutil.GetOperatorNamespace()` always return something. Because
the tests rely on the POD_NAMESPACE variable they fail because this
env variable is never checked.
    
Swapping the order the check is made fixes the problem.

### DIFF
--- a/pkg/controller/common/util.go
+++ b/pkg/controller/common/util.go
@@ -130,13 +130,14 @@ var (
 func GetOperatorNamespace() string {
 	initOperatorNamespace.Do(func() {
 		var err error
+
+		// Check for this variable before calling k8sutil.GetOperatorNamespace()
+		// This variable is set in unit tests - and potentially in dev debugging sessions
+		if operatorNamespace = os.Getenv("POD_NAMESPACE"); operatorNamespace != "" {
+			return
+		}
+
 		if operatorNamespace, err = k8sutil.GetOperatorNamespace(); err != nil {
-			if err == k8sutil.ErrNoNamespace || err == k8sutil.ErrRunLocal {
-				// see if dev is manually specifying this during debugging
-				if operatorNamespace = os.Getenv("POD_NAMESPACE"); operatorNamespace != "" {
-					return
-				}
-			}
 			panic(err)
 		}
 	})


jwendell:
@dgn @luksa hey folks, what do you think about this change? Is it safe?

This fixes an issue with the tests when run in openshift CI, as the `k8sutil.GetOperatorNamespace()` always return something. Because the tests rely on the POD_NAMESPACE variable they fail because this env is never checked.

cc @yxun

yxun:
This change is related to our unit tests running in an OCP pod.
what happened is the following tests failed.
FAIL: TestBootstrapping
FAIL: TestControlPlaneNotAllowedInOperatorNamespace
FAIL: TestMemberInOperatorNamespaceIsRejected
FAIL	github.com/maistra/istio-operator/pkg/controller/servicemesh/webhooks/validation	

By default, a pod running on OCP has the following sa e.g.

```
sh-4.4$ cat /var/run/secrets/kubernetes.io/serviceaccount/namespace 

istio-operatorsh-4.4$  
```

And the k8sutil.GetOperatorNamespace() is returning value from that file
https://github.com/maistra/istio-operator/blob/maistra-2.3/vendor/github.com/operator-framework/operator-sdk/pkg/k8sutil/k8sutil.go#L68

The above tests are comparing operator namespace by calling a common.GetOperatorNamespace() method.
e.g. 
https://github.com/maistra/istio-operator/blob/maistra-2.3/pkg/controller/servicemesh/webhooks/validation/controlplane_test.go#L53
https://github.com/maistra/istio-operator/blob/maistra-2.3/pkg/controller/servicemesh/webhooks/validation/controlplane.go#L51

When we expect a test with operator ns "openshift-operators", that value is not set and we used a wrong ns value from the k8sutil.

Here is a debugging message I added in a test case, the expected operator namespace value should be the "openshift-operators",

```
sh-4.4$ go test -timeout 1h -mod=vendor ./pkg/controller/servicemesh/webhooks/validation/   
--- FAIL: TestControlPlaneNotAllowedInOperatorNamespace (0.00s)
    **controlplane_test.go:58: Debug common.GetOperatorNamespace. Got: istio-operator**
    controlplane_test.go:60: Expected validator to reject ServiceMeshControlPlane in operator's namespace
--- FAIL: TestMemberInOperatorNamespaceIsRejected (0.00s)
    member_test.go:57: Expected validator to reject creation of ServiceMeshMember in operator namespace
FAIL
FAIL    github.com/maistra/istio-operator/pkg/controller/servicemesh/webhooks/validation        0.189s
FAIL
sh-4.4$ 
```

We need an additional review of this change. Because this changes a common util. Is there a reason we have to get the value from k8sutil before the os.Getenv one ? In there any place other than the tests will be affected by this change ?

luksa:
I think this change is okay.

luksa (approved):

PR #1006, opened 2022-09-23
OSSM-1320 Add option to run OSSM in cluster-wide mode

luksa:


pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -151,20 +152,29 @@ func add(mgr manager.Manager, r *MemberRollReconciler) error {
...
+				NamespacedName: common.ToNamespacedName(&smmr),
+			})
+		}
 	}
 	return requests
 }
 
+func contains(members []string, ns string) bool {
+	for _, member := range members {
+		return member == ns
```
jwendell: Is this right? It always returns on first interaction

pkg/controller/versions/strategy_v2_3.go
```diff
@@ -224,6 +273,71 @@ func (v *versionStrategyV2_3) ValidateUpgrade(ctx context.Context, cl client.Cli
...
+	oldClusterScoped, err := isClusterScoped(oldSMCP)
+	if err != nil {
+		return err
+	}
+	newClusterScoped, err := isClusterScoped(newSMCP)
+	if err != nil {
+		return err
+	}
+	if oldClusterScoped != newClusterScoped {
+		return fmt.Errorf("field spec.techPreview.%s is immutable", v2.ControlPlaneModeKey)
```
jwendell: What does this error mean? Can it be reworded so that it's easier to understand?

pkg/controller/versions/strategy_v2_3.go
```diff
@@ -224,6 +273,71 @@ func (v *versionStrategyV2_3) ValidateUpgrade(ctx context.Context, cl client.Cli
...
+				Group:    "*",
+				Resource: "*",
+			},
+		},
+	}
+	err := cl.Create(ctx, sar)
+	if err != nil {
+		return false, err
+	}
+	return sar.Status.Allowed && !sar.Status.Denied, nil
```
jwendell, line 338: Is the status field guaranteed to be filled when Create() returns?

pkg/controller/servicemesh/memberroll/controller.go
luksa: oh, wow, what was I smoking here?

I've now fixed this.

pkg/controller/versions/strategy_v2_3.go
luksa: That's a standard k8s error. See https://www.datree.io/resources/kubernetes-error-codes-field-is-immutable

pkg/controller/versions/strategy_v2_3.go
luksa: Yes.

pkg/controller/versions/strategy_v2_3.go
jwendell: So, this means that users are not allowed to change this field ever? They must destroy a control plane and create a new one instead? If so, what do you think about being more explicit in the error message? Something like: `Changes to spec.techPreview.%s are not allowed. In order to change the scope of this control plane installation, you need to destroy this control plane and install a new one with the desired scope`

build/patch-charts.sh
```diff
@@ -143,7 +143,7 @@ function patchGalley() {
   - apiGroups: [""]\
     resources: ["pods/finalizers"]\
     verbs: ["update"]' "${HELM_DIR}/istio-control/istio-discovery/templates/clusterrole.yaml"
-  sed_wrap -i -e 's/, *"nodes"//' "${HELM_DIR}/istio-control/istio-discovery/templates/clusterrole.yaml"
+#  sed_wrap -i -e 's/, *"nodes"//' "${HELM_DIR}/istio-control/istio-discovery/templates/clusterrole.yaml"
```
jwendell: nit: remove this

pkg/apis/maistra/v2/servicemeshcontrolplane_types.go
```diff
@@ -7,6 +7,12 @@ import (
 	v1 "github.com/maistra/istio-operator/pkg/apis/maistra/v1"
 )
 
+const (
+	ControlPlaneModeKey                = "controlPlaneMode"
+	ControlPlaneModeValueClusterScoped = "ClusterScoped"
+	ControlPlaneModeValueMultiTenant   = "MultiTenant"
```
jwendell, line 13: If I understand it correctly we do not validate this field, right? We only check if it's set to `ClusterScoped` and if it's anything else we handle it as multitentant, right?

luksa: Correct. We'll add validation when we move the field out of `techPreview`.

jwendell (approved):

build/patch-charts.sh
```diff
@@ -155,10 +154,12 @@ function patchGalley() {
...
   convertClusterRoleBinding "${HELM_DIR}/istio-control/istio-discovery/templates/clusterrolebinding.yaml" ""
   sed_wrap -i -e '/- "discovery"/ a\
+{{- if not .Values.global.clusterScoped }}\
           - --memberRollName=default\
+          - --disableNodeAccess=true\
+{{- end}}\
           - --enableCRDScan=false\
-          - --enableIngressClassName=false\
-          - --disableNodeAccess=true' "${deployment}"
+          - --enableIngressClassName=false' "${deployment}"
```
dgn: same here, this can be removed for cluster-scoped mode, right?

build/patch-charts.sh
```diff
@@ -155,10 +154,12 @@ function patchGalley() {
 
   convertClusterRoleBinding "${HELM_DIR}/istio-control/istio-discovery/templates/clusterrolebinding.yaml" ""
   sed_wrap -i -e '/- "discovery"/ a\
+{{- if not .Values.global.clusterScoped }}\
           - --memberRollName=default\
+          - --disableNodeAccess=true\
+{{- end}}\
           - --enableCRDScan=false\
```
dgn, line 159: we don't need to disable this anymore in cluster-scoped mode, do we?

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -258,8 +270,22 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
...
+	if roll.Spec.IsClusterScoped() {
+		nsList := &corev1.NamespaceList{}
+		err = r.Client.List(ctx, nsList)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		memberNamespaces = []string{}
+		for _, ns := range nsList.Items {
+			if ns.Name != meshNamespace && !isExcludedNamespace(ns.Name) {
+				memberNamespaces = append(memberNamespaces, ns.Name)
```
dgn, line 288: won't this become very costly in large clusters? we're creating `Member` resources one at a time...

dgn (commented):
I understand I'm very late to the party on this, and I suspect we chose the Member-based approach to this to keep implementation effort down. Do we have a migration path away from it, though? With cluster-scoped deployments, the MemberRoll and Member controllers should be idle IMO. We're putting a lot of load on the API server for functionality we don't require - the ignoring of certain namespaces could be done in the istio controlplane/injection webhook itself.

How hard will it be to remove Member/MemberRoll controllers from the equation later on? I can see we're relying on them for a) avoiding injection into special namespaces and b) avoiding mixing of cluster-wide and multi-tenant control planes

build/patch-charts.sh
luksa: I wasn't sure, but I've now enabled this for cluster-wide installs.

build/patch-charts.sh
luksa: Done.

pkg/controller/servicemesh/memberroll/controller.go
luksa: Costly in what way?

luksa:
> I understand I'm very late to the party on this, and I suspect we chose the Member-based approach to this to keep implementation effort down. Do we have a migration path away from it, though? With cluster-scoped deployments, the MemberRoll and Member controllers should be idle IMO. We're putting a lot of load on the API server for functionality we don't require - the ignoring of certain namespaces could be done in the istio controlplane/injection webhook itself.

When I started working on this, I understood the main goal was only to improve the performance in large clusters. I kept the Member/MemberRoll because I felt that it's better to keep the "feel" of OSSM the same as it was. Also, we can't just get rid of the Member controller, since we need it to create NetworkAttachmentDefinitions (and potentially NetworkPolicies, etc.) in the member namespaces. Sure, we could modify the controller to do the same work without requiring a Member resource to be present, but what would be the point? The Member resource itself allows you to see if a namespace is part of the mesh and get the status of the member. And it's not really putting any load on the API server. Not sure what you meant.

> How hard will it be to remove Member/MemberRoll controllers from the equation later on? I can see we're relying on them for a) avoiding injection into special namespaces and b) avoiding mixing of cluster-wide and multi-tenant control planes

If we decide to stop supporting multi-tenant altogether, we could remove them, but until then, I feel it's best to keep them, so that both cluster-wide and multi-tenant have the same API (i.e. the same way of configuring OSSM).

build/patch-charts.sh
jewertow: Could you explain why?

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -151,20 +152,31 @@ func add(mgr manager.Manager, r *MemberRollReconciler) error {
...
 	if err != nil {
 		log.Error(err, "Could not list ServiceMeshMemberRolls")
 	}
 
 	var requests []reconcile.Request
 	for _, smmr := range list.Items {
-		requests = append(requests, reconcile.Request{
-			NamespacedName: common.ToNamespacedName(&smmr),
-		})
+		if smmr.Spec.IsClusterScoped() || contains(smmr.Spec.Members, namespace) {
```
jewertow: I don't understand this condition. First condition will return true in cluster-wide mode and the second one will return true in multi-tenant mode, so why we need to check this condition if there is no other alternative?

pkg/controller/servicemesh/memberroll/controller.go
dgn: nevermind, my thinking was that the initial creation of these Member resource might be in big clusters. but I guess it's negligible given that we also have to create NetworkAttachmentDefinitions

pkg/controller/versions/strategy_v2_3.go
```diff
@@ -27,6 +30,11 @@ import (
 	"github.com/maistra/istio-operator/pkg/controller/common/helm"
 )
 
+const (
+	clusterScopedSMCPNamespace = "istio-system"
```
dgn, line 34: why hard-code the namespace?

pkg/controller/servicemesh/memberroll/controller.go
dgn: he used to filter based on `spec.members` above, but removed the filter, so that we get notified for every namespace event- this means that either we're in cluster-scoped mode, then we accept all (technically we could pre-filter here based on `isExcludedNamespace()`) or we're in MT mode where we still have to filter based on `spec.members`

dgn:
> > I understand I'm very late to the party on this, and I suspect we chose the Member-based approach to this to keep implementation effort down. Do we have a migration path away from it, though? With cluster-scoped deployments, the MemberRoll and Member controllers should be idle IMO. We're putting a lot of load on the API server for functionality we don't require - the ignoring of certain namespaces could be done in the istio controlplane/injection webhook itself.
> 
> When I started working on this, I understood the main goal was only to improve the performance in large clusters. I kept the Member/MemberRoll because I felt that it's better to keep the "feel" of OSSM the same as it was. Also, we can't just get rid of the Member controller, since we need it to create NetworkAttachmentDefinitions (and potentially NetworkPolicies, etc.) in the member namespaces. Sure, we could modify the controller to do the same work without requiring a Member resource to be present, but what would be the point? The Member resource itself allows you to see if a namespace is part of the mesh and get the status of the member. And it's not really putting any load on the API server. Not sure what you meant.

To the load argument: I simply thought that creation and deletion of Member resources and all the namespace labelling would be unnecessary load. You're right in that it's probably not a lot though, as it only happens on SMCP creation. Load is not really my main concern.

I would argue that the MemberRoll/Member semantics don't make sense in a cluster-wide mesh, as everything's a member (except for some hard-coded namespaces we define in code - another thing I dislike a bit :wink:). And when you look at their controllers, they basically configure a namespace by:
1. Creating NetworkAttachmentDefinition
2. Creating NetworkPolicies
3. Labelling it (member-of), which is then used for injection (and the NetworkPolicy targeting)

IMO none of these are needed for cluster-wide meshes. I'll go through all three:
1. There could be a single NetworkAttachmentDefinition, [referenced from other namespaces](https://github.com/k8snetworkplumbingwg/multus-cni/issues/713#issuecomment-911706484).
2. NetworkPolicy creation (at least the way we did it until now) is just outright dangerous in this setting. You're basically configuring the cluster to allow traffic from any namespace to any other namespace - except the openshift- and kube- namespaces (no idea what implications this might have tbh), overriding any existing NetworkPolicy setup there might be that created hard boundaries between namespaces. So I'd suggest we don't even allow NetworkPolicy creation in cluster-wide mode (ie we default to disabling it and deny validation if you try to enable it)
3. We simply don't need the labels anymore. What meaning does this member-of label have if every namespace gets it? We could just remove the labelSelector on the namespace in the WebhookConfig. Shouldn't be an issue as we're still opt-in.

In the end, the cluster-wide mesh could exist without the two-step install that results from the SMCP/SMM(R) controller duality. It would e.g. solve the problem that people have where they have to wait for their namespace to be configured before they can `kubectl apply` their workloads. Now, they just need to wait until the SMCP is reconciled and get started.



> > How hard will it be to remove Member/MemberRoll controllers from the equation later on? I can see we're relying on them for a) avoiding injection into special namespaces and b) avoiding mixing of cluster-wide and multi-tenant control planes

> If we decide to stop supporting multi-tenant altogether, we could remove them, but until then, I feel it's best to keep them, so that both cluster-wide and multi-tenant have the same API (i.e. the same way of configuring OSSM).

Member and MemberRoll made sense for multi-tenancy, but let's not carry them over for continuity's sake- in this current setup, users have to do 4 (!) things to deploy a cluster-wide mesh:

1. Set the techPreview field
2. Name their SMCP appropriately (clusterscoped)
3. Deploy it to the right namespace (istio-system)
4. Create the correct SMMR (with an asterisk as member)

I would hope step 1 to be enough. All the rest can be deduced by us and IMO shouldn't require additional configuration by the user, it's just confusing.

dgn:
As I said: I know I'm late to the party here, so I don't intend to block this PR. We can merge it as-is. My only ask is that we don't back ourselves into a corner regarding user experience. As this will be tech preview I guess we can still change the behaviour

build/patch-charts.sh
luksa: Since CustomResourceDefinition is a cluster-scoped resource, you need a cluster-scoped privilege to list them. In cluster-scoped mode, we give istiod many other cluster-scoped privileges, so might as well allow it to list CRDs.

pkg/controller/servicemesh/memberroll/controller.go
luksa: Exactly. I've now made this a bit more clear extracting this into a new function `isMember(ns, smmr)`. This new function now also checks if the namespace is excluded. Thanks @dgn.

pkg/controller/versions/strategy_v2_3.go
luksa: For the same reason I hardcoded the SMCP name - so it's a singleton. Of course, singleton resources are typically cluster-scoped, whereas SMCP isn't. We may end up introducing a cluster-scoped resource before this goes GA.

luksa:
> I would argue that the MemberRoll/Member semantics don't make sense in a cluster-wide mesh, as everything's a member (except for some hard-coded namespaces we define in code - another thing I dislike a bit wink). 

This was my first stab at this feature. I thought this PR would be merged weeks ago, and I'd improve on it. One of the plans is to also introduce `SMMR.spec.excludedNamespaces`, which users could configure themselves, but would also automatically be populated with those kube-* and openshift-* namespaces that are currently hardcoded - this would make it easy for users to see which namespaces are excluded by default.

And when you look at their controllers, they basically configure a namespace by:
> 
> 1. Creating NetworkAttachmentDefinition
> 2. Creating NetworkPolicies
> 3. Labelling it (member-of), which is then used for injection (and the NetworkPolicy targeting)
> 
> IMO none of these are needed for cluster-wide meshes. I'll go through all three:
> 
> 1. There could be a single NetworkAttachmentDefinition, [referenced from other namespaces](https://github.com/k8snetworkplumbingwg/multus-cni/issues/713#issuecomment-911706484).
> 2. NetworkPolicy creation (at least the way we did it until now) is just outright dangerous in this setting. You're basically configuring the cluster to allow traffic from any namespace to any other namespace - except the openshift- and kube- namespaces (no idea what implications this might have tbh), overriding any existing NetworkPolicy setup there might be that created hard boundaries between namespaces. So I'd suggest we don't even allow NetworkPolicy creation in cluster-wide mode (ie we default to disabling it and deny validation if you try to enable it)
> 3. We simply don't need the labels anymore. What meaning does this member-of label have if every namespace gets it? We could just remove the labelSelector on the namespace in the WebhookConfig. Shouldn't be an issue as we're still opt-in.
> 
> In the end, the cluster-wide mesh could exist without the two-step install that results from the SMCP/SMM(R) controller duality. It would e.g. solve the problem that people have where they have to wait for their namespace to be configured before they can `kubectl apply` their workloads. Now, they just need to wait until the SMCP is reconciled and get started.

I agree with all this, but I wasn't sure what was actually required by our users, so I kept the changes to a minimum. One of the requirements was to not allow more than one mesh in a cluster if the cluster-wide option is used. But it's already clear that this won't be ok. 

We'll discuss further development of this feature once we get initial feedback from QE, customers, support, etc.

> > > How hard will it be to remove Member/MemberRoll controllers from the equation later on? I can see we're relying on them for a) avoiding injection into special namespaces and b) avoiding mixing of cluster-wide and multi-tenant control planes
 
I don't see anything I did in this PR that would prevent us from later removing these resources/controllers.

> > If we decide to stop supporting multi-tenant altogether, we could remove them, but until then, I feel it's best to keep them, so that both cluster-wide and multi-tenant have the same API (i.e. the same way of configuring OSSM).
> 
> Member and MemberRoll made sense for multi-tenancy, but let's not carry them over for continuity's sake- in this current setup, users have to do 4 (!) things to deploy a cluster-wide mesh:
> 
> 1. Set the techPreview field
> 2. Name their SMCP appropriately (clusterscoped)
> 3. Deploy it to the right namespace (istio-system)
> 4. Create the correct SMMR (with an asterisk as member)
> 
> I would hope step 1 to be enough. All the rest can be deduced by us and IMO shouldn't require additional configuration by the user, it's just confusing.

+1

PR #1011, opened 2022-09-26
OSSM-1712 Use vanilla prometheus

yannuil:


jwendell:
@yannuil I think you need to rebase your PR. I think your base branch is 2.2 and you opened this PR against 2.3?

pkg/controller/common/config.go
```diff
@@ -84,15 +84,16 @@ type v2_1ImageNames struct {
...
-	ProxyInit  string `json:"proxy-init,omitempty"`
-	ProxyV2    string `json:"proxyv2,omitempty"`
-	WASMCacher string `json:"wasm-cacher,omitempty"`
-	RLS        string `json:"rls,omitempty"`
+	ThreeScale               string `json:"3scale-istio-adapter,omitempty"`
+	CNI                      string `json:"cni,omitempty"`
+	Grafana                  string `json:"grafana,omitempty"`
+	Pilot                    string `json:"pilot,omitempty"`
+	Prometheus               string `json:"prometheus,omitempty"`
+	PrometheusConfigReloader string `json:"prometheus-config-reloader,omitempty"`
```
jewertow, line 93: I guess that you forgot to move this change to `v2_3ImageNames`.

deploy/servicemesh-operator.yaml
```diff
@@ -10727,6 +10727,7 @@ spec:
         olm.relatedImage.v2_3.grafana: ${OSSM_GRAFANA_IMAGE}
         olm.relatedImage.v2_3.pilot: ${OSSM_PILOT_IMAGE}
         olm.relatedImage.v2_3.prometheus: ${OSSM_PROMETHEUS_IMAGE}
+        olm.relatedImage.v2_3.prometheus-config-reloader: ${OSSM_PROMETHEUS_CONFIG_RELOADER_IMAGE}
```
jewertow, line 10713: You should also add this image to `deploy/maistra-operator.yaml` and `deploy/src/deployment-maistra.yaml`.

manifests-servicemesh/2.3.0/servicemeshoperator.v2.3.0.clusterserviceversion.yaml
```diff
@@ -572,6 +572,7 @@ spec:
                 olm.relatedImage.v2_3.grafana: ${OSSM_GRAFANA_IMAGE}
                 olm.relatedImage.v2_3.pilot: ${OSSM_PILOT_IMAGE}
                 olm.relatedImage.v2_3.prometheus: ${OSSM_PROMETHEUS_IMAGE}
+                olm.relatedImage.v2_3.prometheus-config-reloader: ${OSSM_PROMETHEUS_CONFIG_RELOADER_IMAGE}
```
jewertow, line 550: You should also update a corresponding manifest in `manifests-maistra`.

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -323,6 +342,19 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
...
 	}
 
+	// 7. tell Prometheus about all the namespaces in the mesh
+	var prometheusErr error
+
+	if mesh.Status.AppliedSpec.IsPrometheusEnabled() {
+		cv, err := versions.ParseVersion(mesh.Status.AppliedSpec.Version)
+
+		if err != nil {
+			prometheusErr = err
```
jewertow, line 394: I think that it would be nice to add more context to this error, e.g.
```
prometheusErr = pkgerrors.Wrapf(err, "could not reconcile Prometheus, because of unknown SMCP version")
```
WDYT?

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -361,6 +397,8 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
 		return reconcile.Result{}, err
 	} else if kialiErr != nil {
 		return reconcile.Result{}, kialiErr
+	} else if prometheusErr != nil {
+		return reconcile.Result{}, prometheusErr
```
jewertow, line 452: Some time ago I added helpers to improve readability of reconciliation, so you could replace this with:
```
return common.RequeueWithError(prometheusErr)
```

I would even suggest to refactor this if-else statement entirely as you can see below. I really don't like reusing `err` variable, because it makes harder to ensure that all errors are handled properly.
```
	if err = r.updateStatus(ctx, roll); err != nil {
		return common.RequeueWithError(err)
	} else if kialiErr != nil {
		return common.RequeueWithError(kialiErr)
	} else if prometheusErr != nil {
		return common.RequeueWithError(prometheusErr)
	}
	return common.Reconciled()
```

pkg/controller/versions/strategy_v2_3.go
```diff
@@ -100,6 +100,9 @@ func (v *versionStrategyV2_3) SetImageValues(_ context.Context, _ *common.Contro
 	if err := common.UpdateField(smcpSpec.Istio, "prometheus.image", common.Config.OLM.Images.V2_3.Prometheus); err != nil {
 		return err
 	}
+	if err := common.UpdateField(smcpSpec.Istio, "prometheusConfigReloader.image", common.Config.OLM.Images.V2_2.PrometheusConfigReloader); err != nil {
```
jewertow, line 107:
```suggestion
	if err := common.UpdateField(smcpSpec.Istio, "prometheusConfigReloader.image", common.Config.OLM.Images.V2_3.PrometheusConfigReloader); err != nil {
```

jewertow:
What Prometheus images should I use to test this change?

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -598,6 +636,126 @@ func (r *defaultKialiReconciler) reconcileKiali(ctx context.Context, kialiCRName
...
+func (r *defaultPrometheusReconciler) reconcilePrometheus(ctx context.Context, prometheusCMName, prometheusNamespace string, members []string) error {
+	reqLogger := common.LogFromContext(ctx)
+	reqLogger.Info("Attempting to get Prometheus ConfigMap", "prometheusNamespace", prometheusNamespace, "prometheusCMName", prometheusCMName)
+
+	cm := &corev1.ConfigMap{}
+
+	err := r.Client.Get(ctx, client.ObjectKey{Name: prometheusCMName, Namespace: prometheusNamespace}, cm)
+	if err != nil {
+		if meta.IsNoMatchError(err) || errors.IsNotFound(err) || errors.IsGone(err) {
+			reqLogger.Info("Prometheus Config Map does not exist.")
```
jewertow, line 747:
```suggestion
			reqLogger.Info("Prometheus ConfigMap '%s' does not exist", prometheusCMName)
```
or even
```suggestion
			reqLogger.Info("Prometheus ConfigMap '%s/%s' does not exist", prometheusNamespace, prometheusCMName)
```

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -598,6 +636,126 @@ func (r *defaultKialiReconciler) reconcileKiali(ctx context.Context, kialiCRName
...
+			return nil
+		}
+		return pkgerrors.Wrap(err, "error retrieving Prometheus ConfigMap from mesh")
+	}
+
+	data := make(map[string]interface{})
+
+	err = yaml.Unmarshal([]byte(cm.Data[prometheusConfigurationFilename]), &data)
```
jewertow, line 755: Does it work when config map was not found?

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -598,6 +636,126 @@ func (r *defaultKialiReconciler) reconcileKiali(ctx context.Context, kialiCRName
...
+			return fmt.Errorf("error converting scrape_config from %v", s)
+		}
+
+		if scrape["job_name"] == "pilot" {
+			continue
+		}
+
+		f, ok, err := unstructured.NestedFieldNoCopy(scrape, prometheusScrapeSDConfigKeyName)
+
+		if err == nil {
```
jewertow, line 788: When you handle errors first, then you have less nested code and less indentations.
```suggestion
		if err != nil {
		        return pkgerrors.Wrap(err, fmt.Sprintf("error getting sd from %v", s))
		}
		if ok {
		        ...
```
It would be more readable in my opinion. WDYT?

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -598,6 +636,126 @@ func (r *defaultKialiReconciler) reconcileKiali(ctx context.Context, kialiCRName
...
+				} else {
+					return fmt.Errorf("error can not process sd %v", f)
+				}
+			} else {
+				reqLogger.Info(fmt.Sprintf("Ignoring scrape %v", s))
+			}
+		} else {
+			return pkgerrors.Wrap(err, fmt.Sprintf("error getting sd from %v", s))
+		}
+	}
```
jewertow, line 815: The logic above is quite tricky and difficult to read, so I think that we have to extract it to a separate function and write unit tests to make sure this is correct.

jewertow:
Could you add a short summary to this PR? I would like to read a high-level overview of the implementation, like why `prometheus-config-reloader` is needed, and what I should pay attention to in particular.

jewertow:
I'm sorry that I care about the code style, but I think this code could be more concise. IMHO there is too many new lines which distract me. Please try to follow existing style.

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -598,6 +636,126 @@ func (r *defaultKialiReconciler) reconcileKiali(ctx context.Context, kialiCRName
...
 
+type PrometheusReconciler interface {
+	reconcilePrometheus(ctx context.Context, prometheusCMName, prometheusNamespace string, configuredMembers []string) error
+}
+
+type defaultPrometheusReconciler struct {
+	Client client.Client
+}
+
+func (r *defaultPrometheusReconciler) reconcilePrometheus(ctx context.Context, prometheusCMName, prometheusNamespace string, members []string) error {
```
jewertow, line 738: It's quite a big function, so I would prefer to move it to another file in the same package and split it into smaller functions to improve readability. WDYT?

dgn:
Sorry, I must have overlooked this PR. @yannuil I see you're using a container called `prometheus-config-reloader` to do the reloading of the config. I'm not sure that's something we want to do, as it would mean we have to productize yet another container image and that's always a pain (plus we need to monitor and update it for CVEs etc). Could you add a description here that describes why you took this route, and maybe also if there might be one that does not need a new container image?

pkg/controller/servicemesh/memberroll/controller.go
dgn: +1 and let's move this into `prometheus.go` (and maybe the kiali stuff to `kiali.go` while we're at it)

pkg/controller/versions/strategy_v2_3.go
```diff
@@ -351,6 +354,35 @@ func (v *versionStrategyV2_3) Render(ctx context.Context, cr *common.ControllerR
 			}
 		}
 	}
+
+	if isComponentEnabled(spec.Istio, v2_3ChartMapping[PrometheusChart].enabledField) {
+		members := &v1.ServiceMeshMemberList{}
+
+		err := cr.Client.List(ctx, members, client.MatchingFields{"spec.controlPlaneRef.namespace": smcp.GetNamespace()})
```
dgn, line 492: as discussed on the call, we should be `Get()`ing the `MemberRoll` here and look at its `.status.configuredMembers` instead of listing the `Members`

dgn:
@yannuil let's resurrect this PR :)

yannuil:
@dgn I am working on adding the suggested changes and tests.

yannuil:
Closed in favor of #1089

PR #1014, opened 2022-09-27
OSSM-2053 Don't remove configured (but out-of-date) members from SMMR.status.configuredMembers [maistra-2.2]

luksa:
After an SMCP is updated, the previously configured members should never disappear from the SMMR.status.configuredMembers list, since this causes istiod to push empty configs to the proxies. These namespaces should still be considered configured, even though they might not be up-to-date with the SMCP until the member controller reconciles them.

Here, we ensure that a namespace that was previously configured successfully, but is now out-of-date with regard to the SMCP, appears in both the configuredMembers and the pendingMembers list.


jwendell:
2.2 is failing until https://github.com/maistra/istio-operator/pull/1009 gets merged

rcernich:
I updated the test case so it will pass.

rcernich (approved):

PR #1016, opened 2022-09-28
OSSM-1188 Remove ServiceMeshExtension support

dgn:
This removes the wasm-cache chart from the 2.3 helm charts. It also adds upgrade validation: upgrading to v2.3 will fail if you have existing SMEs in your member namespaces. This check however is only performed when upgrading, not when creating new SMCPs, reasons being:

- when you're recreating an SMCP, you probably don't expect an existing SME to "still work". It's a new deployment after all.
- the check is a little costly, so it makes sense to only run it when absolutely necessary


pkg/controller/versions/strategy_v2_3.go
```diff
@@ -198,6 +200,41 @@ func (v *versionStrategyV2_3) validateAddons(spec *v2.ControlPlaneSpec, allError
 	}
 	return allErrors
 }
+func (v *versionStrategyV2_3) validateServiceMeshExtensionsRemoved(ctx context.Context, cl client.Client, smcp metav1.Object) error {
+	logger := logf.Log.WithName("smcp-validator-2.3")
+
+	serviceMeshExtensions := &apiv1.ServiceMeshExtensionList{}
+	if err := cl.List(ctx, serviceMeshExtensions); err != nil {
+		fmt.Println(err)
```
jwendell: remove this?

pkg/controller/versions/strategy_v2_3.go
```diff
@@ -198,6 +200,41 @@ func (v *versionStrategyV2_3) validateAddons(spec *v2.ControlPlaneSpec, allError
...
+		if !errors.IsNotFound(err) {
+			logger.Error(err, "failed to list ServiceMeshExtensions")
```
jwendell: it failed because there's no SME CR, so, it's expected and not an error?

pkg/controller/versions/strategy_v2_3.go
```diff
@@ -198,6 +200,41 @@ func (v *versionStrategyV2_3) validateAddons(spec *v2.ControlPlaneSpec, allError
...
+		err := cl.Get(ctx, client.ObjectKey{Name: common.MemberRollName, Namespace: smcp.GetNamespace()}, smmr)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				// log error, but don't fail validation: we'll just assume that the control plane namespace is the only namespace for now
+				logger.Error(err, "failed to retrieve SMMR for SMCP")
+				smmr = nil
+			}
+		}
+		meshNamespaces := common.GetMeshNamespaces(smcp.GetNamespace(), smmr)
+		fmt.Println(meshNamespaces.List())
```
jwendell: remove this?

pkg/controller/versions/strategy_v2_3.go
dgn: whoops

pkg/controller/versions/strategy_v2_3.go
dgn: whoops

pkg/controller/versions/strategy_v2_3.go
dgn: not sure what you mean? the test passes locally

jwendell: I mean: trying to list a SME and get a failure should not be logged as an error, should it? If there's no SME CRD in the cluster, for example, an error is expected and desirable, right?

jwendell: Also, there's a println hidden in there :D

dgn: > I mean: trying to list a SME and get a failure should not be logged as an error, should it? If there's no SME CRD in the cluster, for example, an error is expected and desirable, right?

Hmm in that case you'd be right. We will have to still install the CRD though, as a user might install a pre-2.3 version - and we always install the same set of CRDs for all control plane versions. But I can check for that error


> Also, there's a println hidden in there :D

not anymore, I removed that :wink: it's just still visible here for some reason

pkg/controller/versions/strategy_v2_3.go
```diff
@@ -199,6 +201,41 @@ func (v *versionStrategyV2_3) validateAddons(spec *v2.ControlPlaneSpec, allError
 	return allErrors
 }
 
+func (v *versionStrategyV2_3) validateServiceMeshExtensionsRemoved(ctx context.Context, cl client.Client, smcp metav1.Object) error {
+	logger := logf.Log.WithName("smcp-validator-2.3")
+
+	serviceMeshExtensions := &apiv1.ServiceMeshExtensionList{}
+	if err := cl.List(ctx, serviceMeshExtensions); err != nil {
+		if !errors.IsNotFound(err) && !meta.IsNoMatchError(err) {
+			// log error, but don't fail validation- this is a convenience check
```
jewertow: I think that this assumption is not correct. Let's say that a user has SME instances and upgrades to 2.3, but the operator couldn't list SMEs, because of a timeout or any other transient error. In such a case, validation will pass and upgrade will be performed while it shouldn't.

pkg/controller/versions/strategy_v2_3.go
```diff
@@ -199,6 +201,41 @@ func (v *versionStrategyV2_3) validateAddons(spec *v2.ControlPlaneSpec, allError
...
+		}
+	}
+	if len(serviceMeshExtensions.Items) > 0 {
+		smmr := &v1.ServiceMeshMemberRoll{}
+		err := cl.Get(ctx, client.ObjectKey{Name: common.MemberRollName, Namespace: smcp.GetNamespace()}, smmr)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				// log error, but don't fail validation: we'll just assume that the control plane namespace is the only namespace for now
+				logger.Error(err, "failed to retrieve SMMR for SMCP")
+				smmr = nil
```
jewertow: I think that this assignment is right only when  `errors.IsNotFound(err)` is true. Otherwise we shouldn't assume anything, because of transient errors, like timeouts etc.

pkg/controller/versions/strategy_v2_3.go
dgn: yeah the trade-off here was between 
- accidentally blocking a valid upgrade
- accidentally allowing an upgrade where there are still SMEs in the member namespaces

I figured that accidentally blocking a valid upgrade would be worse, but I'm ready to be convinced otherwise

pkg/controller/versions/strategy_v2_3.go
dgn: again, same trade-off as above

jwendell (approved):

pkg/controller/versions/strategy_v2_3.go
jewertow: My point of view is: why to validate something if we ignore errors and force passing?

jewertow: As a user, I would prefer to get an error on validation and try again or stop upgrade, instead of get false postive validation and accidentally upgrade operator and then investigate why my WASM plugins does not work.

jewertow: I was thinking a lot about this trade-off and my final position is that we should not allow for an upgrade in case of validation failure, because consequences of an unexpected upgrade are incomparably worse than failed upgrade that can be retried. We don't know what WASM filters customers use, so we should assume that breaking them may cause a downtime.

dgn: You convinced me, @jewertow :wink: I made the changes, PTAL

jewertow (approved):

PR #1017, opened 2022-09-28
OSSM-1968 Move 2.3 CNI container into separate DaemonSets

yxun:
Signed-off-by: Yuanlin <yuanlin.xu@redhat.com>

This PR replaces the cni template daemonset.yaml with separated ones: daemonset-v2.0.yaml, daemonset-v2.1.yaml, daemonset-v2.2.yaml and daemonset-v2.3.yaml

I use the existing cni.supportedReleases values to filter a SMCP controller version, because there is no direct mapping between cni bootstrap and a controller spec version. 

In upgrade case, the new boostrap code will check and delete the existing "istio-node" and "istio-cni-node" DaemonSets if they exist.

### DIFF
--- a/pkg/bootstrap/cni.go
+++ b/pkg/bootstrap/cni.go
@@ -20,13 +20,20 @@ import (
 
 // InstallCNI makes sure all Istio CNI resources have been created.  CRDs are located from
 // files in controller.HelmDir/istio-init/files
-func InstallCNI(ctx context.Context, cl client.Client, config cni.Config) error {
+func InstallCNI(ctx context.Context, cl client.Client, config cni.Config, instanceVersion versions.Version) error {
+	// instanceVersion is from a SMCP spec version
+	if instanceVersion == nil {
+		instanceVersion = versions.DefaultVersion.Version()
+	}
 	// we should run through this each reconcile to make sure it's there
-	return internalInstallCNI(ctx, cl, config)
+	return internalInstallCNI(ctx, cl, config, instanceVersion)
 }
 
-func internalInstallCNI(ctx context.Context, cl client.Client, config cni.Config) error {
-	renderings, err := internalRenderCNI(ctx, cl, config, versions.GetSupportedVersions())
+func internalInstallCNI(ctx context.Context, cl client.Client, config cni.Config, instanceVersion versions.Version) error {
+	if instanceVersion == nil {
+		instanceVersion = versions.DefaultVersion.Version()
+	}
+	renderings, err := internalRenderCNI(ctx, cl, config, versions.GetSupportedVersions(), instanceVersion)
 	if err != nil {
 		return err
 	}
@@ -34,7 +41,7 @@ func internalInstallCNI(ctx context.Context, cl client.Client, config cni.Config
 }
 
 func internalRenderCNI(ctx context.Context, cl client.Client, config cni.Config,
-	supportedVersions []versions.Version,
+	supportedVersions []versions.Version, instanceVersion versions.Version,
 ) (renderings map[string][]manifest.Manifest, err error) {
 	log := common.LogFromContext(ctx)
 	log.Info("ensuring Istio CNI has been installed")
@@ -82,6 +89,12 @@ func internalRenderCNI(ctx context.Context, cl client.Client, config cni.Config,
 		releases = append(releases, versions.DefaultVersion.String())
 	}
 	cni["supportedReleases"] = releases
+	// v2.3 render the required cni daemonset,
+	// instanceVersion is from a SMCP spec version
+	if instanceVersion == nil {
+		instanceVersion = versions.DefaultVersion.Version()
+	}
+	cni["instanceVersion"] = instanceVersion.String()
 
 	values := make(map[string]interface{})
 	values["cni"] = cni
--- a/pkg/bootstrap/cni.go
+++ b/pkg/bootstrap/cni.go
@@ -21,19 +21,13 @@ import (
 
 // InstallCNI makes sure all Istio CNI resources have been created.  CRDs are located from
 // files in controller.HelmDir/istio-init/files
+// ver is from a SMCP spec version
 func InstallCNI(ctx context.Context, cl client.Client, config cni.Config, dc discovery.DiscoveryInterface, ver versions.Version) error {
-	// ver is from a SMCP spec version
-	if ver == nil {
-		ver = versions.DefaultVersion.Version()
-	}
 	// we should run through this each reconcile to make sure it's there
 	return internalInstallCNI(ctx, cl, config, dc, ver)
 }
 
 func internalInstallCNI(ctx context.Context, cl client.Client, config cni.Config, dc discovery.DiscoveryInterface, ver versions.Version) error {
-	if ver == nil {
-		ver = versions.DefaultVersion.Version()
-	}
 	renderings, err := internalRenderCNI(ctx, cl, config, dc, versions.GetSupportedVersions(), ver)
 	if err != nil {
 		return err
@@ -67,7 +61,7 @@ func internalRenderCNI(ctx context.Context, cl client.Client, config cni.Config,
 	cni["configMap_v2_0"] = "cni_network_config_v2_0"
 	cni["configMap_v2_1"] = "cni_network_config_v2_1"
 	cni["configMap_v2_2"] = "cni_network_config_v2_2"
-	cni["configMap_v2_3"] = "cni_network_config_v2_3"
+	cni["configMap_v2_3"] = "cni_network_config"
 
 	cni["chained"] = !config.UseMultus
 	if config.UseMultus {
@@ -92,9 +86,6 @@ func internalRenderCNI(ctx context.Context, cl client.Client, config cni.Config,
 	cni["supportedReleases"] = releases
 	// v2.3 render the required cni daemonset,
 	// instanceVersion is from a SMCP spec version
-	if ver == nil {
-		ver = versions.DefaultVersion.Version()
-	}
 	cni["instanceVersion"] = ver.String()
 
 	values := make(map[string]interface{})
--- a/pkg/bootstrap/cni_test.go
+++ b/pkg/bootstrap/cni_test.go
@@ -26,36 +26,49 @@ func TestCNISupportedVersionRendering(t *testing.T) {
 	testCases := []struct {
 		name              string
 		supportedVersions []versions.Version
+		instanceVersion   versions.Version
 		containerNames    []string
 	}{
 		{
-			name:              "Default Supported Versions",
+			name:              "Default Supported Versions v2.2",
 			supportedVersions: versions.GetSupportedVersions(),
-			containerNames:    []string{"install-cni-v1-1", "install-cni-v2-0", "install-cni-v2-1", "install-cni-v2-2", "install-cni-v2-3"},
+			instanceVersion:   versions.V2_2.Version(),
+			containerNames:    []string{"install-cni-v1-1", "install-cni-v2-0", "install-cni-v2-1", "install-cni-v2-2"},
+		},
+		{
+			name:              "Default Supported Versions v2.3",
+			supportedVersions: versions.GetSupportedVersions(),
+			instanceVersion:   versions.V2_3.Version(),
+			containerNames:    []string{"install-cni-v2-3"},
 		},
 		{
 			name:              "v1.1 only",
 			supportedVersions: []versions.Version{versions.V1_1},
+			instanceVersion:   versions.V1_1.Version(),
 			containerNames:    []string{"install-cni-v1-1"},
 		},
 		{
 			name:              "v2.0 only",
 			supportedVersions: []versions.Version{versions.V2_0},
+			instanceVersion:   versions.V2_0.Version(),
 			containerNames:    []string{"install-cni-v2-0"},
 		},
 		{
 			name:              "v2.1 only",
 			supportedVersions: []versions.Version{versions.V2_1},
+			instanceVersion:   versions.V2_1.Version(),
 			containerNames:    []string{"install-cni-v2-1"},
 		},
 		{
 			name:              "v2.2 only",
 			supportedVersions: []versions.Version{versions.V2_2},
+			instanceVersion:   versions.V2_2.Version(),
 			containerNames:    []string{"install-cni-v2-2"},
 		},
 		{
 			name:              "v2.3 only",
 			supportedVersions: []versions.Version{versions.V2_3},
+			instanceVersion:   versions.V2_3.Version(),
 			containerNames:    []string{"install-cni-v2-3"},
 		},
 	}
@@ -67,7 +80,7 @@ func TestCNISupportedVersionRendering(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			cl, _ := test.CreateClient()
-			renderings, err := internalRenderCNI(ctx, cl, config, tc.supportedVersions)
+			renderings, err := internalRenderCNI(ctx, cl, config, tc.supportedVersions, tc.instanceVersion)
 			assert.Success(err, "internalRenderCNI", t)
 			assert.True(renderings != nil, "renderings should not be nil", t)
 			cniManifests := renderings["istio_cni"]
--- a/pkg/bootstrap/cni_test.go
+++ b/pkg/bootstrap/cni_test.go
@@ -29,48 +29,56 @@ func TestCNISupportedVersionRendering(t *testing.T) {
 		supportedVersions []versions.Version
 		instanceVersion   versions.Version
 		containerNames    []string
+		daemonsetName     string
 	}{
 		{
-			name:              "Default Supported Versions v2.2",
+			name:              "Default Supported Versions SMCP v2.2",
 			supportedVersions: versions.GetSupportedVersions(),
 			instanceVersion:   versions.V2_2.Version(),
 			containerNames:    []string{"install-cni-v1-1", "install-cni-v2-0", "install-cni-v2-1", "install-cni-v2-2"},
+			daemonsetName:     "istio-cni-node",
 		},
 		{
-			name:              "Default Supported Versions v2.3",
+			name:              "Default Supported Versions SMCP v2.3",
 			supportedVersions: versions.GetSupportedVersions(),
 			instanceVersion:   versions.V2_3.Version(),
-			containerNames:    []string{"install-cni-v2-3"},
+			containerNames:    []string{"install-cni"},
+			daemonsetName:     "istio-cni-node-v2-3",
 		},
 		{
 			name:              "v1.1 only",
 			supportedVersions: []versions.Version{versions.V1_1},
 			instanceVersion:   versions.V1_1.Version(),
 			containerNames:    []string{"install-cni-v1-1"},
+			daemonsetName:     "istio-cni-node",
 		},
 		{
 			name:              "v2.0 only",
 			supportedVersions: []versions.Version{versions.V2_0},
 			instanceVersion:   versions.V2_0.Version(),
 			containerNames:    []string{"install-cni-v2-0"},
+			daemonsetName:     "istio-cni-node",
 		},
 		{
 			name:              "v2.1 only",
 			supportedVersions: []versions.Version{versions.V2_1},
 			instanceVersion:   versions.V2_1.Version(),
 			containerNames:    []string{"install-cni-v2-1"},
+			daemonsetName:     "istio-cni-node",
 		},
 		{
 			name:              "v2.2 only",
 			supportedVersions: []versions.Version{versions.V2_2},
 			instanceVersion:   versions.V2_2.Version(),
 			containerNames:    []string{"install-cni-v2-2"},
+			daemonsetName:     "istio-cni-node",
 		},
 		{
 			name:              "v2.3 only",
 			supportedVersions: []versions.Version{versions.V2_3},
 			instanceVersion:   versions.V2_3.Version(),
-			containerNames:    []string{"install-cni-v2-3"},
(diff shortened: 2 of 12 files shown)


yxun:
@jewertow @luksa 
PR ready for a review but not ready for merge. I will test my operator image on an OCP later today. Thanks,

yxun:
I will convert it from Draft to ready for review after fixing CI issues. The CI issues are waiting for CI team approval in another repo PR.

dgn (commented):
Wasn't the intention to only install those DaemonSets that are required? In that case, I would say we extend `InstallCNI()` to also pass the SMCP version (it's called from within the reconciler, so that should be easy), which can then be passed on to `internalRenderCNI()` so we don't always render all DaemonSets

jewertow:
I think that the goal of this issue was to install a separated daemonset only for 2.3. Previous versions should work as before.

luksa:
> I think that the goal of this issue was to install a separated daemonset only for 2.3. Previous versions should work as before.

+1

Although it would be nice to have separate daemonsets for each version, this would be a bigger change and could introduce upgrade issues. That was one of the reasons we decided to keep one daemonset for all the older versions and only introduce a new one for 2.3 (when we do 2.4, we'll add another daemonset, and so on...)

yxun:
> I think that the goal of this issue was to install a separated daemonset only for 2.3. Previous versions should work as before.

Ok, that makes this PR easier. i will update it today. 

> Wasn't the intention to only install those DaemonSets that are required? In that case, I would say we extend InstallCNI() to also pass the SMCP version (it's called from within the reconciler, so that should be easy), which can then be passed on to internalRenderCNI() so we don't always render all DaemonSets

Yes, totally agrees. I spent some time trying to pass the SMCP controller version but it turned out to be a bigger change than I expected. 
After 2.3 release, I can read more and think about a way of extending that. Thanks.

yxun:
unit test passed in my local laptop and the new job but failed in our prow job. is there anything changed in our prow ?

pkg/bootstrap/cni_test.go
```diff
@@ -31,7 +31,7 @@ func TestCNISupportedVersionRendering(t *testing.T) {
 		{
 			name:              "Default Supported Versions",
 			supportedVersions: versions.GetSupportedVersions(),
-			containerNames:    []string{"install-cni-v1-1", "install-cni-v2-0", "install-cni-v2-1", "install-cni-v2-2", "install-cni-v2-3"},
+			containerNames:    []string{"install-cni-v2-3", "install-cni-v1-1", "install-cni-v2-0", "install-cni-v2-1", "install-cni-v2-2"},
```
jewertow, line 34: Did this order matter?

jewertow: Does order of these names matter?

yxun: yes, it matters now. because it finds the v2.3 one and append it first

pkg/bootstrap/cni_test.go
```diff
@@ -85,18 +86,21 @@ func TestCNISupportedVersionRendering(t *testing.T) {
 					assert.Success(err, "resource decoding", t)
 
 					containers, found, err := unstructured.NestedSlice(resource.UnstructuredContent(), "spec", "template", "spec", "containers")
+					if tc.name == "v2.3 only" && !found {
+						// moved v2.3 cni container into a new daemonset, skip testing the old one
+						continue
+					}
```
jewertow: To be honest, I don't understand what's going on here. Shouldn't we just remove case "v2.3 only"?

manifests-maistra/2.3.0/maistraoperator.v2.3.0.clusterserviceversion.yaml
```diff
@@ -15,7 +15,7 @@ metadata:
       The Maistra Operator enables you to install, configure, and manage an instance of Maistra service mesh. Maistra is based on the open source Istio project.
 
     containerImage: quay.io/maistra/istio-ubi8-operator:2.3.0
-    createdAt: 2022-09-26T09:28:03MDT
+    createdAt: 2022-09-29T11:23:46EDT
```
jewertow: We usually avoid updating `createdAt`, but I am not sure if it is a problem before we release a particular version...?

pkg/bootstrap/cni_test.go
jewertow: Thanks!

pkg/bootstrap/cni_test.go
yxun: Because in v2.3 case, we will have two daemonsets. One from our original daemonset.yaml and one from the new template daemonset-v2.3.yaml.

When we create a v2.3 SMCP, supportedVersion only has [v2.3] value, so the daemonset.yaml will not include any container. That's why 
```
unstructured.NestedSlice(resource.UnstructuredContent(), "spec", "template", "spec", "containers") 
```
returns nil and found is false in that case.

However, if I change the rest of the if statements in the daemonset.yaml, that will break existing clusters in upgrade.
In that case, only removing the v2.3 container section in the daemonset.yaml template will be safe in upgrade.

manifests-maistra/2.3.0/maistraoperator.v2.3.0.clusterserviceversion.yaml
yxun: That's in the second commit. There are new changes generated from the `make gen` so I think I need to update the manifest date. If that's not good, I can checkout the original values

pkg/bootstrap/cni_test.go
jewertow: Ok, thanks! Now I understand. Could you add more precise comment?

yxun: sure, i will add more

manifests-maistra/2.3.0/maistraoperator.v2.3.0.clusterserviceversion.yaml
jewertow: Yes, but manifest did not change itself. @jwendell could you confirm?

jewertow (commented):
Generally looks good to me. But before I approve, I would like to test it locally, ok?

manifests-maistra/2.3.0/maistraoperator.v2.3.0.clusterserviceversion.yaml
yxun: It changed. I talked to jonh in a slack thread
https://coreos.slack.com/archives/C01R9E3SFEV/p1664464705920469

luksa:
@yxun when I create a 2.3 SMCP, the operator installs both the `istio-cni-node` and `istio-cni-node-v2.3`. The goal was for the operator to only install the 2.3 daemonset. The operator should install the old daemonset only when an SMCP v2.0, v2.1 or v2.2 is created.

yxun:
> @yxun when I create a 2.3 SMCP, the operator installs both the `istio-cni-node` and `istio-cni-node-v2.3`. The goal was for the operator to only install the 2.3 daemonset. The operator should install the old daemonset only when an SMCP v2.0, v2.1 or v2.2 is created.

@luksa Hi Marko, thanks. I updated this PR. And the two tests cover your comment above:
- "v2.3 only" verifies the correct cni container is rendered
https://github.com/maistra/istio-operator/blob/f258ef451cfef2e66aaf30c35bd8c5bea8aebdcc/pkg/bootstrap/cni_test.go#L57

The reason why I moved the "names := []string{}" outside of the for loop is for matching the ""Default Supported Versions" test case.

- ""bootstrap-clean-install-cni-no-errors" verifies the correct daemonset is able to be installed and the other one is not 
https://github.com/maistra/istio-operator/blob/f258ef451cfef2e66aaf30c35bd8c5bea8aebdcc/pkg/controller/servicemesh/controlplane/integration_test.go#L205
https://github.com/maistra/istio-operator/blob/f258ef451cfef2e66aaf30c35bd8c5bea8aebdcc/pkg/controller/servicemesh/controlplane/integration_test.go#L213

luksa:
@yxun, I'm not talking about `supportedVersions`. One of the acceptance criteria in the Jira is the following:

> - operator deploys the correct DaemonSet version when an SMCP of a particular version (2.3 and forward) is deployed

So, you'll need to pass the SMCP version to `InstallCNI()` in [1] and make sure that either the old daemonset, or the 2.3 daemonset gets created, depending on what the SMCP version is.

[1] https://github.com/Maistra/istio-operator/blob/36327f8a0e3f832d1c3a02145de16e4aa4cd268c/pkg/controller/servicemesh/controlplane/reconciler.go#L212

yxun:
@luksa I misunderstood @dgn and your comments before. I have updated the change.

resources/helm/overlays/istio_cni/templates/daemonset-v2.3.yaml
```diff
@@ -0,0 +1,128 @@
+{{ if .Values.cni.enabled }}
+{{- if and (.Values.cni.supportedReleases) (eq .Values.cni.instanceVersion "v2.3") }}
```
luksa: Why check if `supportedReleases` is not nil? Shouldn't this check whether `supportedReleases` contains `v2.3`?

resources/helm/overlays/istio_cni/templates/daemonset.yaml
```diff
@@ -1,4 +1,6 @@
 {{ if .Values.cni.enabled }}
+{{- if (.Values.cni.supportedReleases) }}
```
luksa: Why this check?

resources/helm/overlays/istio_cni/templates/daemonset-v2.3.yaml
```diff
@@ -0,0 +1,128 @@
...
+{{- if .Values.cni.cniConfFileName_v2_3 }}
+            # Name of the CNI config file to create.
+            - name: CNI_CONF_NAME
+              value: "{{ .Values.cni.cniConfFileName_v2_3 }}"
+{{- end }}
+            # The CNI network config to install on each node.
+            - name: CNI_NETWORK_CONFIG
+              valueFrom:
+                configMapKeyRef:
+                  name: istio-cni-config
```
luksa: I'd prefer to see the `istio-cni-configmap` also split into two ConfigMaps, so that every version above 2.3 has its own ConfigMap. This should make it easier to rebase to new versions.

pkg/bootstrap/cni.go
```diff
@@ -20,21 +20,28 @@ import (
 
 // InstallCNI makes sure all Istio CNI resources have been created.  CRDs are located from
 // files in controller.HelmDir/istio-init/files
-func InstallCNI(ctx context.Context, cl client.Client, config cni.Config) error {
+func InstallCNI(ctx context.Context, cl client.Client, config cni.Config, instanceVersion versions.Version) error {
+	// instanceVersion is from a SMCP spec version
+	if instanceVersion == nil {
+		instanceVersion = versions.DefaultVersion.Version()
+	}
```
luksa: Is this necessary? If it is, why not just do this in one place (for example, `internalRenderCNI()`)? But I'd prefer to completely remove this and instead have the code fail. This makes it easier to reason about the code (you either pass the version when you invoke the function, or you don't... if the function uses a default value when the caller passes in nil, it's possible this will eventually lead to some weird bug).

resources/helm/overlays/istio_cni/templates/daemonset-v2.3.yaml
```diff
@@ -0,0 +1,128 @@
+{{ if .Values.cni.enabled }}
+{{- if and (.Values.cni.supportedReleases) (eq .Values.cni.instanceVersion "v2.3") }}
+# This manifest installs the Istio install-cni container, as well
+# as the Istio CNI plugin and config on
+# each master and worker node in a Kubernetes cluster.
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: istio-cni-node-v2.3
```
luksa: Do we want to call this `istio-cni-node-v2-3` instead so that the name and the `k8s-app` label value are consistent?

luksa (changes requested):

pkg/bootstrap/cni.go
yxun: I found a nil case from the cni_test.go
https://github.com/maistra/istio-operator/blob/maistra-2.3/pkg/bootstrap/cni_test.go#L35

In our unit test we called the internalRenderCNI() directly. I can remove my DefaultVersion setting above and then modify the cni_test unit test code.
https://github.com/maistra/istio-operator/blob/maistra-2.3/pkg/bootstrap/cni_test.go#L72

resources/helm/overlays/istio_cni/templates/daemonset-v2.3.yaml
yxun: I checked the versions code and I don't see a case that supportedReleases will be nil as long as ver.Info.MinimumSupportedVersion is not InvalidVersion
https://github.com/maistra/istio-operator/blob/maistra-2.3/pkg/controller/versions/versions.go#L85

and yes, I updated that line with 
{{- if and (has "v2.3" .Values.cni.supportedReleases) (eq .Values.cni.instanceVersion "v2.3") }}
If a future release operator will deprecate v2.3, the first if check will skip rendering that daemonset.

resources/helm/overlays/istio_cni/templates/configmap-v2.3.yaml
```diff
@@ -0,0 +1,26 @@
...
+# This ConfigMap is used to configure a self-hosted Istio CNI installation.
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: istio-cni-config-v2-3
+  namespace: {{ .Release.Namespace }}
+data:
+  # The CNI network configuration to add to the plugin chain on each node.  The special
+  # values in this config will be automatically populated.
+  cni_network_config_v2_3: |-
```
luksa: We should probably change this to just `cni_network_config`, since it's the only entry in this ConfigMap, but we can do this in a follow-up PR.

luksa (approved):
LGTM! Nice job.

resources/helm/overlays/istio_cni/templates/configmap-v2.3.yaml
yxun: I rebased the CM data entry name change in this PR. The only place we set that network config name is in:
https://github.com/maistra/istio-operator/pull/1017/files#diff-fb85c6f4f7143e3ebb7cf98f29325df00f2945710262a52c73b64d790c12ed53R64

jewertow:
I tried to install SMCP from your branch and CNI daemon set fails with the following error:
```
2022-10-04T14:41:56.694363Z	error	Installer exits with open /host/etc/cni/multus/net.d/v2-3-istio-cni.kubeconfig.tmp.3542813523: no such file or directory
2022-10-04T14:41:56.694369Z	info	install	Cleaning up.
2022-10-04T14:41:56.694378Z	info	install	Removing binary: /host/opt/cni/bin/v2-3-istio-cni
Error: open /host/etc/cni/multus/net.d/v2-3-istio-cni.kubeconfig.tmp.3542813523: no such file or directory
```

yxun:
> 
v2-3-istio-cni  that file should be created from the configmap. Thanks, Jaecek. I am checking if that missing file was caused by the last comment above when I renamed the 2.3 configmap data entry.

luksa:
@yxun Why is this on hold? Let's merge it, since it needs to go into 2.3.

luksa:
> > 
> 
> v2-3-istio-cni that file should be created from the configmap. Thanks, Jaecek. I am checking if that missing file was caused by the last comment above when I renamed the 2.3 configmap data entry.

Definitely. I tested everything before you renamed the entry and it worked fine.

yxun:
> @yxun Why is this on hold? Let's merge it, since it needs to go into 2.3.

@luksa Jaecek commented above . He tested and found an error.

yxun:
@jewertow I reverted the configmap data entry name change. If you have a running cluster, could you check it now ?
I am waiting on my cluster and then I will verify it again. Thanks.

jewertow:
Ok, let me test it.

jewertow:
@yxun still the same error...

jewertow:
When I installed SMCP v2.2, then DaemonSet `istio-cni-node-v2-3` started to work.

I noticed that `istio-cni-node-v2-3` did not evaluate `__LOG_UDS_ADDRESS__` and `__KUBECONFIG_FILEPATH__` - it logs:
```
2022-10-04T17:34:25.629064Z     info    CNI install configuration:
CNINetDir: /etc/cni/multus/net.d
MountedCNINetDir: /host/etc/cni/multus/net.d
CNIConfName: v2-3-istio-cni.conf
ChainedCNIPlugin: false
CNINetworkConfigFile:
CNINetworkConfig: {
  "cniVersion": "0.3.1",
  "name": "v2-3-istio-cni",
  "type": "v2-3-istio-cni",
  "log_level": "",
  "log_uds_address": "__LOG_UDS_ADDRESS__",
  "kubernetes": {
      "kubeconfig": "__KUBECONFIG_FILEPATH__",
      "cni_bin_dir": "/opt/multus/bin",
      "exclude_namespaces": [ "istio-operator" ]
  }
}
CNIEnableInstall: true
CNIEnableReinstall: true
LogLevel: warn
KubeconfigFilename: v2-3-istio-cni.kubeconfig
KubeconfigMode: 0600
KubeCAFile:
SkipTLSVerify: false
K8sServiceProtocol:
K8sServiceHost: 10.217.4.1
K8sServicePort: 443
K8sNodeName: istio-cni-node-v2-3-rrp4t
UpdateCNIBinaries: true
SkipCNIBinaries: []
MonitoringPort: 15014
LogUDSAddress: /var/run/istio-cni/log.sock
```
I didn't remove failing daemon set and just updated SMCP version to `v2.2`. Then both daemon sets got up and running, but logs contain evaluated paths:
```
2022-10-04T17:58:33.912648Z	info	install	CNI config: {
  "cniVersion": "0.3.1",
  "name": "v2-3-istio-cni",
  "type": "v2-3-istio-cni",
  "log_level": "",
  "log_uds_address": "/var/run/istio-cni/log.sock",
  "kubernetes": {
      "kubeconfig": "/etc/cni/multus/net.d/v2-3-istio-cni.kubeconfig",
      "cni_bin_dir": "/opt/multus/bin",
      "exclude_namespaces": [ "istio-operator" ]
  }
}
```

yxun:
I can reproduce this error. The only difference in the configmap data. In v2.3, we don't mention any iptable executable or script.

yxun:
It looks like a race condition in the initial v2.3 SMCP creation.
When I deleted the previous SMCP, daemonset and configmap, the next run of v2.3 SMCP creation works fine and all v2-3 daemonset pods running fine now.
I will check the NewManifestProcessor code and add a change if needed.

luksa:
> I can reproduce this error. The only difference in the configmap data. In v2.3, we don't mention any iptable executable or script.

In 2.3 the executable is built into the istio-cni plugin, that's why it does need to be mentioned.

yxun:
The correct log from a running cni-v2-3 daemonset pod

```
2022-10-04T21:53:40.995476Z info Start a UDS server for CNI plugin logs
2022-10-04T21:53:40.995536Z info repair Start CNI race condition repair.
2022-10-04T21:53:41.000283Z info ControlZ available at 127.0.0.1:9876
2022-10-04T21:53:41.109696Z info install Copied istio-cni to /host/opt/cni/bin/v2-3-istio-cni.
2022-10-04T21:53:41.109729Z info install Directory /host/secondary-bin-dir is not writable, skipping.
2022-10-04T21:53:41.109922Z info install write kubeconfig file /host/etc/cni/multus/net.d/v2-3-istio-cni.kubeconfig with:
# Kubeconfig file for Istio CNI plugin.
apiVersion: v1
kind: Config
...
```

The error cni pod log from a clean cluster initial run

```
2022-10-04T14:41:56.694363Z	error	Installer exits with open /host/etc/cni/multus/net.d/v2-3-istio-cni.kubeconfig.tmp.3542813523: no such file or directory
2022-10-04T14:41:56.694369Z	info	install	Cleaning up.
2022-10-04T14:41:56.694378Z	info	install	Removing binary: /host/opt/cni/bin/v2-3-istio-cni
Error: open /host/etc/cni/multus/net.d/v2-3-istio-cni.kubeconfig.tmp.3542813523: no such file or directory
```

"repair Start CNI race condition repair."  There is a race condition and it triggers a StartRepair step in :
https://github.com/maistra/istio/blob/maistra-2.3/cni/pkg/repair/repair.go#L226

yxun:
> __KUBECONFIG_FILEPATH__  and __LOG_UDS_ADDRESS__
We didn't set those two in our cni.go renderCNI function. In previous releases, they were the same place holders in the configmap

luksa:
> I noticed that `istio-cni-node-v2-3` did not evaluate `__LOG_UDS_ADDRESS__` and `__KUBECONFIG_FILEPATH__` - it logs:

@jewertow Are you saying that the actual config file contained these two placeholders or that you saw them only in the log? It's normal for this to be in the log statement that reads `CNINetworkConfig: {`. The CNI installer running in the DaemonSet is the one that replaces the placeholders with actual values when it writes the file (the file contents are printed in the log statement that reads `install	CNI config: {`. 

So, there's no issue here.

luksa:
@yxun, @jewertow I can also reproduce this. I know what the problem is and I'll have a fix ready in a few minutes.

luksa:
The fix for the issue is in https://github.com/maistra/istio/pull/638

luksa:
@yxun You can change the config map entry key back to `cni_network_config`. And while you're at it, I suggest also changing the container name in the 2.3 DaemonSet from `install-cni-v2-3` to just `install-cni`.

luksa:
@yxun okay to merge?

yxun:
> @yxun okay to merge?

ok

PR #1019, opened 2022-09-30
OSSM-2069 Configure kiali for cluster-wide operation when SMMR targets all namespaces

luksa:


pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -396,13 +400,25 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
 }
 
 func isExcludedNamespace(ns string) bool {
+	// keep this in sync with getExcludedNamespaces()
```
jwendell: perhaps this one could be refactored to sth like `return ns in  getExcludedNamespaces()`

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -396,13 +400,25 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
...
 		ns == "openshift" ||
 		strings.HasPrefix(ns, "kube-") ||
 		strings.HasPrefix(ns, "openshift-")
 }
 
+func getExcludedNamespaces() []string {
+	// keep this in sync with isExcludedNamespace()
+	return []string{
+		common.GetOperatorNamespace(),
+		"kube",
```
jwendell: Kiali seems to use regex for this set: https://github.com/kiali/kiali-operator/blob/master/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml#L22

Do we need to use `^` too?

jwendell (approved):
LGTM, just minor nits.

Do you know if, just by setting namespaces to `**`, Kiali automatically configures the exclude set? Or this is something we should manually configure as in this PR?

pkg/controller/servicemesh/memberroll/controller.go
luksa: Good catch. Wonder why it seemed to work correctly when I manually tested this with Kiali. Anyway, I've now fixed this.

pkg/controller/servicemesh/memberroll/controller.go
luksa: I've now done this, since getExcludedNamespaces() returns regex patterns.

luksa:
@jwendell please review again.

jwendell (approved):

pkg/controller/servicemesh/memberroll/controller.go
jwendell: It worked because it was a match. If we don't specify the `^` we are matching any part of the string.

My other question was if this is necessary or if Kiali takes care of this when namespaces is set to `**`.

luksa: I mean, it correctly ignored all `openshift-xyz` namespaces, even though the regex was wrong (`"openshift-*"` instead of `"openshift-.*"`).

Kiali doesn't take care of this. You have to explicitly set excluded namespaces when you use `**`.

luksa: Oh, I just realized why it worked.. because `openshift-xyz` matches the regex `openshift`, which was also in there.

PR #1028, opened 2022-10-07
OSSM-1094 Use SHA-256 for htpasswd

cam-garrison:
Update htpasswd secret to be created with SHA-256 rather than SHA-1. 

Tested with SMCP versions 2.1, 2.2, 2.3 using [this bookinfo test  ](https://istio.io/latest/docs/tasks/observability/metrics/querying-metrics/).

### DIFF
--- a/pkg/controller/servicemesh/controlplane/htpasswd.go
+++ b/pkg/controller/servicemesh/controlplane/htpasswd.go
@@ -34,7 +34,7 @@ func (r *controlPlaneInstanceReconciler) patchHtpasswdSecret(ctx context.Context
 		}
 		h := sha256.New()
 		h.Write([]byte(rawPassword))
-		auth = "internal:{SHA}" + base64.StdEncoding.EncodeToString(h.Sum(nil))
+		auth = "internal:{SHA-2}" + base64.StdEncoding.EncodeToString(h.Sum(nil))
 	}
 
 	b64Password := base64.StdEncoding.EncodeToString([]byte(rawPassword))
--- a/pkg/controller/servicemesh/controlplane/htpasswd.go
+++ b/pkg/controller/servicemesh/controlplane/htpasswd.go
@@ -3,7 +3,7 @@ package controlplane
 import (
 	"context"
 	"crypto/rand"
-	"crypto/sha1"
+	"crypto/sha256"
 	"encoding/base64"
 	"fmt"
 	"regexp"
@@ -32,7 +32,7 @@ func (r *controlPlaneInstanceReconciler) patchHtpasswdSecret(ctx context.Context
 			log.Error(err, "failed to generate the HTPasswd password")
 			return err
 		}
-		h := sha1.New()
+		h := sha256.New()
 		h.Write([]byte(rawPassword))
 		auth = "internal:{SHA}" + base64.StdEncoding.EncodeToString(h.Sum(nil))
 	}


pkg/controller/servicemesh/controlplane/htpasswd.go
```diff
@@ -32,9 +32,9 @@ func (r *controlPlaneInstanceReconciler) patchHtpasswdSecret(ctx context.Context
 			log.Error(err, "failed to generate the HTPasswd password")
 			return err
 		}
-		h := sha1.New()
+		h := sha256.New()
 		h.Write([]byte(rawPassword))
-		auth = "internal:{SHA}" + base64.StdEncoding.EncodeToString(h.Sum(nil))
+		auth = "internal:{SHA-2}" + base64.StdEncoding.EncodeToString(h.Sum(nil))
```
jwendell, line 37: Is this correct?

When I run `htpasswd` locally, I see a different output (`-2` means sha256 and `-s` means sha-1):
```
[jwendell@jw-p1 h]$ htpasswd -n -2 -b internal random_password
internal:$5$RY8UyNHMWKk6b41z$kl2uJdF1Ns5xAeIRz8CrBbsMlESWhX3MgyxShG2fDD1

[jwendell@jw-p1 h]$ htpasswd -n -s -b internal random_password
internal:{SHA}1q7M/obgg9a8tcLtJFm4p+T0n+w=
```

jwendell (commented):
Have you tested if grafana opens ok?
The [oauth proxy doc](https://github.com/openshift/oauth-proxy) says to use sha-1, so, I'm not sure if it would work with sha-256.

pkg/controller/servicemesh/controlplane/htpasswd.go
luksa: +1

`$5$` seems to be the correct prefix for SHA-256 passwords (see [here](https://github.com/tg123/go-htpasswd/blob/v1.2.0/cryptsha.go#L20), for example)

cam-garrison: Thank you for pointing that out, I wasn't sure where to find the standard prefix. I will amend that in a new commit. Testing grafana now.

jwendell:
I don't think oauth supports 256: https://github.com/openshift/oauth-proxy/blob/master/htpasswd.go#L15 (and see the Validate() method on this file.

I think we should close that issue as invalid.

cam-garrison:
> I don't think oauth supports 256: https://github.com/openshift/oauth-proxy/blob/master/htpasswd.go#L15 (and see the Validate() method on this file.
> 
> I think we should close that issue as invalid.

Agree now after testing Grafana with this change, there are indeed login problems to access the Grafana dashboard. Thought prometheus was the one affected but you were correct. Will close the issue on jira!

luksa:
The code shows that bcrypt is supported. Why not use it instead?

cam-garrison:
> The code shows that bcrypt is supported. Why not use it instead?

Good idea, let me re-open the issue and look into using bcrypt.

luksa:
You can just reopen the existing issue, since it didn't ask for SHA256. It just asked to replace SHA1 with something else.

jwendell:
I still don't know what is the benefit of such change, TBH.
We store the raw password in the same secret, so, what are we trying to protect here?
This htpasswd file is only used internally, and not exposed/consumed by end users...

PR #1030, opened 2022-10-11

luksa:


pkg/bootstrap/cni.go
```diff
@@ -82,6 +83,11 @@ func internalRenderCNI(ctx context.Context, cl client.Client, config cni.Config,
 	values := make(map[string]interface{})
 	values["cni"] = cni
 
+	serverVersion, err := dc.ServerVersion()
```
jwendell, line 86: Do you know if the result is cached so that next calls to `ServerVersion()` don't reach the api server?

We are now invoking this function on every reconcile, right? Do you think this is an issue if it's not cached? The server version cannot change while the operator is alive, can it?

jwendell (approved):

pkg/bootstrap/cni.go
luksa: Unfortunately, the server version can change while the operator is alive. I'm not sure if the discoveryClient that is passed here is the cachedDiscoveryClient or not. 

However, even if we're hitting the API server every time, it's not that big of a deal, because we perform far more requests that update the resources owned by the SMCP.

rcernich:
Does this need to go into 2.3?

luksa:
It's already in 2.3.

PR #1031, opened 2022-10-11

luksa:


luksa:
Closed due to https://github.com/maistra/istio-operator/pull/1032#issuecomment-1277382000

PR #1032, opened 2022-10-12

luksa:


jwendell:
Do we need to update 2.0 and 2.1 branches of operator?

luksa:
@jwendell WDYM? This is a cherry-pick of #1012 that I'm now applying to all previous branches (see also #1030 and #1031).

jwendell:
I meant.. it's unlikely we will release a new 2.0.x operator, right? People should use only the latest (2.2.x ATM) version of the operator, right?

luksa:
Ah, I see what you mean. Yes, I wasn't sure, so I submitted PRs for all previous versions.

PR #1035, opened 2022-10-14
OSSM-1321: Add permissions to update pods/finalizers required for appending validation messages

jewertow:
This change fixes errors logged by istiod when `validationMessages` option is enabled.


luksa (approved):

jwendell:
I thought we wouldn't update older operator...

luksa:
This PR is only a chart update, so that the 2.2 and 2.3 operators will be able to deploy 2.1 properly.

This is also another reason to move the charts from the operator repo to maistra/istio. The PR makes it look as if the intent is to update the 2.1 operator, while we're actually updating the 2.1 charts. If these charts were in maistra/istio and the 2.3 operator pulled the charts directly from maistra/istio and not from the previous branches in maistra/istio-operator, things would make more sense.

PR #1047, opened 2022-10-20
feat(make): simplifies targets for creating charts and other resources

bartoszmajsak:
Instead of duplicating targets for each major version it uses target template where any version can be passed as target's suffix. It also honors special case for creating charts for current version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ SUPPORTED_VERSIONS := 2.0 2.1 2.2 2.3
 
 $(addprefix update-remote-maistra-,$(SUPPORTED_VERSIONS)): update-remote-maistra-%:
 	$(eval version:=$*)
-	@ if [[ ${OFFLINE_BUILD} == "false" || ${MAISTRA_VERSION} == ${version}* ]]; \
+	@ if [[ ${OFFLINE_BUILD} == "false" || ${MAISTRA_VERSION} == ${version}.* ]]; \
 	then \
 		git remote set-branches --add ${GIT_UPSTREAM_REMOTE} maistra-${version}; \
 		git fetch ${GIT_UPSTREAM_REMOTE} maistra-${version}:maistra-${version}; \
@@ -101,7 +101,7 @@ $(addprefix update-charts-,$(SUPPORTED_VERSIONS)): update-charts-%:
 	$(eval version:=$*)
 	@# If we are calling make against current version - download charts.
 	@# Otherwise sync from previous branches and explicitly call dependent target with extracted version
-	@ if [[ ${MAISTRA_VERSION} == ${version}* ]]; \
+	@ if [[ ${MAISTRA_VERSION} == ${version}.* ]]; \
 	then \
 		HELM_DIR=${RESOURCES_DIR}/helm/v${version} ISTIO_VERSION=${ISTIO_VERSION} ${SOURCE_DIR}/build/download-charts.sh; \
 	else \
--- a/Makefile
+++ b/Makefile
@@ -12,10 +12,15 @@
 ## See the License for the specific language governing permissions and
 ## limitations under the License.
 
+# Determine this makefile's path.
+# Be sure to place this BEFORE `include` directives, if any.
+THIS_FILE := $(lastword $(MAKEFILE_LIST))
+
 -include Makefile.overrides
 
 MAISTRA_VERSION        ?= 2.3.0
 MAISTRA_BRANCH         ?= maistra-2.3
+ISTIO_VERSION          ?= 1.14.3
 REPLACES_PRODUCT_CSV   ?= 2.2.3
 REPLACES_COMMUNITY_CSV ?= 2.2.3
 VERSION                ?= development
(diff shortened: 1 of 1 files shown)


bartoszmajsak:
Will assign reviewers after fixing the build (worked on my machine ;))

luksa (approved):

Makefile
```diff
@@ -79,115 +84,51 @@ test:
...
 
 ################################################################################
-# maistra v2.0
+# Helm charts generation
 ################################################################################
-.PHONY: update-remote-maistra-2.0
-update-remote-maistra-2.0:
+.PHONY: update-remote-maistra-%
+update-remote-maistra-%:
+	$(eval version:=$(subst update-remote-maistra-,,$@))
```
jwendell: perhaps you can use `$*` for this purpose (applies everywhere is this pr)

Makefile
```diff
@@ -79,115 +84,51 @@ test:
...
-.PHONY: collect-2.3-charts
-collect-2.3-charts:
+.PHONY: update-charts-%
+update-charts-%:
+	$(eval version:=$(subst update-charts-,,$@))
+	# If we are calling make against current version - download charts.
+	# Otherwise sync from previous branches and explicitly call dependent target with extracted version
+	@ if [[ ${MAISTRA_VERSION} == ${version}* ]]; \
+	then \
+		echo "test"; \
```
jwendell: remove this?

Makefile
```diff
@@ -79,115 +84,51 @@ test:
...
+		HELM_DIR=${RESOURCES_DIR}/helm/v${version} ISTIO_VERSION=${ISTIO_VERSION} ${SOURCE_DIR}/build/download-charts.sh; \
+	else \
+		$(MAKE) -f $(THIS_FILE) update-remote-maistra-${version}; \
+		git checkout ${GIT_UPSTREAM_REMOTE}/maistra-${version} -- ${SOURCE_DIR}/resources/helm/v${version}; \
+		git reset HEAD ${SOURCE_DIR}/resources/helm/v${version}; \
```
jwendell, line 109: do you need `--hard`?

bartoszmajsak: Didn't think about the code - it was just there :) Do you think it would help us avoid some problems?

Makefile
bartoszmajsak:
```suggestion
```

Makefile
bartoszmajsak: Good idea! Thanks. Will do.

Makefile
jwendell: nah, nevermind

Makefile
bartoszmajsak: @jwendell Done in 25845f47

Makefile
```diff
@@ -79,115 +84,50 @@ test:
...
-.PHONY: update-2.3-charts
-update-2.3-charts:
-	HELM_DIR=${RESOURCES_DIR}/helm/v2.3 ISTIO_VERSION=1.14.3 ${SOURCE_DIR}/build/download-charts.sh
-
-.PHONY: collect-2.3-charts
-collect-2.3-charts:
+.PHONY: update-charts-%
+update-charts-%:
+	$(eval version:=$*)
+	# If we are calling make against current version - download charts.
```
jwendell: nit: Could you prepend those comments with `@` so that they aren't printed on the screen?

jwendell (commented):
LGTM.

We lost the ability to fail fast if passing an invalid version in the target. For instance:

```sh
$ make update-templates-9.1
Building maistra operator
# Has to explicitly call dependent target with extracted version
make[1]: Entering directory '/home/jwendell/src/maistra/istio-operator'
Building maistra operator
git remote set-branches --add origin maistra-9.1
git fetch origin maistra-9.1:maistra-9.1
fatal: couldn't find remote ref maistra-9.1
make[1]: *** [Makefile:104: update-remote-maistra-9.1] Error 128
make[1]: Leaving directory '/home/jwendell/src/maistra/istio-operator'
make: *** [Makefile:126: update-templates-9.1] Error 2
```

We could workaround this by defining a validating function like:
```makefile
ALL_VERSIONS := 2.0 2.1 2.2 2.3

define validate_version
  ifeq ($(filter $(1),$(ALL_VERSIONS)),)
    $$(error "$(1)" is not a valid version. Valid values: $(ALL_VERSIONS))
  endif
endef

validate-version-%:
	$(eval $(call validate_version,$*))
```

And then having it as a dependency on the targets, e.g.
```makefile
update-templates-%: validate-version-%
...
```
Which results in:
```sh
$ make update-templates-9.1
Building maistra operator
Makefile:41: *** "9.1" is not a valid version. Valid values: 2.0 2.1 2.2 2.3.  Stop.
```

BTW, we could move the message `Building maistra operator` to the build target, it's annoying :)

bartoszmajsak:
@jwendell thanks for the thorough review. I learned a few things! 

As for validation - that's correct, I was mainly thinking about avoiding copy&paste in the Makefile. your suggestion could actually be used to replace the target which calls version-specific ones instead of validating. I am on it. It also solves another problem in version matching.

thank-you

Makefile
bartoszmajsak: Done in 9af7e20b

jewertow (approved):

Makefile
```diff
@@ -79,115 +84,57 @@ test:
...
-.PHONY: collect-2.2-charts
-collect-2.2-charts:
+# Helm charts generation and templates processing
+################################################################################
+
+SUPPORTED_VERSIONS := 2.0 2.1 2.2 2.3
+
+$(addprefix update-remote-maistra-,$(SUPPORTED_VERSIONS)): update-remote-maistra-%:
+	$(eval version:=$*)
+	@ if [[ ${OFFLINE_BUILD} == "false" || ${MAISTRA_VERSION} != ${version}.* ]]; \
```
bartoszmajsak:
```suggestion
	@ if [[ ${OFFLINE_BUILD} == "false" && ${MAISTRA_VERSION} != ${version}.* ]]; \
```

PR #1048, opened 2022-10-20
OSSM-2170 Ensure ServiceMeshMembers are deleted when SMMR is cluster-scoped

luksa:
This also fixes edge cases where the SMMR.spec.members list didn't reflect the actual SMM objects. For example, when the user removes namespaces from the SMMR.spec.members and then immediately deletes the SMMR. In that case, the controller didn't delete the SMM when finalizing the SMMR, because the SMMR.spec.members list was empty.

### DIFF
--- a/pkg/controller/servicemesh/memberroll/controller.go
+++ b/pkg/controller/servicemesh/memberroll/controller.go
@@ -555,16 +555,19 @@ func (r *MemberRollReconciler) ensureMemberExists(ctx context.Context, ns, meshN
 }
 
 func (r *MemberRollReconciler) finalizeObject(ctx context.Context, roll *maistrav1.ServiceMeshMemberRoll) (ok bool, err error) {
-	for _, ns := range roll.Spec.Members {
-		member := maistrav1.ServiceMeshMember{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      common.MemberName,
-				Namespace: ns,
-			},
+	members := &maistrav1.ServiceMeshMemberList{}
+	err = r.Client.List(ctx, members, client.MatchingFields{"spec.controlPlaneRef.namespace": roll.Namespace})
+	if err != nil {
+		return false, err
+	}
+
+	for _, member := range members.Items {
+		if member.DeletionTimestamp != nil || !isCreatedByThisController(&member) {
+			continue
 		}
 		err := r.Client.Delete(ctx, &member)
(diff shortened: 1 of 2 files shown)


pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -555,16 +555,19 @@ func (r *MemberRollReconciler) ensureMemberExists(ctx context.Context, ns, meshN
...
-				Namespace: ns,
-			},
+	members := &maistrav1.ServiceMeshMemberList{}
+	err = r.Client.List(ctx, members, client.MatchingFields{"spec.controlPlaneRef.namespace": roll.Namespace})
+	if err != nil {
+		return false, err
+	}
+
+	for _, member := range members.Items {
+		if member.DeletionTimestamp != nil || !isCreatedByThisController(&member) {
```
jwendell, line 565: Should this be `== nil`?

luksa: I do have what I call booleanitis, where I inadvertently flip boolean expressions, but in this case, `skip deletion if deletionTimestamp is set` is correct, considering that when `deletionTimestamp` is set, the member was already deleted and doesn't need to be deleted again.

jwendell: got it, thanks

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -555,16 +555,19 @@ func (r *MemberRollReconciler) ensureMemberExists(ctx context.Context, ns, meshN
...
+			continue
 		}
 		err := r.Client.Delete(ctx, &member)
 		if err != nil && !errors.IsNotFound(err) {
-			return false, pkgerrors.Wrapf(err, "Could not delete ServiceMeshMember %s/%s", ns, common.MemberName)
```
jwendell, line 567: Does it make sense not return early and instead try to delete all members and return an error saying which members could not be deleted?

luksa: The only reason a delete would fail is if there was an error communicating with the API server. I think we should actually change our controllers so they always fail fast. Collecting errors in an array just makes the code less readable for no real benefit. Either the controller does its job to completion or it has to retry.

jwendell: makes sense, thanks

jwendell (approved):

PR #1049, opened 2022-10-27
OSSM-2187 Allow usage of spec.proxy.networking.protocol.autoDetect

dgn:
This removes the 'validation' that was effectively removing support for protocol sniffing and allows usage of the fields in 2.2 and 2.3 control planes. [We previously removed this feature from OSSM 2.0](https://github.com/maistra/istio-operator/pull/614/commits/31bb5cb9b1d335196086baedd24c2d00a3da5a61) because of security concerns, however I cannot find any documentation of these concerns and upstream has been enabling it by default since Istio 1.6, so I consider it safe enough to allow usage. Note that we will still default to disabling it in order not to change behavior in any existing deployments.

We might want to discuss enabling it by default in 2.4

### DIFF
--- a/pkg/apis/maistra/conversion/proxy.go
+++ b/pkg/apis/maistra/conversion/proxy.go
@@ -138,9 +138,6 @@ func populateProxyValues(in *v2.ControlPlaneSpec, values map[string]interface{})
 		if proxy.Networking.Protocol != nil && proxy.Networking.Protocol.AutoDetect != nil {
 			autoDetect := proxy.Networking.Protocol.AutoDetect
 			if autoDetect.Timeout != "" {
-				if err := setHelmStringValue(proxyValues, "protocolDetectionTimeout", autoDetect.Timeout); err != nil {
-					return err
-				}
 				if err := setHelmStringValue(meshConfigValues, "protocolDetectionTimeout", autoDetect.Timeout); err != nil {
 					return err
 				}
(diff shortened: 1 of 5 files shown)


luksa (approved):

jewertow (approved):
I tested your change on OCP and I can confirm that it worked. After setting protocol auto detection, istiod had updated env vars:
```
- name: PILOT_ENABLE_PROTOCOL_SNIFFING_FOR_OUTBOUND
  value: 'true'
- name: PILOT_ENABLE_PROTOCOL_SNIFFING_FOR_INBOUND
  value: 'true'
```
When I set autoDetect.timeout then istiod updated connected proxies.

There is only one thing that I would change.
```
if autoDetect.Timeout != "" {
        if err := setHelmStringValue(proxyValues, "protocolDetectionTimeout", autoDetect.Timeout); err != nil {
	        return err
	}
	if err := setHelmStringValue(meshConfigValues, "protocolDetectionTimeout", autoDetect.Timeout); err != nil {
		return err
	}
}
```
Setting `proxyValues` has no effect, because istiod respects only MeshConfig, so maybe we could remove it? It works anyway, so you can ignore this suggestion, but this PR is an opportunity to simplify this piece of code.

dgn:
> There is only one thing that I would change.
> 
> ```
> if autoDetect.Timeout != "" {
>         if err := setHelmStringValue(proxyValues, "protocolDetectionTimeout", autoDetect.Timeout); err != nil {
> 	        return err
> 	}
> 	if err := setHelmStringValue(meshConfigValues, "protocolDetectionTimeout", autoDetect.Timeout); err != nil {
> 		return err
> 	}
> }
> ```
> 
> Setting `proxyValues` has no effect, because istiod respects only MeshConfig, so maybe we could remove it? It works anyway, so you can ignore this suggestion, but this PR is an opportunity to simplify this piece of code.

Good catch Jacek! Thank you. I removed that code

PR #1052, opened 2022-11-02
OSSM-534 Support ingress.hosts for grafana and prometheus

dgn:


resources/helm/overlays/istio-telemetry/grafana/templates/ingress.yaml
```diff
@@ -1,21 +1,29 @@
...
-      {{ $key }}: {{ $value | quote }}
-    {{- end }}
+{{- range $key, $value := $.Values.grafana.ingress.annotations }}
+    {{ $key }}: {{ $value | quote }}
+{{- end }}
+{{- end }}
 spec:
+{{- if ne $host "grafana.local" }}
+  host: {{ $host }}
+{{- end }}
```
luksa: Found the following bug when I manually tested this. If I create an SMCP with:
```
spec:
  addons:
    grafana:
      enabled: true
      install:
        service:
          ingress:
            enabled: true
            hosts:
            - grafana-host.example.com
```
a route with the host `grafana-host.example.com` is created as expected. However, if I then edit the SMCP and remove the `hosts` field, the host in the route is still `grafana-host.example.com`. This Route doesn't match the one that's created if I initially create the SMCP without the `hosts` field:
```
spec:
  addons:
    grafana:
      enabled: true
      install:
        service:
          ingress:
            enabled: true
```
In this case, the host in the Route is `grafana-istio-system.apps.mluksa-mesh1.servicemesh....com`. 

Since the input state (SMCP spec) during the reconciliation of the SMCP is the same in both cases, the output (the Route) should also be the same, regardless of how we got to the input state.

The reason this happens is because you don't set the `host` field here, which means that the PATCH operation doesn't update it.

luksa: I've been trying to find a solution for this, but it looks like there's no way to remove the host. I tried setting it to `nil`, `''`, and removing it with JSONPatch (`op:remove`). Nothing worked.

So, it looks like the only way to fix this would be to delete and recreate the Route.

dgn: Hmm. That's a problem. We also can't default to a `recreate` strategy for Route objects in general as that would change the generated host name

dgn: Okay, I added some code to the Patch Factory so we're now recreating Route objects if the `spec.host` changed

pkg/controller/common/helm/patch.go
```diff
@@ -35,6 +37,17 @@ type basicPatch struct {
...
+		routeHostChanged(p.oldObj, p.newObj) {
+		return nil, &errors.StatusError{
+			ErrStatus: metav1.Status{
+				Status:  metav1.StatusFailure,
+				Reason:  metav1.StatusReasonInvalid,
+				Message: "spec.host of Route objects is immutable",
+			},
+		}
+	}
+
```
luksa: To minimize the possibility of this breaking something, wouldn't it be better to only return the error when `.spec.host` is changed from non-empty to empty/nil?

dgn: in fact, the `Patch()` would always fail anyway (server-side) when changing `spec.host` (because it's immutable), triggering a recreate. this just lets it fail earlier.

luksa: Are you sure it's immutable? I tried changing an existing route's host with `kubectl edit` after I saw the error message above, and it worked.

dgn: Okay, that's strange. I have to test it again, but it's documented as "may not be changed": https://github.com/openshift/api/blob/803c45de7ab5567e8d1575138f014226974768a1/route/v1/types.go#L21-L22

dgn: Oh you are right. You can change it. Weird, I was somehow sure that `spec.host` can't be changed.

Another thing I noticed is that the auto-generated `host` value is actually written to the `spec` (likely using a MutatingWebhook), and an annotation is added: `openshift.io/host.generated: "true"`. So we might have to check that annotation and remove it

luksa: I don't think we need to update the annotation. OpenShift should do this (there was a [TODO](https://github.com/openshift/origin/blob/95117a2f8866f369b1ac2828ee520975c09593bc/pkg/route/registry/route/rest.go#L136) about this, but I guess they never actually did it.

AFAICT that annotation is only used to show the route host as being auto-generated in the OpenShift console, so it's not a big deal if we leave it as is.

luksa (approved):

PR #1055, opened 2022-11-04
OSSM-2190 OSSM-2232 OSSM-2189 Don't reconcile invalid objects

luksa:
These three commits ensure that the three controllers (SMCP, SMMR, and SMM) don't reconcile objects that shouldn't even exist. 

The validation webhook prevents the creation of these objects, but you can create them if you first delete the webhook. If this happens, the controller must ignore them.

### DIFF
--- a/pkg/controller/servicemesh/controlplane/reconciler.go
+++ b/pkg/controller/servicemesh/controlplane/reconciler.go
@@ -81,6 +81,17 @@ func NewControlPlaneInstanceReconciler(controllerResources common.ControllerReso
 func (r *controlPlaneInstanceReconciler) Reconcile(ctx context.Context) (result reconcile.Result, err error) {
 	log := common.LogFromContext(ctx)
 	log.Info("Reconciling ServiceMeshControlPlane", "Status", r.Instance.Status.StatusType)
+
+	earliestSmcp, err := getEarliestSMCPInNamespace(ctx, r.Client, r.Instance.Namespace)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if earliestSmcp != nil && r.Instance.UID != earliestSmcp.UID {
+		return reconcile.Result{}, r.postReconciliationStatus(ctx, status.ConditionReasonMultipleSMCPs,
+			"reconciliation skipped",
+			fmt.Errorf("multiple ServiceMeshControlPlane resources exist in the namespace"))
+	}
+
 	if r.Status.GetCondition(status.ConditionTypeReconciled).Status != status.ConditionStatusFalse {
 		r.initializeReconcileStatus()
 		err := r.PostStatus(ctx)
@@ -339,6 +350,29 @@ func (r *controlPlaneInstanceReconciler) Reconcile(ctx context.Context) (result 
 	hacks.SkipReconciliationUntilCacheSynced(ctx, common.ToNamespacedName(r.Instance))
 	log.Info("Completed ServiceMeshControlPlane reconciliation")
 	return
+}
+
+// finds the first SMCP in the given namespace by creationTimestamp
+func getEarliestSMCPInNamespace(ctx context.Context, cl client.Client, ns string) (*v2.ServiceMeshControlPlane, error) {
+	smcpList := &v2.ServiceMeshControlPlaneList{}
+	if err := cl.List(ctx, smcpList, client.InNamespace(ns)); err != nil {
+		return nil, err
+	}
+
+	var earliest *v2.ServiceMeshControlPlane
+	for _, smcp := range smcpList.Items {
+		if earliest == nil ||
+			smcp.CreationTimestamp.Before(&earliest.CreationTimestamp) ||
+			(smcp.CreationTimestamp == earliest.CreationTimestamp && smcp.Name < earliest.Name) {
+
+			// We can't just do `earliest = &smcp`, because of how for loops are implemented in Go.
+			// Instead, we must create a new variable and store its address in `earliest`.
+			// See discussion at https://github.com/golang/go/discussions/56010
+			goPitfallWorkaround := smcp
+			earliest = &goPitfallWorkaround
+		}
+	}
+	return earliest, nil
 }
 
 func (r *controlPlaneInstanceReconciler) pauseReconciliation(ctx context.Context) (status.ConditionReason, string) {
--- a/pkg/controller/servicemesh/member/controller.go
+++ b/pkg/controller/servicemesh/member/controller.go
@@ -196,6 +196,12 @@ func (r *MemberReconciler) Reconcile(request reconcile.Request) (reconcile.Resul
 
 func (r *MemberReconciler) reconcileObject(ctx context.Context, member *maistrav1.ServiceMeshMember) (reconcile.Result, error) {
 	log := common.LogFromContext(ctx)
+
+	if member.Name != common.MemberName {
+		return reconcile.Result{}, r.reportError(ctx, member, maistrav1.ConditionReasonMemberNameInvalid,
+			fmt.Errorf("the ServiceMeshMember name is invalid; must be %q", common.MemberName))
+	}
+
 	if member.Namespace == member.Spec.ControlPlaneRef.Namespace {
 		return reconcile.Result{}, nil
 	}
--- a/pkg/controller/servicemesh/memberroll/controller.go
+++ b/pkg/controller/servicemesh/memberroll/controller.go
@@ -247,6 +247,13 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
 
 	meshNamespace := roll.Namespace
 
+	if roll.Name != common.MemberRollName {
+		log.Info("Skipping reconciliation of SMMR with invalid name", "project", meshNamespace, "name", roll.Name)
+		setReadyCondition(roll, false, maistrav1.ConditionReasonInvalidName,
+			fmt.Sprintf("the ServiceMeshMemberRoll name is invalid; must be %q", common.MemberRollName))
+		return reconcile.Result{}, r.updateStatus(ctx, roll)
+	}
+
 	// 1. gather status of all members that belong to this roll
 	members := &maistrav1.ServiceMeshMemberList{}
 	err := r.Client.List(ctx, members, client.MatchingFields{"spec.controlPlaneRef.namespace": meshNamespace})
--- a/pkg/apis/maistra/status/status.go
+++ b/pkg/apis/maistra/status/status.go
@@ -55,7 +55,7 @@ func NewStatus() StatusType {
 }
 
 type ComponentStatusList struct {
-	//+optional
+	// +optional
 	ComponentStatus []ComponentStatus `json:"components,omitempty"`
 }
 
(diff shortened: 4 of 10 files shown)


pkg/controller/servicemesh/member/controller.go
```diff
@@ -196,6 +196,12 @@ func (r *MemberReconciler) Reconcile(request reconcile.Request) (reconcile.Resul
 
 func (r *MemberReconciler) reconcileObject(ctx context.Context, member *maistrav1.ServiceMeshMember) (reconcile.Result, error) {
 	log := common.LogFromContext(ctx)
+
+	if member.Name != common.MemberName {
+		return reconcile.Result{}, r.reportError(ctx, member, maistrav1.ConditionReasonMemberNameInvalid,
+			fmt.Errorf("the ServiceMeshMember name is invalid; must be 'default'"))
```
jwendell: nit: Use a constant for this and tests usages?

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -247,6 +247,13 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
 
 	meshNamespace := roll.Namespace
 
+	if roll.Name != common.MemberRollName {
+		log.Info("Skipping reconciliation of SMMR with invalid name", "project", meshNamespace, "name", roll.Name)
+		setReadyCondition(roll, false, maistrav1.ConditionReasonInvalidName,
+			"the ServiceMeshMemberRoll name is invalid; must be 'default'")
```
jwendell: nit: constant as well?

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -341,6 +352,25 @@ func (r *controlPlaneInstanceReconciler) Reconcile(ctx context.Context) (result
...
+	smcpList := &v2.ServiceMeshControlPlaneList{}
+	if err := cl.List(ctx, smcpList, client.InNamespace(ns)); err != nil {
+		return nil, err
+	}
+
+	var earliest *v2.ServiceMeshControlPlane
+	for _, smcp := range smcpList.Items {
+		if earliest == nil ||
+			smcp.CreationTimestamp.Before(&earliest.CreationTimestamp) ||
+			(smcp.CreationTimestamp == earliest.CreationTimestamp && smcp.Name < earliest.Name) {
```
jwendell, line 366: What's the use case here? Is this even possible? (to have two equal timestamps for the same resource)
Why the name is important here?

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -341,6 +352,25 @@ func (r *controlPlaneInstanceReconciler) Reconcile(ctx context.Context) (result
...
+			goPitfallWorkaround := smcp
+			earliest = &goPitfallWorkaround
```
jwendell, line 372: Really? :o

pkg/controller/servicemesh/controlplane/reconciler.go
jewertow: > Why the name is important here?

I think that when timestamps are equal, you can't determine which one should be reconciled, so name is an additional parameter for selecting the right object.

pkg/controller/servicemesh/controlplane/reconciler.go
jewertow, line 372: Since `smcp` is not a pointer, but a value, why do you need this temporary variable?

pkg/controller/servicemesh/controlplane/reconciler.go
jwendell: but name should not matter. alphabetical order doesn't mean anything. Or does it?

luksa: @jewertow is right. The resolution of `creationTimestamp` is 1 second, so you can easily have two SMCPs with the same timestamp. To make sure that the code always selects the same SMCP as the "earliest", I "sort" them by their names.

pkg/controller/servicemesh/controlplane/reconciler.go
luksa: Yes. Go sucks.

luksa: Because the smcp variable gets reused in each step of the for loop (it's like being declared outside of the for loop). So, storing the address of this variable in `earliest` will result in `earliest` always pointing to the last smcp in the loop.

For example, let's say you have 2 smcps that you're iterating over. In the first step, you set `earliest` to the address of the smcp variable (which at that point holds the 1st smcp). At the beginning of the second step, the value in the smcp variable is overridden with the 2nd smcp. Even though you don't update `earliest` in the second step, it will point to the a variable that holds the value of the 2nd smcp.

So, to have `earliest` point to the correct value, you need to create a new variable (`goPitfallWorkaround` in this case) and get its address. 

The go authors plan to change this in a future release.

luksa: See discussion here: https://github.com/golang/go/discussions/56010

jewertow: Yes, but I think you could avoid it by defining `earliest` as `v2.ServiceMeshControlPlane` instead of `*v2.ServiceMeshControlPlane`.

pkg/controller/servicemesh/member/controller.go
luksa: Absolutely! Thanks!

pkg/controller/servicemesh/memberroll/controller.go
luksa: Done. Thanks.

pkg/controller/servicemesh/controlplane/reconciler.go
jwendell: ok, why not use random() instead of going alphabetical. What if 1st name is `b` and second name is `a`?

jwendell: I mean, random() is a joke... I'm just trying to understand why we have this comparison at all. If timestamps are equal, pick any of those.

jwendell: We can't rely on resourceVersion for this, right?

pkg/controller/servicemesh/controlplane/reconciler.go
luksa: True, but `getEarliestSMCPInNamespace()` should return `nil`, when no SMCPs are found. I mean, I could modify the calling code to correctly handle an empty SMCP, but somehow I find it semantically incorrect to return an empty SMCP when no SMCPs are found. But maybe I'm wrong in thinking this way.

pkg/controller/servicemesh/controlplane/reconciler.go
luksa: Which SMCP gets reconciled must be deterministic. If we were to choose which SMCP to reconcile at random, the controller could reconcile both SMCPs or neither. 

Think about what happens when we have two SMCPs with the same timestamp. 
1. The controller is asked to reconcile SMCP1. It chooses the earliest SMCP at random. Let's say it chooses SMCP2 this time, so it skips reconciliation of SMCP1 and adds the error to its status.
2. The controller is then asked to reconcile SMCP2. This time, it chooses SMCP1 as the earliest, so it skips reconciliation of SMCP2 and adds the error to it. 

Neither SMCP gets reconciled. In another scenario, the controller could reconcile both SMCPs. 

That's why we need to sort by name (or some other field), so that the same SMCP is always chosen as the one to reconcile.

luksa: `resourceVersion` is supposed to be opaque, meaning that we shouldn't rely on its value. All we're allowed to do is check if two resourceVersions are equal. 

It's possible that the format of `resourceVersion` will change in the future. It can already be different if you use a datastore other than etcd.

pkg/controller/servicemesh/controlplane/reconciler.go
jewertow: Ok, I forgot about returning nil... In such a case, it's absolutely reasonable to define this variable as a pointer.
But to be honest, I don't like the name `goPitfallWorkaround`. For me it's an expected behavior and I wouldn't call it Go's pitfall, so I would prefer the following alternatives:
1. Initialize the pointer and set its value:
```suggestion
			earliest = &v2.ServiceMeshControlPlane{}
			*earliest = smcp
```
2. Rename `goPitfallWorkaround` to `earliestTmp`:
```suggestion
			earliestTmp := smcp
			earliest = &earliestTmp
```

luksa: The reason to explicitly call this out as a pitfall is because it is a pitfall. It's probably the #1 pitfall in go. If a person is not aware of this, they might accidentally change the code, thinking they are simplifying it, when they are actually breaking it. 

In both of your examples, it's not clear why the code isn't simply `earliest = &smcp`. 

By explicitly calling out the pitfall, no-one will dare to change the code without fully understanding why it's written like it is. The variable name serves as a reminder.

jewertow: Ok, fine. You don't have to change it, but you won't convince me that this is a pitfall. This is just the nature of pointers. I think that the approach with a "tmp" variable is more clear, because it's a typical notation in algorithmic handbooks in C, so most people should quickly see what's going on.

> If a person is not aware of this, they might accidentally change the code

I would say that if a person is not aware of pointers, they should not touch Go. Actually changing the code without understanding how pointers work is safe, because you wrote tests.

jewertow (commented):
I tested this change manually on OCP and for invalid SMM and SMMR everything works fine, but I'm not sure if for SMCP as well. I applied a valid SMCP and then another one in the same namespace, and it was not reconciled as expected. But then I removed the invalid SMCP and both were removed, so I lost my valid deployments.

pkg/controller/servicemesh/controlplane/reconciler.go
luksa: But this isn't about knowing how pointers work. It's about knowing how loops are implemented in Go. And as you can see in the discussion, virtually everyone runs into this problem (which is why they're going to change how loops work):

> ### A decade of experience shows the cost of the current semantics
> I [talked at Gophercon once](https://go.dev/blog/toward-go2#explaining-problems) about how we need agreement about the existence of a problem before we move on to solutions. When we examined this issue in the run up to Go 1, it did not seem like enough of a problem. The general consensus was that it was annoying but not worth changing.
>
> Since then, **I suspect every Go programmer in the world has made this mistake in one program or another. I certainly have done it repeatedly over the past decade, despite being the one who argued for the current semantics and then implemented them. (Sorry!)**

I agree that tests would catch any breaking changes, but when a person runs into a failing test, they still need to figure out what's causing it to fail. This unnecessarily wastes time and money. It's always better to explicitly point out pitfalls like this.

bartoszmajsak: Leaving aside naming, which is always a never-ending debate, what I think could be good to leave in this piece of code is an explanation **why** we do what we do. A short comment with reference to this interesting thread about for-loop semantics will not only serve us as backup memory of why this decision was taken, but help any newcomer who stumbles upon this code base to spend some effort understanding the intricacies and give some extra care when touching it. Just my two cents.

luksa:
> But then I removed the invalid SMCP and both were removed, so I lost my valid deployments.

When you remove the second SMCP, the first one does not get removed, but all its child resources are indeed deleted by the pruner. That's because the label selector that the pruner uses to delete resources only considers the SMCP namespace; it ignores the SMCP name. As this is a different bug, I've opened a new [issue](https://issues.redhat.com/browse/OSSM-2274) for it. 

Thanks @jewertow!

jewertow:
Ok, thanks for the confirmation.

jewertow (approved):

PR #1056, opened 2022-11-07
OSSM-1094 Use bcrypt for htpasswd, stop mounting rawpassword

cam-garrison:
Changes:

-  Change to using bcrypt instead of SHA-1 for htpasswd. 

- Update the oauth-proxy version from 4.4 -> 4.9 (first version to support bcrypt)

- Stop mounting the raw htpasswd in our grafana-proxy and prometheus-proxy containers. 

Tested:
- OpenShift upgrading from operator 2.2 to 2.3 with this change and it worked as expected (keeping SHA-1 but nothing breaks) 
- Fresh install of this operator with SMCP 2.3 where it generates a bcrypt password, doesn't mount in prometheus-proxy and grafana-proxy as intended.

### DIFF
--- a/pkg/controller/servicemesh/controlplane/htpasswd.go
+++ b/pkg/controller/servicemesh/controlplane/htpasswd.go
@@ -3,9 +3,9 @@ package controlplane
 import (
 	"context"
 	"crypto/rand"
-	"crypto/sha1"
 	"encoding/base64"
 	"fmt"
+	"golang.org/x/crypto/bcrypt"
 	"regexp"
 
 	corev1 "k8s.io/api/core/v1"
@@ -32,9 +32,13 @@ func (r *controlPlaneInstanceReconciler) patchHtpasswdSecret(ctx context.Context
 			log.Error(err, "failed to generate the HTPasswd password")
 			return err
 		}
-		h := sha1.New()
-		h.Write([]byte(rawPassword))
-		auth = "internal:{SHA}" + base64.StdEncoding.EncodeToString(h.Sum(nil))
+		hashedPassword, err1 := bcrypt.GenerateFromPassword([]byte(rawPassword), bcrypt.DefaultCost)
+		if err1 != nil {
+			log.Error(err1, "failed to encrypt the raw password")
+			return err1
+		}
+		username := "internal"
+		auth = fmt.Sprintf("%s:%s", username, string(hashedPassword))
 	}
 
 	b64Password := base64.StdEncoding.EncodeToString([]byte(rawPassword))
--- a/pkg/controller/servicemesh/controlplane/htpasswd.go
+++ b/pkg/controller/servicemesh/controlplane/htpasswd.go
@@ -5,14 +5,13 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"fmt"
-	"golang.org/x/crypto/bcrypt"
 	"regexp"
 
+	"github.com/maistra/istio-operator/pkg/controller/common"
+	"golang.org/x/crypto/bcrypt"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	"github.com/maistra/istio-operator/pkg/controller/common"
 )
 
 func (r *controlPlaneInstanceReconciler) patchHtpasswdSecret(ctx context.Context, object *unstructured.Unstructured) error {
--- a/pkg/controller/servicemesh/controlplane/htpasswd.go
+++ b/pkg/controller/servicemesh/controlplane/htpasswd.go
@@ -7,11 +7,12 @@ import (
 	"fmt"
 	"regexp"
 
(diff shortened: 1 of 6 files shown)


pkg/controller/servicemesh/controlplane/htpasswd.go
```diff
@@ -32,9 +32,13 @@ func (r *controlPlaneInstanceReconciler) patchHtpasswdSecret(ctx context.Context
...
 		}
-		h := sha1.New()
-		h.Write([]byte(rawPassword))
-		auth = "internal:{SHA}" + base64.StdEncoding.EncodeToString(h.Sum(nil))
+		hashedPassword, err1 := bcrypt.GenerateFromPassword([]byte(rawPassword), bcrypt.DefaultCost)
+		if err1 != nil {
+			log.Error(err1, "failed to encrypt the raw password")
+			return err1
+		}
+		username := "internal"
```
luksa, line 40: Great job introducing this variable. Helps explain what "internal" is. In the old code, it wasn't clear at all.

pkg/controller/servicemesh/controlplane/htpasswd.go
```diff
@@ -32,9 +32,13 @@ func (r *controlPlaneInstanceReconciler) patchHtpasswdSecret(ctx context.Context
...
+		auth = fmt.Sprintf("%s:%s", username, string(hashedPassword))
```
luksa, line 41: Don't you need to add a special prefix to the password, indicating that it's using bcrypt, or does bcrypt itself add the correct prefix?

jewertow: No, `bcrypt` does it.

jewertow: Screenshot from 2022-11-08 13-03-18

jewertow: This is scheme of the output:
```
$2a$12$R9h/cIPz0gi.URNNX3kh2OPST9/PgBkqquzi.Ss7KIUgO2t0jWMUW
\__/\/ \____________________/\_____________________________/
Alg Cost      Salt                        Hash
```

jewertow (approved):

jewertow:
@cam-garrison did you test it manually on OpenShift?

cam-garrison:
> @cam-garrison did you test it manually on OpenShift?

Yes, tested manually exclusively on OpenShift.

luksa (approved):

resources/smcp-templates/v2.3/maistra
```diff
@@ -29,7 +29,7 @@ spec:
         container:
           imageRegistry: quay.io/openshift
           imageName: origin-oauth-proxy
-          imageTag: "4.4"
+          imageTag: "4.9"
```
jwendell, line 32: I'm afraid this doesn't do what you want.

I think you will want to change this
https://github.com/maistra/istio-operator/blob/maistra-2.3/deploy/src/deployment-servicemesh.yaml#L53

However I'm not sure if we should bump this in a z release or if it's better to do this in 2.4.

jwendell:
btw, the linter failures seems legit.

resources/smcp-templates/v2.3/maistra
cam-garrison: For prometheus and grafana, I found in their respective deployments ([prometheus](https://github.com/maistra/istio-operator/blob/fbc41861f7569f05fec20cd21247983287fd21c7/resources/helm/overlays/istio-telemetry/prometheus/templates/deployment.yaml#L40), [grafana](https://github.com/maistra/istio-operator/blob/81c86c46f8b195f20be7a0a1ada207f1a48bf587/resources/helm/overlays/istio-telemetry/grafana/templates/deployment.yaml#L48)) that the version of oauth-proxy that they reference in their deployments depend on this `global.oauthproxy` variable which I found defined in that smcp-templates file, so I changed that one. 

This change does affect the versions of oauth-proxy used in prometheus-proxy and grafana-proxy, and had the desired effect in my testing, as before I made that change those proxies were unable to work with bcrypt.

Would updating the imageTag in the file you referenced maybe update a different oauth proxy in our deployment of the operator? I'm unsure.

jwendell: I'm concerned with your comment: `Update the oauth-proxy version from 4.4 -> 4.9 (first version to support bcrypt)`

The link I posted is very likely the source of truth. The file you updated in this PR is mostly unused (they are only used if the customer makes use of templates, which by default they don't).

So, if bcrypt is only supported on oauth-proxy 4.9 and by default we use 4.4, then this PR will break things.

jwendell: I'm putting a hold on this PR until someone more experienced in operator can confirm this.
cc @rcernich @luksa

luksa: We need to re-target this PR for 2.4.

cam-garrison:
closing, will open a new PR against maistra-2.4 now and then make necessary changes (run `make gen` to apply changes to `resources/v2.4/`) when `v2.4` exists in the resources folder (likely near the end of the rebase cycle?)

PR #1059, opened 2022-11-09
OSSM-1094 Use bcrypt for htpasswd, stop mounting rawpassword

cam-garrison:
See previous PR #1056, re-targeting for 2.4 release.

Will make necessary changes (run make gen to apply changes to resources/v2.4/) when v2.4 exists in the resources folder (likely near the end of the rebase cycle?)

from previous PR:

### Changes:

- Change to using bcrypt instead of SHA-1 for htpasswd.

- Update the oauth-proxy version from 4.4 -> 4.9 (first version to support bcrypt)

- Stop mounting the raw htpasswd in our grafana-proxy and prometheus-proxy containers.

### Tested:

- OpenShift upgrading from operator 2.2 to 2.3 with this change and it worked as expected (keeping SHA-1 but nothing breaks)
- Fresh install of this operator with SMCP 2.3 where it generates a bcrypt password, doesn't mount in prometheus-proxy and grafana-proxy as intended.


dgn (commented):
lgtm so far, but haven't actually tested it yet. The only thing that gives me a slight headache is the updaete of oauth-proxy to the 4.9 version- I wonder if that increases our minimum supported OCP version to 4.9? but even so, in 2.4 that should be fine

cam-garrison:
closing and reopening for 2.4. rebasing was a poor idea here ✖_✖

PR #1061, opened 2022-11-17
OSSM-2126 Support configurable Grafana storage capacity

bmangoen:


resources/helm/v2.3/istio-telemetry/grafana/templates/pvc.yaml
```diff
@@ -14,5 +14,5 @@ spec:
     - {{ .Values.grafana.accessMode }}
   resources:
     requests:
-      storage: 5Gi
+      storage: {{ .Values.grafana.persistenceResources.requests.storage | quote }}
```
jewertow, line 17: You should apply this change to a corresponding file in the directory `helm/overlays` and then run `make gen`.

pkg/apis/maistra/v2/grafana.go
```diff
@@ -27,7 +27,7 @@ type GrafanaInstallConfig struct {
 	Service *ComponentServiceConfig `json:"service,omitempty"`
 	// Persistence configures a PersistentVolume associated with the grafana installation
 	// .Values.grafana.persist, true if not null
-	// XXX: capacity is not supported in the charts, hard coded to 5Gi
+	// XXX: Capacity is configurable, default: 5Gi
```
luksa: You can remove this line. These XXX lines were added as TODOs by Rob when he initially implemented the v2 SMCP.

resources/helm/v2.3/istio-telemetry/grafana/templates/pvc.yaml
bmangoen: As discussed with @luksa, I do the charts changes using the script [patch-charts](https://github.com/maistra/istio-operator/blob/maistra-2.3/build/patch-charts.sh)

luksa (approved):
LGTM. Tested manually with following SMCP:
```
spec:
  addons:
    grafana:
      enabled: true
      install:
        persistence:
          capacity:
            requests:
              storage: 10Gi
          enabled: true
```

PR #1063, opened 2022-11-22
OSSM-2188 Reconcile SMMR even when SMCP is missing

luksa:
This ensures that any ServiceMeshMembers that were created by the controller while the SMCP existed get removed.

### DIFF
--- a/pkg/controller/servicemesh/memberroll/controller.go
+++ b/pkg/controller/servicemesh/memberroll/controller.go
@@ -281,55 +281,45 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
 		return reconcile.Result{}, err
 	}
 
+	requiredNamespaces, err := r.getRequiredNamespaces(ctx, roll, meshNamespace)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
 	memberStatusMap := map[string]maistrav1.ServiceMeshMemberStatusSummary{}
 	for _, c := range roll.Status.MemberStatuses {
 		memberStatusMap[c.Namespace] = c
 	}
 
 	// 3. create ServiceMeshMember object for each ns in spec.members
-	memberNamespaces := roll.Spec.Members
-	if roll.Spec.IsClusterScoped() {
-		nsList := &corev1.NamespaceList{}
-		err = r.Client.List(ctx, nsList)
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-		memberNamespaces = []string{}
-		for _, ns := range nsList.Items {
-			if ns.Name != meshNamespace && !isExcludedNamespace(ns.Name) {
-				memberNamespaces = append(memberNamespaces, ns.Name)
-			}
-		}
-	}
-	for _, ns := range memberNamespaces {
+	for _, ns := range requiredNamespaces.List() {
 		member, err := r.ensureMemberExists(ctx, ns, mesh.Name, meshNamespace)
 		if err != nil {
 			return reconcile.Result{}, err
-		}
-		if member == nil {
+		} else if member == nil {
 			setMemberCondition(memberStatusMap, ns, maistrav1.ServiceMeshMemberCondition{
 				Type:    maistrav1.ConditionTypeMemberReconciled,
 				Status:  corev1.ConditionFalse,
 				Reason:  maistrav1.ConditionReasonMemberNamespaceNotExists,
 				Message: fmt.Sprintf("Namespace %s does not exist", ns),
 			})
-		} else {
-			ref := member.Spec.ControlPlaneRef
-			if ref.Name != mesh.Name || ref.Namespace != meshNamespace {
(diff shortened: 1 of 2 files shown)


pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -286,34 +283,22 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
...
-		memberNamespaces = []string{}
-		for _, ns := range nsList.Items {
-			if ns.Name != meshNamespace && !isExcludedNamespace(ns.Name) {
-				memberNamespaces = append(memberNamespaces, ns.Name)
+	// 2. create ServiceMeshMember object for each ns in spec.members
+	if mesh != nil {
+		for _, ns := range requiredNamespaces.List() {
+			member, err := r.ensureMemberExists(ctx, ns, mesh.Name, meshNamespace)
+			if err != nil {
+				return reconcile.Result{}, err
```
jewertow, line 291:
```suggestion
                                return common.RequeueWithError(err)
```

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -286,34 +283,22 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
...
+			} else if member == nil {
```
jewertow, line 292: The previous branch of this if-else statement returns from function when it's true, so `else if` is not necessary in this case. What do you think about this?
```suggestion
			}
			if member == nil {
```

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -411,6 +416,25 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
 	return reconcile.Result{}, nil
```
jewertow, line 416: Since you are changing this function, what do you think about refactoring this line?
```suggestion
	return common.Reconciled()
```

jewertow (approved):
I tested this change and I can confirm that the operator removed role bindings, network attachment definition, network policy and label from a member namespace.

pkg/controller/servicemesh/memberroll/controller.go
luksa: I don't see any real benefit of doing this. I see how you could argue that seeing `RequeueWithError()` immediately tells you that the object is getting re-queued, but on the other hand, when you see this function call, you can't know what it does until you look at its implementation. 

I feel invoking a function here just slows down anyone reading this code (until they know by heart what the function does), whereas a simple `return reconcile.Result{}, err` perfectly shows what it does (even though it doesn't say that the object gets re-queued, but that's something every engineer working on k8s controllers should know). Also, who guarantees that the fact that you're returning an error will result in the object being re-queued? That may not be true in some cases. In this case, the function name would be misleading.

pkg/controller/servicemesh/memberroll/controller.go
luksa: You're correct, but the reason I did it like this is to clearly separate the exceptional and the normal outcome of the `ensureMemberExists` function. 

The exceptional outcomes of the function are an actual error or a nil member, both of which require special handling. By using `if ... else if` the handling of these outcomes is visually tied together, whereas a separate if statement wouldn't be (as much). 

In reality, the `ensureMemberExists` function should return a NamespaceNotExist type of error instead of returning a nil member.

pkg/controller/servicemesh/memberroll/controller.go
jewertow: > I don't see any real benefit of doing this.

Ok, I expected that you will like the idea, because I remember your positive feedback when I added this function. Look at this [comment](https://github.com/maistra/istio-operator/pull/910#discussion_r866839422).

> whereas a simple `return reconcile.Result{}, err` perfectly shows what it does (even though it doesn't say that the object gets re-queued, but that's something every engineer working on k8s controllers should know)

It's a subjective point of view and I don't agree with you. I review PRs regularly and every time I see reconciliation I have to stay focus to understand how we reconcile objects. And it's not only my opinion - look at point 7 [here](https://cloud.redhat.com/blog/7-best-practices-for-writing-kubernetes-operators-an-sre-perspective).

luksa: I don't dislike the idea. I just think there are pros and cons (as always) that I outlined in my previous comment. And I agree with point 7, but also think it doesn't address the part about having to go look at the function (and possibly being misleading if the error is handled in one of the calling functions and thus preventing the re-queue).

The thing with these types of helper functions is that nothing forces you to use it, so you'll eventually end up with X% of the code using the function and the remaining code that doesn't. Then the question of consistency comes up. 

I don't really have a strong opinion on using one approach or the other. I guess I have to give your opinion a bit more weight, since I've been writing controllers a long time and it's all the same to me. But if it helps people who work on controller code less often, I agree that we should be using the function.

However, if we decide that we should always use `RequeueWithError` and similar, we should probably add a linter that enforces this decision.

PR #1067, opened 2022-11-25
OSSM-2274 Only prune resources that reference the SMCP for which pruning is being performed

luksa:
This ensures that resources aren't deleted when you start with two SMCPs in the same namespace and then delete the one with the ErrMultipleSMCPs status.


dgn:
Hmm, those lint failures look legit

luksa:
Yeah, the lint failures are legit, but this PR doesn't modify those lines. The fix is in a separate PR: https://github.com/maistra/istio-operator/pull/1074

jwendell (approved):

PR #1071, opened 2022-12-07
[OSSM-1258] bump prometheus/client_golang version

cam-garrison:
Bump module prometheus/client_golang to `v1.11.1`. Tidy and update vendoring with new package versions. 

Tested by:

- built operator and deployed an SMCP on CRC, deployed bookinfo, ensured metrics were visible on grafana.
- locally ran `quay.io/maistra-dev/maistra-builder:2.3` `make test`, those tests passed on my end.

Show me some green tests prow :cold_sweat:


rcernich (approved):

dgn (approved):
stupid question, but.. is this even used? it's an indirect dependency and we're not providing prometheus endpoints or credentials anywhere I don't think, so - likely this was never used?

cam-garrison:
> stupid question, but.. is this even used? it's an indirect dependency and we're not providing prometheus endpoints or credentials anywhere I don't think, so - likely this was never used?

I think this is never used/imported but was still flagged as a CVE as an indirect dependency. @dgn do you think it's worth merging to close out the CVE as done? or we shouldn't bother as it's likely not an attack surface in the operator.

cam-garrison:
you were right, checked with Rob as well, closing as InstrumentHandlerCounter does not appear to be used in our code base or called in methods we use from dependencies.

PR #1073, opened 2022-12-12
[ossm-2221] ensure we do not auto-inject into kiali

cam-garrison:
Now that we enable injection into the SMCP namespace, we must ensure that if a user has auto inject enabled, our SMCP related resources will not be injected into unintentionally. 

This PR ensures that if the user is installing kiali through our SMCP config, by default the kiali pod will be labeled with `sidecar.istio.io/inject="false"`.


cam-garrison:
closing and will make this change for 2.4 in a future pr

PR #1077, opened 2022-12-19
OSSM-2152: Initial support for 2.4

jwendell:


pkg/controller/common/test/client.go
```diff
@@ -134,7 +134,7 @@ func (c *fakeClient) List(ctx context.Context, list runtime.Object, opts ...clie
 		return err
 	}
 	if o == nil {
-		return errors.NewInternalError(fmt.Errorf("no resource returned by Fake"))
+		return errors.NewInternalError(fmt.Errorf("111 no resource returned by Fake"))
```
dgn: is this some leftover debug output?

pkg/controller/common/test/client.go
```diff
@@ -329,7 +329,7 @@ func (c *fakeClient) Status() client.StatusWriter {
 
 func (c *fakeClient) copyObject(source, target runtime.Object) error {
 	if source == nil {
-		return errors.NewInternalError(fmt.Errorf("no resource returned by fake"))
+		return errors.NewInternalError(fmt.Errorf("222 no resource returned by fake"))
```
dgn: same here

yxun:
What's the strategy_v2_4 used for ? Should we deprecate maistra v1 apis in 2.4 operator ?

jwendell:
IDK. This is just an initial port of 2.4 so that we have initial builds for it in brew. Further improvements (like dropping support for 1.1, 2.0, etc), updating SDK, etc will come after this PR is merged.

build/codegen/update-generated.sh
```diff
@@ -3,33 +3,40 @@
 set -o errexit
 set -o nounset
 set -o pipefail
+set -x
 
-bash vendor/k8s.io/code-generator/generate-groups.sh \
+# FIXME: https://issues.redhat.com/browse/OSSM-2367
+CODE_GENERATOR_PACKAGE="k8s.io/code-generator"
+CODE_GENERATOR_VERSION="v0.18.3"
```
dgn:
```suggestion
CODE_GENERATOR_VERSION="$(go list -m -f '{{ if .Replace }} {{ .Replace.Version }} {{ else }} {{ .Version }} {{ end }}' ${CODE_GENERATOR_PACKAGE})"
```

This will pull the module version from `go.mod`. If you use this you can probably remove the FIXME above :)

dgn (approved):
looking good, but haven't actually tried it yet

build/codegen/update-generated.sh
jwendell: Thanks a lot for this!!!!!

PR #1082, opened 2022-12-22
[OSSM-2221 | OSSM-2340] Ensure injection in SMCP ns disabled below v2.4

cam-garrison:
**This PR:** 

- Ensures that the ignore-namespace label is still applied to the SMCP namespace when creating SMCPs below version 2.4. 
- Modifies the v2.4 MutatingWebhook ObjectSelector to never attempt to inject into the istiod pod [OSSM-2340]
- Modifies the v2.4 kiali_cr to label the kiali pod to not inject, fixing the problem when an SMCP was created with auto-inject. 
- Adds another namespaceLabels unit test to test labels added/removed pre v2.4 and post v2.4. 

**Tested on CRC by:**
- Creating a v2.4 SMCP, deleting istiod pod and seeing it self heal.
- Creating a v2.4 SMCP with auto-inject enabled and seeing kiali created successfully with no proxy container.
- Creating v2.3, v2.2 SMCPs and making sure the ignore-namespace label is applied, and deleted when smcp is deleted.

### DIFF
--- a/resources/helm/overlays/istio-control/istio-discovery/templates/mutatingwebhook.yaml
+++ b/resources/helm/overlays/istio-control/istio-discovery/templates/mutatingwebhook.yaml
@@ -48,18 +48,22 @@ webhooks:
         - disabled
       - key: istio-env
         operator: DoesNotExist
-{{- if .Values.sidecarInjectorWebhook.objectSelector.enabled }}
   objectSelector:
     matchExpressions:
+      - key: app
+        operator: NotIn
+        values:
+          - "istiod"
+{{- if .Values.sidecarInjectorWebhook.objectSelector.enabled }}
 {{- if eq .Values.global.proxy.autoInject "enabled" }}
-    - key: "sidecar.istio.io/inject"
-      operator: NotIn
-      values:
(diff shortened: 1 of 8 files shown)


dgn (approved):

resources/helm/overlays/istio-control/istio-discovery/templates/mutatingwebhook.yaml
```diff
@@ -48,18 +48,22 @@ webhooks:
...
+        operator: NotIn
+        values:
+          - "istiod"
+{{- if .Values.sidecarInjectorWebhook.objectSelector.enabled }}
 {{- if eq .Values.global.proxy.autoInject "enabled" }}
-    - key: "sidecar.istio.io/inject"
-      operator: NotIn
-      values:
-      - "false"
+- key: "sidecar.istio.io/inject"
```
dgn: these whitespaces were required though, right?

cam-garrison: You are correct, fixing this now.

PR #1083, opened 2023-01-12
bump version to v2.1.6

rcernich:
OSSM-2431 update goutils to v1.1.1


jwendell (approved):
Do you need to change https://github.com/maistra/istio-operator/blob/maistra-2.1/pkg/controller/servicemesh/controlplane/profiles_test.go#L23 ?

rcernich:
@jwendell, tests pass locally

rcernich:
I'll update the test, then figure out why it's passing locally.

PR #1090, opened 2023-01-17
OSSM-2492 Pass tolerations to Jaeger

luksa:


jwendell (approved):

jwendell:
@luksa I don't think a cherry-pick will work for 2.4, correct? We should instead wait for the automated job that picks 2.3 changes into 2.4?

luksa:
@jwendell The job that picks the 2.3 charts into 2.4 won't update the overlays dir, so we definitely need a cherry-pick. And as you say, the automated cherry-pick will probably fail, because it won't update the 2.4 charts, causing gen-check to fail.

PR #1095, opened 2023-01-25
DNM: ossm-744 Update operator SDK

yxun:


yxun:
This is a draft PR about upgrading operator sdk dependencies from our existing sdk 0.18.0 to current latest version 1.25.0
There are several major code changes in the PR. I will collect some reference and write a note in drive.

Need some help review the part of our Reconcile methods parameter change (replacing or adding a golang context param). I think that's one of the major breaking change in this draft now. 

Could I get some help about the correct use of golang context.Context ? Thanks,

Makefile
```diff
@@ -46,6 +46,7 @@ ifeq "${GIT_UPSTREAM_REMOTE}" ""
 GIT_UPSTREAM_REMOTE = "ci-upstream"
 $(warning Could not find git remote for maistra/istio-operator, adding as '${GIT_UPSTREAM_REMOTE}')
 $(shell git remote add ${GIT_UPSTREAM_REMOTE} https://github.com/maistra/istio-operator.git)
+$(shell git config --global --add safe.directory /go/src/github.com/maistra/istio-operator)
```
jwendell: This won't work in a local build. This is meant to workaround an issue in openshift prow.

jwendell: Perhaps using `'*'` instead of hardcoding the path.

PR #1101, opened 2023-02-03
OSSM-1330: Fix integration with cert-manger-istio-csr

jewertow:

This change makes `pilotSecretName` and `rootCAConfigMapName` optional, because:
1. Istio Operator manages istiod webhooks and therefore must know `pilotSecretName` before SMCP is deployed. With this change, the operator will ignore this field and we will have to deprecate it in 2.4 and remove in 2.5.
2. istio-csr does not allow to customize `rootCAConfigMapName` since v0.3.0 and it's always provided with a fixed name, so it does not make sense to require this values since it's constant. Custom value is respected, but by default `istio-ca-root-cert` is used.

### DIFF
--- a/deploy/examples/cert-manager/README.md
+++ b/deploy/examples/cert-manager/README.md
@@ -3,7 +3,7 @@
 #### Requirements
 
 OpenShift Service Mesh requires cert-manager to provide:
-* a secret with an intermediate CA key and certificate for istiod:
+* a TLS secret with an intermediate CA key and certificate for istiod:
     * name: `istiod-tls`;
     * namespace: the same one in which SMCP is deployed;
     * data keys: `tls.key` and `tls.crt`.
--- a/deploy/examples/cert-manager/README.md
+++ b/deploy/examples/cert-manager/README.md
@@ -1,13 +1,29 @@
-# Integration with cert-manager-istio-csr
+## Integration with cert-manager and istio-csr
+
+#### Requirements
+
+OpenShift Service Mesh requires cert-manager to provide:
+* a secret with an intermediate CA key and certificate for istiod:
+    * name: `istiod-tls`;
+    * namespace: the same one in which SMCP is deployed;
+    * data keys: `tls.key` and `tls.crt`.
+* a config map with a root certificate:
+    * name: `istio-ca-root-cert`;
+    * namespace: the same one in which SMCP is deployed;
+    * data key: `root-cert.pem`.
+
+Currently, istio-csr does not allow customizing namespace in which secrets and config maps are created,
+but if it will change in the future, OSSM users will have to configure istio-csr to provide objects named as described above.
+
+#### Steps
 
 1. Install cert-manager
 ```shell
-kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.11.0/cert-manager.crds.yaml
-helm install \
-    cert-manager jetstack/cert-manager \
+helm install cert-manager jetstack/cert-manager \
     --namespace cert-manager \
     --create-namespace \
-    --version v1.11.0
+    --version v1.11.0 \
+    --set installCRDs=true
 ```
 
 2. Provision certificates:
--- a/deploy/examples/cert-manager/README.md
+++ b/deploy/examples/cert-manager/README.md
@@ -24,11 +24,12 @@ helm install -n cert-manager cert-manager-istio-csr \
 
 4. Deploy Istio:
 ```shell
-oc apply -f deploy/examples/cert-manager/smcp.yaml
+oc apply -f deploy/examples/cert-manager/smcp.yaml -n istio-system
 ```
 
 5. Deploy bookinfo app:
 ```shell
+oc new-project bookinfo
 oc apply -f https://raw.githubusercontent.com/maistra/istio/maistra-2.3/samples/bookinfo/platform/kube/bookinfo.yaml -n bookinfo
 ```
 
--- a/deploy/examples/cert-manager/README.md
+++ b/deploy/examples/cert-manager/README.md
@@ -2,12 +2,12 @@
 
 1. Install cert-manager
 ```shell
-kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.8.0/cert-manager.crds.yaml
+kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.11.0/cert-manager.crds.yaml
 helm install \
     cert-manager jetstack/cert-manager \
     --namespace cert-manager \
     --create-namespace \
-    --version v1.8.0
+    --version v1.11.0
 ```
 
 2. Provision certificates:
@@ -31,6 +31,27 @@ oc apply -f deploy/examples/cert-manager/smcp.yaml -n istio-system
 ```shell
 oc new-project bookinfo
 oc apply -f https://raw.githubusercontent.com/maistra/istio/maistra-2.3/samples/bookinfo/platform/kube/bookinfo.yaml -n bookinfo
+```
+
+### Integration with cert-manager in OSSM 2.3.0 and 2.3.1
+
+To make SMCP 2.3.0 and 2.3.1 work with cert-manager it's required to manually adjust certificate provided by cert-manager.
+
+1. Follow steps 1-3 from the previous section.
+
+2. Get intermediate certificate for Istio provided by cert-manager:
+```shell
+oc get secret istiod-tls -n istio-system -o json | jq -r '.data."tls.crt"' | base64 -d > ca-cert.pem
+```
+
+3. Create secret for Istio Operator from retrieved certificate:
+```shell
+oc create secret generic istio-ca-secret -n istio-system --from-file=ca-cert.pem
+```
+
+4. Deploy Istio:
+```shell
+oc apply -f deploy/examples/cert-manager/smcp-2.3.1.yaml -n istio-system
 ```
 
 ### Verification
--- a/deploy/examples/cert-manager/README.md
+++ b/deploy/examples/cert-manager/README.md
@@ -8,7 +8,7 @@ OpenShift Service Mesh requires cert-manager to provide:
     * namespace: the same one in which SMCP is deployed;
     * data keys: `tls.key` and `tls.crt`.
 * a config map with a root certificate:
-    * name: `istio-ca-root-cert`;
+    * name: `istio-ca-root-cert` or custom specified in `spec.security.certificateAuthority.cert-manager.rootCAConfigMapName`;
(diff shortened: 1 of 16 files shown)


pkg/apis/maistra/conversion/security.go
```diff
@@ -158,16 +158,7 @@ func populateSecurityValues(in *v2.ControlPlaneSpec, values map[string]interface
 			extraArgs := []string{
 				"--tlsCertFile=/etc/cert-manager/tls/tls.crt",
 				"--tlsKeyFile=/etc/cert-manager/tls/tls.key",
-				"--caCertFile=/etc/cert-manager/tls/ca.crt",
-			}
-
-			suppliedRootConfigMap := (CertManagerConf.RootCAConfigMapName != "")
-
-			if suppliedRootConfigMap {
-				extraArgs[2] = "--caCertFile=/etc/cert-manager/ca/root-cert.pem"
```
jewertow, line 167: To be honest, I don't understand why it was suggested to override `--caCertFile`, because secret `istiod-tls/ca.crt` and config map `istio-ca-root-cert/root-cert.pem` contains the same certificate.
@rcernich @SpectralHiss any idea?

pkg/controller/servicemesh/webhookca/controller.go
```diff
@@ -45,20 +46,28 @@ const (
...
+			sidecarInjectorSecretName: common.IstioRootCertKey,
+		},
 	},
 	istiodWebhookNamePrefix: &SecretCABundleSource{
-		SecretNames: []string{istiodCustomCertSecretName, istiodSecretName},
-		Key:         common.IstiodCertKey,
+		SecretNameAndKeyPairs: map[string]string{
+			istiodCustomCertSecretName:  common.IstiodCertKey,
+			istiodSecretName:            common.IstiodCertKey,
+			istiodCertManagerSecretName: common.IstiodCertManagerCertKey,
```
jewertow: This line, and the corresponding line below, are crucial changes in this PR. The operator was not able to reconcile webhooks, because it couldn't find proper secret.

pkg/apis/maistra/conversion/security.go
SpectralHiss: It is safer to use the supplied configmap , in cert-manager it is not guaranteed that in the future that ca.crt key will exist , the root is managed by istio-csr in this installation.

pkg/controller/servicemesh/webhookca/controller.go
SpectralHiss: Yes these are crucial and it would work with this code change, but we are working on a fix that doesn't require you to change this. we will show you it in a few days tops once it's out. if there are any downsides to the alternative solution we can come back to this one. There is risk with changing this key for other providers i think.

SpectralHiss: Actually after consulting with other istio-csr maintainers, we think that perhaps this change is the better approach, in the meanwhile it wouldn't be unreasonable to add some docs with the steps for version v2.3.0 to just run a pre-installation step to explicitely create the secret in the istio-system namespace manually, it is best practice to get this secret and statically set it in cert-manager namespace anyways, so manually doing it would also be fine, we can even keep doing that for version v2.3.1 if we wanted but i am happy to accept these changes as well.

The only risk we need to make sure is that by changing this we break people who are manually loading in certs etc..

jewertow: > in the meanwhile it wouldn't be unreasonable to add some docs with the steps for version v2.3.0 to just run a pre-installation step to explicitely create the secret in the istio-system namespace manually

Good idea. I completely forgot, that this secret could be customized by users manually to make it work with 2.3.0.

pkg/apis/maistra/conversion/security.go
SpectralHiss: but if you think it makes things much simpler i wouldn't disagree too hard. more of a future-proofing thing we do because we want people to rely less on ca.crt

pkg/controller/servicemesh/webhookca/controller.go
jewertow: Taking into account that users can customize secret to make it work with istio-operator 2.3.0, I have to revert removal of `pilotSecretName` and `rootCAConfigMapName`. However, I'm still thinking why we have these values configurable.
1. `rootCAConfigMapName` is no longer configurable in istio-csr since the release v.0.3.0, so shouldn't we deprecate it?
2. If it comes to `pilotSecretName`, I can imagine that it will become configurable in the future, but what if secret keys will change and we don't enable users to configure them?

Instead of making these values configurable in istio-operator, they should be configurable in istio-csr, so that users can adjust them for Istio, instead of adjusting Istio for istio-csr.

SpectralHiss: You make a good point, i think i was testing with an old version when i first set it up, i don't think even that `pilotSecretName` will change in the future so we could simply get rid of them both!

SpectralHiss: You make a good point, i think i was testing with an old version when i first set it up, i don't think even that `pilotSecretName` will change in the future so we could simply get rid of the configuration parameters for both! (but still use the ca root)

On the previous point for documentation, what's the best place to write up all this information and a quick installation guide in the red hat openshift website or maistra docs?

pkg/apis/maistra/conversion/security.go
jewertow: No, I don't think so. Thank you for explaining this. I will change this to use the config map.

pkg/controller/servicemesh/webhookca/controller.go
jewertow: We will document this in the official OpenShift docs.

SpectralHiss: Can i help in any way with that or are you good?

jewertow: It would be great if you could review our PRs related to this integration. I will request your reviews, ok?

jewertow:
Ok, this is ready for final review. @SpectralHiss could you take a look once again?

deploy/examples/cert-manager/README.md
```diff
@@ -1,35 +1,92 @@
...
-        cert-manager jetstack/cert-manager \
-        --namespace cert-manager \
-        --create-namespace \
-        --version v1.6.1 \
-        --set installCRDs=true
+#### Requirements
 
-## Prepare istio-system Namespace
+OpenShift Service Mesh requires cert-manager to provide:
+* a secret with an intermediate CA key and certificate for istiod:
```
luksa: You can call this a "TLS secret", as that's exactly how a secret created via `kubectl create secret tls` looks like.

deploy/examples/cert-manager/README.md
```diff
@@ -1,35 +1,92 @@
...
-## Create Istio Control Plane
+2. Get intermediate certificate for Istio provided by cert-manager:
+```shell
+oc get secret istiod-tls -n istio-system -o json | jq -r '.data."tls.crt"' | base64 -d > ca-cert.pem
+```
 
-    kubectl apply -f deploy/examples/cert-manager/smcp.yaml
+3. Create secret for Istio Operator from retrieved certificate:
+```shell
+oc create secret generic istio-ca-secret -n istio-system --from-file=ca-cert.pem
```
luksa, line 65: Why does the certificate have to be stored in a Secret rather than a ConfigMap? Since this is a public certifcate, it doesn't have to be kept secure.

deploy/examples/cert-manager/README.md
```diff
@@ -1,35 +1,92 @@
...
+    openssl s_client -showcerts istiod-test-prototype.istio-system:15012 < /dev/null
+```
+2. Check app certificate:
+```shell
+kubectl exec $(kubectl get pods -l app=productpage -o jsonpath='{.items[0].metadata.name}') -c istio-proxy -- \
+    openssl s_client -showcerts details.bookinfo:9080 < /dev/null
+```
+3. Check certificates with `istioctl`:
+```shell
+istioctl pc s $(kubectl get pods -l app=productpage -o jsonpath='{.items[0].metadata.name}') -o json |\
```
luksa: Since we don't support istioctl, do we need to use an alternative command in this example?

deploy/examples/cert-manager/smcp-2.3.1.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: maistra.io/v2
```
luksa, line 1: Why do we need a separate SMCP manifest for 2.3.1?

luksa: If we really do need it, would renaming it to `smcp-2.3.x.yaml` be more appropriate?

luksa (approved):
LGTM. I only added a few questions.

deploy/examples/cert-manager/README.md
jewertow: This is required, because Istio Operator expects to find secrets `cacerts` and `istio-ca-secret`.

deploy/examples/cert-manager/README.md
jewertow: No, we don't need. I will remove this command.

deploy/examples/cert-manager/smcp-2.3.1.yaml
jewertow: > Why do we need a separate SMCP manifest for 2.3.1?

Because 2.3.0/1 need fields `pilotSecretName` and `rootCAConfigMapName`. In 2.3.2 the first will be ignored, and the second will be optional.

> would renaming it to smcp-2.3.x.yaml be more appropriate?

I don't think so, because in 2.3.2+ `pilotSecretName` and `rootCAConfigMapName` will be not recommended to use.

deploy/examples/cert-manager/README.md
jewertow: Done

deploy/examples/cert-manager/README.md
jewertow: You're right. Good point.

pbajjuri20:
Deployed the Cert-Manager on the OCP cluster with 2.3.1 SM Operator, it is working as expected.

SpectralHiss:
> Ok, this is ready for final review. @SpectralHiss could you take a look once again?

Sorry i missed this, all looks fine for me, only thing is i want to make sure that we surface some reference to those steps and a quick explanation between 2.3.0 workaround and 2.3.1 in the official docs?

jewertow:
Thanks @SpectralHiss!

If it comes to documenting workaround for 2.3.0 and 2.3.1, I'm discussing this with the docs team.

PR #1105, opened 2023-02-07
OSSM-3611 Fix readOnly argument in cert-manager and ca-root-cert volumeMount

aidy:


jewertow:
@aidy thank you for the PR and I'm sorry for late reply. Could you rebase this branch?

aidy:
@jewertow Done

luksa:
@aidy Thank you for this PR. What issue does it solve? (I'd like to file it in our Jira)

aidy:
@luksa It corrects the read-only argument that was intended to be specified for the volume mount. As it stands, it's an invalid specification and is silently ignored.

dgn (approved):

jewertow (approved):

PR #1106, opened 2023-02-08
OSSM-3341: Prometheus jobs for dual-stack

jewertow:
These relabel configurations work in IPv4 and IPv6 and should work in dual stack. I tested this manually on IPv4 and IPv6 clusters and I verified these rules with Prometheus [unit tests](https://github.com/prometheus/prometheus/compare/main...jewertow:prometheus:relabel-dual-stack-ip?expand=1).


resources/helm/overlays/istio-telemetry/prometheus/templates/configmap.yaml
```diff
@@ -34,11 +34,15 @@ data:
       - source_labels: [__meta_kubernetes_pod_container_port_name]
         action: keep
         regex: '.*-envoy-prom'
-      - source_labels: [__address__, __meta_kubernetes_pod_annotation_prometheus_io_port]
+      - source_labels: [__meta_kubernetes_pod_ip]
```
luksa, line 37: How come you are changing this?

jewertow: It's much easier to write a regex for a pod IP than for an address, because the address contains a port. In this case we don't need `__meta_kubernetes_pod_annotation_prometheus_io_port`, because 15090 is used when `enablePrometheusMerge` is disabled.

luksa (approved):

PR #1109, opened 2023-02-10
OSSM-2374 Delete SMMR only when no SMM resources point to it

luksa:


pkg/controller/servicemesh/member/controller_test.go
```diff
@@ -173,36 +175,129 @@ func TestReconcileRemovesFinalizerFromMemberWhenMemberDeleted(t *testing.T) {
...
+			name:           "created by controller, no spec.members and no other members",
+			annotations:    map[string]string{common.CreatedByKey: controllerName},
+			expectDeletion: true,
+		},
+		{
+			name:           "created by controller, members in spec.members",
+			annotations:    map[string]string{common.CreatedByKey: controllerName},
+			specMembers:    []string{"other-member"},
+			expectDeletion: false,
+		},
```
jewertow, line 201: Why the following test case below fails?
```
{
	name:           "created by controller, members in spec.members",
	annotations:    map[string]string{common.CreatedByKey: controllerName},
	specMembers:    []string{appNamespace},
	expectDeletion: true,
},
```
I expected that SMMR will be deleted if it was created by a controller and it contains the same namespace as the removed SMM.

jewertow (approved):
I reproduced and verified this bug fix.

pkg/controller/servicemesh/member/controller_test.go
luksa: Because this means that some has added the namespace to `spec.members`. The controller doesn't remove the SMMR, since it has been customized by the user. When the user removes the namespace from `spec.members`, the SMMR will be deleted.

PR #1118, opened 2023-02-16
[wip] OSSM-3248 Use a single global NetworkAttachmentDefinition

luksa:
I implemented this as an experiment for cluster-wide mode, but have decided not to include this change in 2.4. Nevertheless, I'm creating this PR so that we don't lose this work, since we might want this change to go into maistra-next.


dgn:
Let's close this- with any luck we won't be needing network-attachment-defs anymore

PR #1119, opened 2023-02-20
OSSM-3374 Add servicemesh_members metric

bmangoen:
We want to add the Service Mesh Control Plane Namespaces counter metric that would be exposed by Telemetry


pkg/internalmetrics/metrics.go
```diff
@@ -0,0 +1,46 @@
...
+const (
+	labelVersion      = "version"
+	labelNamespace    = "namespace"
+	labelSMCPVersion  = "smcp_version"
+	labelSMCPTopology = "smcp_topology"
+)
+
+// MetricsType defines all of istio-operator's metrics.
+type MetricsType struct {
+	SMCPCounter *prometheus.CounterVec
```
luksa: This should be called something like "MemberCount(er?)", not SMCPCount(er), since it doesn't count the number of SMCPs, but the number of member namespaces.

pkg/apis/maistra/v2/servicemeshcontrolplane_types.go
```diff
@@ -202,3 +202,14 @@ func (s ControlPlaneSpec) IsClusterScoped() (bool, error) {
...
+func (s ControlPlaneSpec) GetControlPlaneMode() (string, error) {
+	isClusterScoped, err := s.IsClusterScoped()
+	if err != nil {
+		return "", err
+	}
+	if isClusterScoped {
+		return ControlPlaneModeValueClusterScoped, nil
+	}
+	return ControlPlaneModeValueMultiTenant, nil
+}
```
luksa: This needs two changes:
- This shouldn't return the two constants, but fixed values "ClusterWide" or "MultiTenant", as the constants in 2.4 will have different values.
- The `s.IsClusterScoped()` call doesn't work when cluster-wide mode is enabled in an SMCP profile. This is a general problem that doesn't just affect this PR. Let's pair on this, so I can help you figure out how to make this work with profiles.

pkg/internalmetrics/metrics.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+// These constants define the different label names for the different metric timeseries
+const (
+	labelVersion      = "version"
+	labelNamespace    = "namespace"
+	labelSMCPVersion  = "smcp_version"
+	labelSMCPTopology = "smcp_topology"
```
luksa: Can we rename this to "mode"? In 2.4, `controlPlaneMode` will be renamed to `mode`, so it makes sense to use the same term in metrics. Is there a reason we're calling this "topology"?

bmangoen: Right I changed it to `smcp_mode` instead

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -413,6 +414,13 @@ func (r *MemberRollReconciler) reconcileObject(ctx context.Context, roll *maistr
 	} else if kialiErr != nil {
 		return reconcile.Result{}, kialiErr
 	}
+	if mesh != nil {
+		meshMode, err := getMeshMode(mesh)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		internalmetrics.GetMemberCounter(mesh.GetNamespace(), mesh.Spec.Version, meshMode).Set(float64(len(roll.Status.Members)))
```
luksa: We may need to consider whether using `roll.status.configuredMembers` makes more sense, because `roll.status.members` includes namespaces that don't exist (but are listed in `spec.members`), whereas `configuredMembers` contains only those namespaces that are actually part of the mesh.

dgn: we might just want to measure both - configured_members and specified_members maybe?

jewertow:
@bmangoen could you rebase this PR?

pkg/controller/servicemesh/memberroll/controller.go
bmangoen: IMO we should get only the namespaces that are actually part of the mesh. It would be a better "picture" of the mesh

dgn:
how best to test this? it looks like it's integrated with cluster-monitoring, but I don't know how to verify that

pkg/internalmetrics/metrics.go
```diff
@@ -0,0 +1,55 @@
...
+	MemberCounter *prometheus.GaugeVec
+}
+
+// Metrics contains all of istio-operator's own internal metrics.
+// These metrics can be accessed directly to update their values, or
+// you can use available utility functions defined below.
+var Metrics = MetricsType{
+	MemberCounter: prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "servicemesh_member_count",
```
simonpasquier: you don't want to name it with a `_count` suffix if it's not a counter. `servicemesh_members` would be a better choice.

simonpasquier: ~~or even `servicemesh_namespaces`.~~

scratching that after reading the code.

pkg/internalmetrics/metrics.go
```diff
@@ -0,0 +1,55 @@
...
+			Help: "Counts the total number of Service Mesh Control Plane Namespaces.",
```
simonpasquier:
```suggestion
			Help: "Number of configured members per namespace, mesh mode and version.",
```

simonpasquier (commented):
I'd suggest to implement a [Collector](https://pkg.go.dev/github.com/prometheus/client_golang@v1.14.0/prometheus#hdr-Custom_Collectors_and_constant_Metrics) interface which is less error-prone from my experience than setting/deleting  the metric on the fly.
Here is an [example](https://github.com/prometheus-operator/prometheus-operator/blob/7b3935ffad3d946f8ec29561a3066dc9f05848aa/pkg/operator/operator.go#L295-L319) of such implementation.

bmangoen:
> how best to test this? it looks like it's integrated with cluster-monitoring, but I don't know how to verify that

@dgn You can check these metrics from the OpenShift web console `Observe > Metrics` but you first need to label the operator namespace `oc label ns openshift-operators openshift.io/cluster-monitoring=true` (it will be done later at the operator install via OLM)

pkg/internalmetrics/metrics.go
bmangoen: Actually it is the number of members in the mesh/meshes. I didn't use a `prometheus.Counter` type because I needed to reset it in specific cases but it is used like a counter

luksa:
@bmangoen is this now ready to be merged or does it need more work? is the metric name now correct?

bmangoen:
@luksa It is ready to be merged IMO. I renamed the metric `service_member_count` as recommended `servicemesh_members`.

luksa:
@bmangoen thanks. Can you update the PR title.  I'll review today.

pkg/internalmetrics/metrics.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// RegisterMetrics must be called at startup to prepare the Prometheus scrape endpoint.
+func RegisterMetrics() {
+	metrics.Registry.MustRegister(
+		Metrics.MeshMembers,
+	)
+}
+
+func GetMeshMembers(smcpNamespace, smcpVersion, smcpMode string) prometheus.Gauge {
```
luksa, line 45: Why is this called `GetMeshMembers` instead of `UpdateMeshMembers` or something similar? I assume that's what it does. I assume it doesn't just return a value, because we never use this value anywhere.

pkg/controller/servicemesh/webhooks/hacks.go
```diff
@@ -185,6 +187,20 @@ func createWebhookResources(ctx context.Context, mgr manager.Manager, log logr.L
...
+	monclient, _ := clientmonitoring.NewForConfig(mgr.GetConfig())
+	if _, err := monclient.PrometheusRules(operatorNamespace).Create(context.TODO(),
+		newPrometheusRule(operatorNamespace,
+			"maistra-operator-prometheusrule",
+			"sum without (smcp_namespace) (servicemesh_members)",
+			"cluster:servicemesh_members:sum"), metav1.CreateOptions{}); err != nil {
+		if errors.IsAlreadyExists(err) {
+			log.Info("Maistra Operator webhook PrometheusRule already exists")
+		} else {
+			return pkgerrors.Wrap(err, "error creating Maistra Operator webhook PrometheusRule")
```
luksa: What if the cluster doesn't support the `PrometheusRule` resource? In that case, we probably shouldn't stop the operator from starting.

pkg/internalmetrics/metrics.go
```diff
@@ -0,0 +1,55 @@
...
+	return Metrics.MeshMembers.With(prometheus.Labels{
+		labelSMCPNamespace: smcpNamespace,
+		labelSMCPVersion:   smcpVersion,
+		labelSMCPMode:      smcpMode,
+	})
+}
+
+func ResetMeshMembers() {
+	Metrics.MeshMembers.Reset()
+}
```
luksa: What exactly does this do? This gets called if *any* SMCP is not found, right? But there is only one metric for the entire cluster, so this means that a problem with a s single SMCP resets the metric for the entire cluster (for other SMCPs, too).

luksa (commented):
I'm not sure I understand this code, so I'd like someone else to take a look and review. @jewertow  can you take a look when you're back?

pkg/internalmetrics/metrics.go
bmangoen: Actually this function looks for the `MeshMembers` metric "filtered" by its specific labels (namespace, version and mode) and returns it. The update of the count is done [here](https://github.com/maistra/istio-operator/blob/746e91ec87496b2f3eda105c60fd11dde74146e8/pkg/controller/servicemesh/memberroll/controller.go#L429)

luksa: Ah, I didn't set the `.Set()` call in that line. I just saw the `.Get()`. Can you move the `Set` to the next line, so it's more visible?

pkg/internalmetrics/metrics.go
bmangoen: Right, the reset deletes the metric of all SMCPs. Should be fixed deleting the metric with specific labels

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -711,3 +722,16 @@ func getNamespaces(members *maistrav1.ServiceMeshMemberList) []string {
 	}
 	return namespaces.List()
 }
+
+// Returns the Service Mesh ControlPlane mode
+// Used for the internal custom metrics
+func getMeshMode(mesh *maistrav2.ServiceMeshControlPlane) (string, error) {
```
bmangoen, line 793: @luksa  Would it be possible to use `SMCP.Status.AppliedSpec.Mode` instead of `mesh.Status.AppliedValues.Istio.GetBool("global.clusterWide")` ?

pkg/controller/servicemesh/webhooks/hacks.go
bmangoen: Should we only log a warning that this `PrometheusRule` is not created?

luksa: Yes, I think so. Log a warning + emit a warning event.

pkg/controller/servicemesh/memberroll/controller.go
luksa: Only if the SMCP version is 2.4. For 2.4, you need to check AppliedSpec.TechPreview.ControlPlaneMode or whatever the name was....

pkg/controller/servicemesh/webhooks/hacks.go
bmangoen: Done

rcernich:
@bmangoen , it looks like this needs to be rebased

pkg/controller/servicemesh/webhooks/hacks.go
```diff
@@ -185,6 +187,23 @@ func createWebhookResources(ctx context.Context, mgr manager.Manager, log logr.L
 		os.Exit(0)
 	}
 
+	log.Info("Creating Maistra Operator webhook PrometheusRule")
+	monclient, _ := clientmonitoring.NewForConfig(mgr.GetConfig())
+	prometheusRule := newPrometheusRule(operatorNamespace,
+		"maistra-operator-prometheusrule",
+		"sum without (smcp_namespace) (servicemesh_members)",
```
jan--f: From the telemetry point of view (see PR https://github.com/openshift/cluster-monitoring-operator/pull/1925) I think `sum by (smcp_version, smcp_mode)  (servicemesh_members)` woul be preferable.
Using `by` is more explicit in terms of what ends up in telemetry and avoids the risk of shipping new metric labels that might get added later to telemetry without further review.

bmangoen: It looks even better with this rule, it displays only the labels we need. Thanks @jan--f for this feedback
image

bmangoen: @jan--f does this `sum by` rule work with `prometheus/client_golang` version `v1.5.1` package?

jewertow:
I would like to have metrics, which will answer to the following questions:
1) How many SMCPs of a particular version are deployed.
2) How many tenants our customers have per cluster.
3) Average, maximum and total number of namespaces included in a mesh.

Current metrics do not provide enough information and we can only conclude which SMCP versions are still used, but not how many of a particular version and how many tenants users have per cluster, because metrics are summed up. Additionally, SMCP version is irrelevant in the context of namespaces included in the mesh, so this will create unnecessary time series.

I suggest to collect the following metrics:
1) `servicemesh_controlplanes{version,mode,namespace}` - we will not export the namespace, but it's necessary to create unique metrics for each SMCP controller.
2) `servicemesh_members{mode,namespace}` - namespace will not be exported but it's necessary to create a unique metrics for each tenant.

And then we have to create the following PrometheusRule:
```
apiVersion: monitoring.coreos.com/v1
kind: PrometheusRule
metadata:
  name: service-mesh-2
  namespace: openshift-operators
spec:
  groups:
  - name: service-mesh-2-control-planes
    rules:
    - expr: |-
        sum by (version,mode) (servicemesh_controlplanes)
      record: 'cluster:servicemesh_controlplanes:sum'
  - name: service-mesh-2-members
    rules:
    - expr: |-
        avg by (mode) (servicemesh_members)
      record: 'cluster:servicemesh_members:avg'
     - expr: |-
        max by (mode) (servicemesh_members)
      record: 'cluster:servicemesh_members:max'
     - expr: |-
        sum by (mode) (servicemesh_members)
      record: 'cluster:servicemesh_members:sum'
```

bmangoen:
> I would like to have metrics, which will answer to the following questions:
> 
> 1. How many SMCPs of a particular version are deployed.
> 2. How many tenants our customers have per cluster.
> 3. Average, maximum and total number of namespaces included in a mesh.
> 
> Current metrics do not provide enough information and we can only conclude which SMCP versions are still used, but not how many of a particular version and how many tenants users have per cluster, because metrics are summed up. Additionally, SMCP version is irrelevant in the context of namespaces included in the mesh, so this will create unnecessary time series.
> 
> I suggest to collect the following metrics:
> 
> 1. `servicemesh_controlplanes{version,mode,namespace}` - we will not export the namespace, but it's necessary to create unique metrics for each SMCP controller.
> 2. `servicemesh_members{mode,namespace}` - namespace will not be exported but it's necessary to create a unique metrics for each tenant.
> 
> And then we have to create the following PrometheusRule:
> 
> ```
> apiVersion: monitoring.coreos.com/v1
> kind: PrometheusRule
> metadata:
>   name: service-mesh-2
>   namespace: openshift-operators
> spec:
>   groups:
>   - name: service-mesh-2-control-planes
>     rules:
>     - expr: |-
>         sum by (version,mode) (servicemesh_controlplanes)
>       record: 'cluster:servicemesh_controlplanes:sum'
>   - name: service-mesh-2-members
>     rules:
>     - expr: |-
>         avg by (mode) (servicemesh_members)
>       record: 'cluster:servicemesh_members:avg'
>      - expr: |-
>         max by (mode) (servicemesh_members)
>       record: 'cluster:servicemesh_members:max'
>      - expr: |-
>         sum by (mode) (servicemesh_members)
>       record: 'cluster:servicemesh_members:sum'
> ```

@jewertow Does this still make sense to implement?

bmangoen:
This won't be implemented as is

PR #1126, opened 2023-03-01
OSSM-3259 Move controlPlaneMode out of SMCP.spec.techPreview

luksa:


pkg/controller/versions/util.go
```diff
@@ -38,6 +38,53 @@ const (
...
+		return append(allErrors, err)
+	}
+	smcps := v2.ServiceMeshControlPlaneList{}
+	err = cl.List(ctx, &smcps)
+	if err != nil {
+		return append(allErrors, err)
+	}
+
+	if isClusterScoped {
+		if len(smcps.Items) > 1 || len(smcps.Items) == 1 && smcps.Items[0].UID != meta.GetUID() {
```
jewertow, line 67: Do I understand correctly that `len(smcps.Items) == 1 && smcps.Items[0].UID == meta.GetUID()` means that an existing SMCP is updated? Maybe is it worth to leave a comment? I had to think a moment to be sure that I understand this condition.

pkg/controller/versions/versions.go
```diff
@@ -110,6 +110,11 @@ type Version interface {
 	// a negative value is returned.  If other is the same version, zero is
 	// returned.
 	Compare(other Version) int
+	// AtLeast returns true if this version is greater or equal to the specified
+	// other version.
+	AtLeast(other Version) bool
+	// LessThan returns true if this version is less than the specified version.
+	LessThan(other Version) bool
```
jewertow, line 117: I needed these functions recently. :+1:

jewertow (approved):
LGTM, but didn't we agree that SMMR will be optional in cluster-wide mode?

luksa:
> LGTM, but didn't we agree that SMMR will be optional in cluster-wide mode?

It will. This is just the first commit. As soon as this one merges, I'll submit two more.

PR #1128, opened 2023-03-16
OSSM-3551 Add metadata stub for 3.0

dgn:
This is really just to get CPaaS builds going. Once we implement proper operator support (or later), we'll need to think about how we manage metadata in the future, as we'll want to maintain separate channels for 2.x and 3.x streams. It also pulls in some newer commits from the 2.4 branch


jwendell (approved):

dgn:
cc @aslakknutsen

PR #1132, opened 2023-03-17
Cluster-wide SMMR changes

luksa:


pkg/apis/maistra/v1/servicemeshmemberroll_types.go
```diff
@@ -58,17 +61,114 @@ type ServiceMeshMemberRollSpec struct {
...
+	// +optional
+	MemberSelector *metav1.LabelSelector `json:"memberSelector,omitempty"`
+
+	// List of namespaces that should be excluded, even if matched by a wildcard
+	// in spec.members or by the memberSelector.
+	// +optional
+	ExcludeNamespaces []string `json:"excludeNamespaces,omitempty"`
+}
+
+func (smmr *ServiceMeshMemberRoll) IsWildcard(ns string) bool {
```
jewertow, line 75: Do we want to allow for wildcard or multiple wildcards + namespaces:
```
spec:
  members:
  - "*"
  - bookinfo
  - "*"
```
It does not make sense and additionally it may result it unexpected corner cases. I think that members should be a wildcard or a list of namespaces and then this function should be:
```suggestion
func (smmr *ServiceMeshMemberRoll) IsWildcard() bool {
	return len(smmr.Spec.Members) == 1 && smmr.Spec.Members[0] == "*"
```

pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -139,7 +139,12 @@ func add(mgr manager.Manager, r *MemberRollReconciler) error {
 			})
 
 			// reconcile ServiceMeshMemberRolls that have the ServiceMeshMember's namespace in spec.members
-			requests = append(requests, toRequests(ctx, mgr.GetClient(), member.Namespace)...)
+			ns := &corev1.Namespace{}
+			err := mgr.GetClient().Get(ctx, types.NamespacedName{Name: member.Namespace}, ns)
+			if err != nil {
+				log.Error(err, "Could not get Namespace")
```
jewertow: Could you add name of this namespace? It will simplify understanding what happens.
Btw. shouldn't error messages be lowercase?

pkg/apis/maistra/v1/servicemeshmemberroll_types.go
luksa: We need to allow wildcards like `kube-*` in `excludeNamespaces`. So, to keep things consistent, we allow them also in `members` (instead of only allowing `*` there).

pkg/controller/servicemesh/memberroll/controller.go
luksa: Fixed. Thanks!

jewertow:
I think that I found bugs. Please look at the following steps:
1. Create SMCP:
```
oc apply -n istio-system -f - <<EOF
apiVersion: maistra.io/v2
kind: ServiceMeshControlPlane
metadata:
  name: basic
spec:
  mode: ClusterWide
  proxy:
    accessLogging:
      file:
        name: /dev/stdout
  security:
    dataPlane:
      mtls: true
  tracing:
    type: None
EOF
```
2. Deploy bookinfo:
```
oc new-project bookinfo
kubectl label namespace bookinfo istio-injection=enabled
oc apply -n bookinfo -f https://raw.githubusercontent.com/maistra/istio/maistra-2.4/samples/bookinfo/platform/kube/bookinfo.yaml
```
3. Deploy sleep:
```
oc new-project sleep
oc apply -n sleep -f https://raw.githubusercontent.com/maistra/istio/maistra-2.4/samples/sleep/sleep.yaml
```
So far, everything works as expected. Bookinfo has sidecars, sleep does not have.

4. Label sleep for injection and redeploy it:
```
kubectl label namespace sleep istio-injection=enabled
oc delete -n sleep -f https://raw.githubusercontent.com/maistra/istio/maistra-2.4/samples/sleep/sleep.yaml
oc apply -n sleep -f https://raw.githubusercontent.com/maistra/istio/maistra-2.4/samples/sleep/sleep.yaml
```
Sleep has sidecar injected, so everything is still correct.

5. Exclude sleep namespace from injection:
```
oc apply -n istio-system -f - <<EOF
apiVersion: maistra.io/v1
kind: ServiceMeshMemberRoll
metadata:
  name: default
spec:
  members:
  - "*"
  excludeNamespaces:
  - istio-operator
  - sleep
EOF
```

6. Redeploy sleep:
```
oc delete -n sleep -f https://raw.githubusercontent.com/maistra/istio/maistra-2.4/samples/sleep/sleep.yaml
oc apply -n sleep -f https://raw.githubusercontent.com/maistra/istio/maistra-2.4/samples/sleep/sleep.yaml
```
Sleep has no sidecar anymore.

7. Configure custom injection label:
```
oc apply -n istio-system -f - <<EOF
apiVersion: maistra.io/v1
kind: ServiceMeshMemberRoll
metadata:
  name: default
spec:
  members:
  - "*"
  excludeNamespaces:
  - istio-operator
  - sleep
  memberSelector:
    matchLabels:
      maistra-injection: enabled
EOF
```

8. Redeploy bookinfo:
```
oc delete -n bookinfo -f https://raw.githubusercontent.com/maistra/istio/maistra-2.4/samples/bookinfo/platform/kube/bookinfo.yaml
oc apply -n bookinfo -f https://raw.githubusercontent.com/maistra/istio/maistra-2.4/samples/bookinfo/platform/kube/bookinfo.yaml
```

Bookinfo still has sidecars, while it does not have `maistra-injection: enabled` label!

9. Deploy sleep in another project:
```
oc new-project test
oc apply -n sleep -f https://raw.githubusercontent.com/maistra/istio/maistra-2.4/samples/sleep/sleep.yaml
```

And sleep got sidecar, while it does not have `maistra-injection: enabled` nor `istio-injection: enabled` label!
I see that this namespace has label `maistra.io/member-of: istio-system`.

luksa:
The behavior is correct. You set `spec.members` to `*`, which means ALL namespaces are in the mesh, regardless of the selector (except `istio-operator` and `sleep`, which are explicitly excluded). To determine the final member namespace set, the union of the two sets (the namespaces in `spec.members` and those that match `spec.memberSelector`) is calculated. Then the excluded namespaces are subtracted from the set.

Step 9 seems to have an error. You're creating the namespace `test`, but applying to `sleep`.

luksa:
To achieve what you're trying to achieve, you should omit `spec.members` and set only `spec.memberSelector` (and optionally `spec.excludeNamespaces`).

luksa:
I've been thinking about the confusion around the combination of `members` and `memberSelector` and now think we should only allow one to be used at a time. Either you use `members` with no wildcards (not even just `*`) or you use `memberSelectors`. I also think we can simply remove `excludeNamespaces`, since `memberSelectors` allow you to exclude using the `NOT IN` operator. The exclusion of system namespaces can be implicit.

jewertow:
I think that we should allow to use (`members=["*"]` and `excludeNamespces`) or only `excludeNamespaces`. We cannot objectively say which option is better, but I would prefer to implement as simple semantics as we can.
I also think that `members=["*"]` + `excludeNamespaces` may be intuitive for Istio users, because for example `traffic.sidecar.istio.io/excludeInboundPorts` follows this rule: "Only applies when all inbound traffic (i.e. '*') is being redirected."

I also was very surprised that `memberSelectors` does not overwrite default selector `istio-inection=enabled`. Currently, I am not able to disable injection for `istio-injection=enabled`. I think that `istio-injection=enabled` is the right default value, but when custom `memberSelectors` is defined, then the default label should be ignored.

You're right that `members` and `memberSelectors` should be sufficient, but `excludeNamespaces` allows to exclude namespaces chosen by a mesh admin, so that project admins are not able to join the mesh by labeling their projects, and I believe that this option may desired by some users.

luksa:
> I think that we should allow to use (`members=["*"]` and `excludeNamespces`) or only `excludeNamespaces`. We cannot objectively say which option is better, but I would prefer to implement as simple semantics as we can. 

But the simplest approach is to have a set-based (as opposed to equality-based) `memberSelector`, since it allows you to both include and exclude namespaces based on labels and/or names. This is also the standard mechanism for including Kubernetes objects in a set, so all Kubernetes users expect a feature like this (which is exactly why customers have asked for it).

> I also was very surprised that `memberSelectors` does not overwrite default selector `istio-inection=enabled`. Currently, I am not able to disable injection for `istio-injection=enabled`. I think that `istio-injection=enabled` is the right default value, but when custom `memberSelectors` is defined, then the default label should be ignored.

It does. You weren't using it correctly. You set `members` to `*` and then expected the selector to filter namespaces. But, as I said in my previous comment, we calculate the union of the two sets. So, if you want the selector to do anything, you shouldn't set `members` to `*`, as that already includes all namespaces. 

> You're right that `members` and `memberSelectors` should be sufficient, but `excludeNamespaces` allows to exclude namespaces chosen by a mesh admin, so that project admins are not able to join the mesh by labeling their projects, and I believe that this option may desired by some users.

That is the only reason to keep `excludeNamespaces`, but you need to keep in mind the following:
- only cluster admins can label namespaces
- cluster admins can use RBAC rules to specify which users can add namespaces to a mesh
- when designing an API, it's better to do it iteratively; Also, it's best not to duplicate functionality (for example, if we add `excludeNamespaces`, there would be two ways of excluding namespaces (the other is through `memberSelector`)).

luksa:
Replaced by #1134.

PR #1136, opened 2023-03-21
OSSM-3291: Promote envoyExtAuthzHttp to GA

jewertow:
Up to OSSM 2.3, external authorization was tech preview feature and could be enabled as follows:
```
apiVersion: maistra.io/v2
kind: ServiceMeshControlPlane
metadata:
  name: basic
spec:
  techPreview:
    meshConfig:
      extensionProviders:
      - name: ext-authz
        envoyExtAuthzHttp:
          service: ext-authz.foo.svc.cluster.local
          port: 8000
          includeRequestHeadersInCheck: ["x-ext-authz"]
```
In OSSM 2.4, this feature will be GA and can be enabled as follows:
```
apiVersion: maistra.io/v2
kind: ServiceMeshControlPlane
metadata:
  name: basic
spec:
  extensionProviders:
  - name: ext-authz
    envoyExtAuthzHttp:
      service: ext-authz.foo.svc.cluster.local
      port: 8000
      includeRequestHeadersInCheck: ["x-ext-authz"]
```


pkg/apis/maistra/conversion/extension_providers.go
```diff
@@ -18,34 +19,153 @@ func populateExtensionProvidersValues(in *v2.ControlPlaneSpec, values map[string
...
 		}
-		prometheus := []map[string]interface{}{
-			{
+		if ext.Prometheus != nil && ext.EnvoyExtAuthzHTTP != nil {
+			return fmt.Errorf("extension provider %s must specify only one type of provider: prometheus or envoyExtAuthzHttp", ext.Name)
+		}
+		if ext.Prometheus != nil {
+			extensionProvidersValues = append(extensionProvidersValues, map[string]interface{}{
+				// TODO: check empty string
 				"name":       ext.Name,
```
luksa: Why do we call this `name` instead of `type`?

In Kubernetes, you usually have a `type` field and then another field to configure the thing (the name of this field matches the value of the `type`, but the first letter is lower-cased).

pkg/apis/maistra/conversion/extension_providers.go
```diff
@@ -18,34 +19,153 @@ func populateExtensionProvidersValues(in *v2.ControlPlaneSpec, values map[string
...
+					})
+				}
+				if rawEnvoyExtAuthzHTTP, ok := extProvider["envoyExtAuthzHttp"]; ok {
+					if envoyExtAuthzHTTP, ok := rawEnvoyExtAuthzHTTP.(map[string]interface{}); ok {
+						extProviderName := extProvider["name"].(string)
+						envoyExtAuthzHTTPConfig := &v2.ExtensionProviderEnvoyExternalAuthorizationHTTPConfig{}
+						if service, ok := envoyExtAuthzHTTP["service"]; ok {
+							envoyExtAuthzHTTPConfig.Service = service.(string)
+						} else {
+							return fmt.Errorf("extension provider envoyExtAuthzHTTP '%s' must specify field 'service'", extProviderName)
```
luksa: Can we add this requirement to the CRD schema?

jewertow: I don't think so, but I'm going to move these validations to validation webhook and then populate* functions will only perform mapping.

pkg/apis/maistra/conversion/extension_providers.go
jewertow: > Why do we call this name instead of type?

Because this 1:1 copy of the upstream API.

pkg/apis/maistra/v2/extension_providers.go
```diff
@@ -19,9 +19,9 @@ type ExtensionProviderEnvoyExternalAuthorizationHTTPConfig struct {
 	// service defined by the Kubernetes service or ServiceEntry.
 	//
 	// Example: "my-ext-authz.foo.svc.cluster.local" or "bar/my-ext-authz.example.com".
-	Service string `json:"service"`
+	Service *string `json:"service"`
 	// REQUIRED. Specifies the port of the service.
-	Port int64 `json:"port"`
+	Port *int64 `json:"port"`
```
luksa: If the port is required, why is this a pointer? By convention, in an API, pointers are used to indicate that a value is optional.

jewertow: Right, I will revert this change.

jewertow: Done

pkg/apis/maistra/conversion/extension_providers.go
jewertow: Done

pkg/apis/maistra/conversion/extension_providers.go
```diff
@@ -18,34 +16,120 @@ func populateExtensionProvidersValues(in *v2.ControlPlaneSpec, values map[string
...
+				}
+				if ext.EnvoyExtAuthzHTTP.IncludeRequestBodyInCheck.PackAsBytes != nil {
+					includeRequestBodyInCheckValues["packAsBytes"] = *ext.EnvoyExtAuthzHTTP.IncludeRequestBodyInCheck.PackAsBytes
+				}
+				envoyExtAuthzHTTPValues["includeRequestBodyInCheck"] = includeRequestBodyInCheckValues
+			}
+			extensionProvidersValues = append(extensionProvidersValues, map[string]interface{}{
+				"name":              ext.Name,
+				"envoyExtAuthzHttp": envoyExtAuthzHTTPValues,
+			})
```
luksa: nit: If you extract `ext.EnvoyExtAuthzHTTP` into a variable, a lot of this code becomes shorter.

luksa: you could also rename `envoyExtAuthzHTTPValues` to `v` or something short like that to make this code a bit faster to read.

luksa: Same in `populateExtensionProvidersConfig`.

luksa (approved):
LGTM, but I'd rename a few variables to make the code easier to read.

pkg/apis/maistra/conversion/extension_providers.go
jewertow: I am an ex-Java dev, so you have to forgive me long variable names.

luksa: ha, yeah, me too. I also used to prefer long variable names, but now I see the benefit of having short names for variables that have limited scope. And with this much code, it makes a real difference if the repetitive variable name is short.

jewertow: Please take a look once again. I renamed variables to be shorter and readable (for ex-Java devs) as well. I also refactored these functions a little bit, so that there are less conditions and indentations.

luksa: I like the `allValues` & `values` solution. And using `provider` instead of `ext`. Can you also rename `includeRequestBodyInCheckValues` to something shorter?

pkg/apis/maistra/conversion/extension_providers.go
```diff
@@ -1,55 +1,134 @@
...
-		if len(rawExtProviders) == 0 {
-			out.ExtensionProviders = []*v2.ExtensionProviderConfig{}
-		}
-		for _, rawExtProvider := range rawExtProviders {
-			if extProvider, ok := rawExtProvider.(map[string]interface{}); ok {
+	rawProviders, ok, err := in.GetSlice("meshConfig.extensionProviders")
+	if err != nil {
+		return err
+	} else if !ok {
+		return nil
```
luksa: if you rename `ok` to `exists` or `found` this code would be a bit clearer. I wondered why we're returning a nil error when it's "not ok". But if this was `if !found { return nil }`, it would be very clear.

jewertow: Done

luksa (approved):

mk46:
> In OSSM 2.4, this feature will be GA and can be enabled as follows:
> 
> ```
> apiVersion: maistra.io/v2
> kind: ServiceMeshControlPlane
> metadata:
>   name: basic
> spec:
>   extensionProviders:
>   - name: ext-authz
>     envoyExtAuthzHttp:
>       service: ext-authz.foo.svc.cluster.local
>       port: 8000
>       includeRequestHeadersInCheck: ["x-ext-authz"]
> ```

Hi @jewertow,
Please update as below
```
apiVersion: maistra.io/v2
kind: ServiceMeshControlPlane
metadata:
  name: basic
spec:
  meshConfig:
    extensionProviders:
    - name: ext-authz
      envoyExtAuthzHttp:
        service: ext-authz.foo.svc.cluster.local
        port: 8000
        includeRequestHeadersInCheck: ["x-ext-authz"]
```

Thanks!

PR #1137, opened 2023-03-21
OSSM 3376 Deprecate fields pilotSecretName and rootCAConfigMapName

mayleighnmyers:
https://issues.redhat.com/browse/OSSM-3376

### DIFF
--- a/pkg/apis/maistra/v2/servicemeshcontrolplane_types.go
+++ b/pkg/apis/maistra/v2/servicemeshcontrolplane_types.go
@@ -204,17 +204,10 @@ func (s ControlPlaneSpec) IsClusterScoped() (bool, error) {
 }
 
 func (s ControlPlaneSpec) IsPilotSecretNameEnabled() bool {
-	return s.Security.CertificateAuthorization.CertManager != nil && s.Security.CertificateAuthorization.CertManager.PilotCertSecretName != ""
-	if s.Security.CertificateAuthorization.CertManager.PilotCertSecretName != nil {
-		return true
-	}
-	return false
+	return s.Security.CertificateAuthority.CertManager != nil && s.Security.CertificateAuthority.CertManager.PilotCertSecretName != ""
 }
 
-func (s ControlPlaneSpec) IsrootCAConfigMapNameEnabled() bool {
-	return s.Security.CertificateAuthorization.CertManager != nil && s.Security.CertificateAuthorization.CertManager.rootCAConfigMapName != ""
-	if s.Security.CertificateAuthorization.CertManager.rootCAConfigMapName != nil {
-		return true
-	}
-	return false
+func (s ControlPlaneSpec) IsRootCAConfigMapNameEnabled() bool {
+	return s.Security.CertificateAuthority.CertManager != nil && s.Security.CertificateAuthority.CertManager.RootCAConfigMapName != "" 
+
 }
--- a/pkg/apis/maistra/v2/servicemeshcontrolplane_types.go
+++ b/pkg/apis/maistra/v2/servicemeshcontrolplane_types.go
@@ -202,3 +202,19 @@ func (s ControlPlaneSpec) IsClusterScoped() (bool, error) {
 	}
 	return controlPlaneMode == ControlPlaneModeValueClusterScoped, nil
 }
+
+func (s ControlPlaneSpec) IsPilotSecretNameEnabled() bool {
+	return s.Security.CertificateAuthorization.CertManager != nil && s.Security.CertificateAuthorization.CertManager.PilotCertSecretName != ""
+	if s.Security.CertificateAuthorization.CertManager.PilotCertSecretName != nil {
+		return true
+	}
+	return false
+}
+
+func (s ControlPlaneSpec) IsrootCAConfigMapNameEnabled() bool {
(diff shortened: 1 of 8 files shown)


pkg/controller/versions/strategy_v2_4.go
```diff
@@ -274,10 +274,41 @@ func (v *versionStrategyV2_4) ValidateRequest(ctx context.Context, cl client.Cli
...
+}
 	return admission.ValidationResponse(true, "")
 }
 
+	func ContainsPilotSecretNameField(smcp metav1.Object) bool {
+		switch s := smcp.(type) {
+	case *v1.ServiceMeshControlPlane:
+		return false
+	case *v2.ServiceMeshControlPlane:
+		return s.Spec.IsPilotSecretNameEnabled()
```
luksa, line 293: I think it would be better if the code in this function was inlined here (i.e. remove the function and move the code here).

pkg/controller/versions/strategy_v2_4.go
```diff
@@ -274,10 +274,41 @@ func (v *versionStrategyV2_4) ValidateRequest(ctx context.Context, cl client.Cli
...
+	default:
+		return false 
+		}
+}
+	func ContainsRootCAConfigMapNameField(smcp metav1.Object) bool {
+		switch s := smcp.(type) {
+	case *v1.ServiceMeshControlPlane:
+		return false
+	case *v2.ServiceMeshControlPlane:
+		return s.Spec.IsRootCAConfigMapNameEnabled()
```
luksa, line 303: Same here.

pkg/controller/versions/strategy_v2_4.go
```diff
@@ -274,10 +274,41 @@ func (v *versionStrategyV2_4) ValidateRequest(ctx context.Context, cl client.Cli
 			return admission.ValidationResponse(false, "a cluster-scoped SMCP may only be created by users with cluster-admin permissions")
 		}
 	}
-
+	hasPilotSecretName := ContainsPilotSecretNameField(smcp)
+	if hasPilotSecretName {
+		return admission.ValidationResponse(false, "SMCP does not allow PilotSecretName field")
```
luksa, line 279: Try to put yourself in the user's shoes. When they get this error message, will they know what's wrong? I think the user might then ask "why doesn't the SMCP allow this field?". It's best if the message itself explains why.

pkg/apis/maistra/v2/servicemeshcontrolplane_types.go
```diff
@@ -202,3 +202,12 @@ func (s ControlPlaneSpec) IsClusterScoped() (bool, error) {
 	}
 	return controlPlaneMode == ControlPlaneModeValueClusterScoped, nil
 }
+
+func (s ControlPlaneSpec) IsPilotSecretNameEnabled() bool {
+	return s.Security.CertificateAuthority.CertManager != nil && s.Security.CertificateAuthority.CertManager.PilotCertSecretName != ""
```
luksa, line 207: `Security` and `CertificateAuthority` are pointers, so you need to check if they're `nil`.

pkg/controller/versions/strategy_v2_4.go
```diff
@@ -274,10 +274,41 @@ func (v *versionStrategyV2_4) ValidateRequest(ctx context.Context, cl client.Cli
...
+}
+	hasRootCAConfigMapName := ContainsRootCAConfigMapNameField(smcp)
+	if hasRootCAConfigMapName {
+	return admission.ValidationResponse(false, "SMCP does not allow RootCAConfigMapName field")
+}
```
luksa, line 284: I noticed you add this validation code to the `ValidateRequest` function, which isn't the best place for it. Instead, I think you should create a `validateSecurity` function and call it in `ValidateV2()`. The `ValidateRequest` function exists for cases where you need to actually check the `admission.Request` object.

pkg/controller/versions/strategy_v2_4.go
mayleighnmyers: I am having a hard time coming up with a more intellect response in this field. Any idea on what to say?

pkg/apis/maistra/v2/servicemeshcontrolplane_types.go
mayleighnmyers: Got this fixed and working :)

pkg/controller/versions/strategy_v2_4.go
luksa: Oh, I re-read the Jira issue and it looks like we shouldn't return an error but a deprecation warning. So, the message could read something like `spec.foo.bar.baz.pilotSecretName is deprecated; will be removed in 2.5.0; use foo.bar.baz.xyz instead`. 

We need to figure out how to emit a warning.

jewertow: > Oh, I re-read the Jira issue and it looks like we shouldn't return an error but a deprecation warning.

I can see in the JIRA issue "...if the deprecated fields exist, a warning should be returned."

luksa: Yes, as I said: we must return a warning not an error.  And we can't do that until we upgrade k8s.io/api to v0.19+.

jewertow: Ah, ok, nevermind. I misunderstood your comment and I thought you mean I specified wrong requirements.

PR #1140, opened 2023-03-23
[maistra-3.0] Pull 2.4 commits in

dgn:
I need these for the 3.0 support work

### DIFF
--- a/deploy/maistra-operator.yaml
+++ b/deploy/maistra-operator.yaml
@@ -10162,6 +10162,29 @@ spec:
             type: object
           spec:
             properties:
+              memberSelector:
+                properties:
+                  matchExpressions:
+                    items:
+                      properties:
+                        key:
+                          type: string
+                        operator:
+                          type: string
+                        values:
+                          items:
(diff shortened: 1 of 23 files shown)


luksa (approved):

jwendell:
if 3.0 is going to be a mirror of 2.4 for a significant amount of time we can setup a job that syncs the branches automatically

dgn:
> if 3.0 is going to be a mirror of 2.4 for a significant amount of time we can setup a job that syncs the branches automatically

I don't think so, this is really just to get something up and running

jwendell:
we can always use the cherrypick labels when merging stuff in 2.4. But we might forget about it...

dgn:
> we can always use the cherrypick labels when merging stuff in 2.4. But we might forget about it...

Yeah, that might be an idea. I'll demo what I have on Tuesday, we can use that as a base for discussion. I'm not sure how long it'll be until we diverge too much for that to work, but I hope it's not going to be long :)

PR #1143, opened 2023-03-24
OSSM-3291: Allow to set `spec.extensionProviders` only in 2.4+

jewertow:
We wanted to make extension providers GA in OSSM 2.4, but currently it's possible to enable this feature for previous versions as well.

### DIFF
--- a/pkg/controller/versions/util.go
+++ b/pkg/controller/versions/util.go
@@ -51,6 +51,13 @@ func checkControlPlaneModeNotSet(spec *v2.ControlPlaneSpec, allErrors []error) [
 	return allErrors
 }
 
+func checkExtensionProvidersNotSet(spec *v2.ControlPlaneSpec, allErrors []error) []error {
+	if spec.ExtensionProviders != nil {
+		return append(allErrors, fmt.Errorf("the spec.extensionProviders field is only supported in version 2.4 and above"))
+	}
+	return allErrors
+}
+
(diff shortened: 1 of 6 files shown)


pkg/controller/versions/util.go
```diff
@@ -239,6 +239,14 @@ func validateProtocolDetection(spec *v2.ControlPlaneSpec, allErrors []error) []e
 	return allErrors
 }
 
+func validateUnsupportedAPIs(version Ver, spec *v2.ControlPlaneSpec, allErrors []error) []error {
+	if spec.ExtensionProviders != nil && !version.AtLeast(V2_4) {
```
luksa:
```suggestion
	if spec.ExtensionProviders != nil && version.LessThan(V2_4) {
```

luksa (approved):
LGTM, but I'd replace the `AtLeast` call.

jewertow:
@luksa shouldn't you add similar validation for `controlPlaneMode`?

luksa:
> @luksa shouldn't you add similar validation for `controlPlaneMode`?

I already check it: https://github.com/Maistra/istio-operator/blob/db6f0e08d62b49a197b2fc83f29ad0ef63ec1175/pkg/controller/versions/util.go#L41

jewertow:
Oh, I missed that check. I will combine my check with that one.

jewertow:
@luksa please take a look now.

luksa (approved):

PR #1145, opened 2023-03-27
Remove conversions of extension providers from v1 to v2

jewertow:
Since we are not able to deploy SMCP v2.x using v1 API (as you can see below), it does not make sense to implement conversion from v1 to v2 for extension providers, so I'm removing it.

```
oc apply -f - <<EOF
apiVersion: maistra.io/v1
kind: ServiceMeshControlPlane
metadata:
  name: basic
  namespace: istio-system
spec:
  istio:
    global:
      enableTracing: false
      meshExpansion:
        enabled: false
        useILB: false
      multiCluster:
        enabled: false
        multiClusterOverrides:
          expansionEnabled: null
          multiClusterEnabled: null
      proxy:
        accessLogFile: /dev/stdout
        tracer: zipkin
    grafana:
      enabled: false
    kiali:
      enabled: false
    meshConfig:
      accessLogFile: /dev/stdout
      enableTracing: false
      extensionProviders:
      - name: prom
        prometheus: {}
      - envoyExtAuthzHttp:
          includeRequestHeadersInCheck:
          - x-ext-authz
          port: 8000
          service: ext-authz.foo.svc.cluster.local
        name: foo-ext-authz
    mixer:
      adapters:
        prometheus:
          enabled: false
    prometheus:
      enabled: false
    telemetry:
      v2:
        prometheus:
          enabled: false
    tracing:
      enabled: false
      provider: none
  profiles:
  - default
  template: default
  version: v2.4
EOF
```

```
Error from server (BadRequest): error when creating "mesh-v1.yaml": admission webhook "smcp.validation.maistra.io" denied the request: must use v2 ServiceMeshControlPlane resource for v2.0+ installations
```


luksa:
Doesn't this conversion get called internally in some places?

jewertow:
> Doesn't this conversion get called internally in some places?

I think it does not. To make sure that I'm right, I created SMCP with extension providers, then I fetched this SMCP as v1 and v2 and didn't get any error in the operator logs.

dgn:
AFAIK in older clusters it could be that the storage version of SMCP is v1 (ie etcd will always convert to v1 before persisting it), so the conversion will be required, otherwise you're losing this information

dgn:
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version

luksa:
In light of why @dng said, I don't think it's safe for us to remove this. I'd prefer to keep it in for 2.4. Is there a good reason why we'd want to remove it?

jewertow:
I agree that it's too late to include this change in 2.4.

> Is there a good reason why we'd want to remove it?

I don't want to waste time on implementing functions that are not used. As I said in the PR description, users are not able to apply SMCP v1 with `version: v2.x`, so converting extension providers from v1 to v2 is pointless.

luksa:
Yes, but it's possible that users still use v1 as the storage version. So even if they create a v2.x SMCP, it gets stored as v1 and then converted from v1 to v2 every time the resource is read from etcd. So if this is indeed the case, we need to make sure that conversion from v1 to v2 works even for extension providers.

jewertow:
> it's possible that users still use v1 as the storage version

Ah right. Now I understand. Thanks! Still so much to learn about k8s internals...

PR #1150, opened 2023-04-05
OSSM-3704: Add support for cacerts secret in kubernets.io/tls format

jewertow:
Istio supports cacerts secret in kubernetes.io/tls format since the release 1.14 and Istio Operator must support this secret type to work properly with certificates provided by cert-manager.


pkg/controller/servicemesh/webhookca/source.go
```diff
@@ -31,14 +31,15 @@ type SecretCABundleSource struct {
...
-	Key        string
+	Keys       []string
 }
 
 var _ CABundleSource = (*SecretCABundleSource)(nil)
 
 func (s *SecretCABundleSource) GetCABundle(ctx context.Context, client client.Client) ([]byte, error) {
 	var caBundle []byte
 	var errList []error
+loop:
```
luksa: maybe call this `outerLoop` since there are two loops?

luksa: or extract the outer loop into a function and use `return` instead of `break`?

pkg/controller/servicemesh/webhookca/source.go
```diff
@@ -50,17 +51,19 @@ func (s *SecretCABundleSource) GetCABundle(ctx context.Context, client client.Cl
...
+			caBundle, ok = secret.Data[key]
+			if !ok {
+				errList = append(errList, fmt.Errorf(
+					"secret %s does not contain root certificate under key %s",
+					namespacedName,
+					key,
+				))
+				continue
+			}
+			break loop
```
luksa: This (and the previous version) looks more complicated that it is. How about:
```suggestion
			var ok bool
			caBundle, ok = secret.Data[key]
			if ok {
				break mainLoop
			}
			errList = append(errList, fmt.Errorf(
				"secret %s does not contain root certificate under key %s",
				namespacedName,
				key,
			))
```

Actually, instead of `break mainLoop`, this really should just `return caBundle, nil`. I don't see why some people are obsessed with having a single `return` at the end of the function. As in this case, it makes things more convoluted.

pkg/controller/servicemesh/webhookca/controller.go
```diff
@@ -49,47 +47,47 @@ var autoRegistrationMap = map[string]CABundleSource{
...
 			},
 		},
 	},
 	istiodWebhookNamePrefix: &SecretCABundleSource{
 		SecretNameKeyPairs: []SecretNameKeyPair{
 			{
 				SecretName: istiodCustomCertSecretName,
-				Key:        common.IstiodCertKey,
+				Keys:       []string{common.IstiodCertKey, common.IstiodTLSSecretCertKey},
 			},
```
luksa: Wouldn't creating two `SecretNameKeyPair` entries here instead of changing  `Key string` to `Keys []string` be a simpler approach? Or is the problem that you can't use the same `SecretName` in two entries (even if the `Key` is different)?

pkg/controller/servicemesh/webhookca/source.go
jewertow: > Actually, instead of `break mainLoop`, this really should just return `caBundle, nil`

I like this approach, I will do it.

> why some people are obsessed with having a single `return` at the end of the function

It's not my preference. I just modified the existing code with the minimal impact.

luksa: I know, I wasn't referring to you, but people in the dev community in general.

pkg/controller/servicemesh/webhookca/controller.go
jewertow: > is the problem that you can't use the same SecretName in two entries

No, there is no obstacle to do it. I implemented this PR at night, so my brain wasn't bright enough... I will revert this change and will just add another secret name and key pair. It will significantly simplify this PR. Then I will be able to revert those nested loops with `break label`.

pkg/controller/servicemesh/webhookca/source.go
luksa: But I honestly think you don't need these changes at all. You could just add a second `SecretNameKeyPair`, as I mentioned above.

pkg/controller/servicemesh/webhookca/controller.go
luksa: Ah, yes. Welcome to fatherhood :)

That's exactly why I need people to review my own PRs. I make silly mistakes like this all the time.

pkg/controller/servicemesh/webhookca/source.go
jewertow: Removed

pkg/controller/servicemesh/webhookca/source.go
jewertow: >  You could just add a second SecretNameKeyPair

Done

luksa (approved):

PR #1155, opened 2023-04-06
OSSM-3235 Ensure operator can't deadlock because istiod isn't running

luksa:
If istiod isn't running, but the validating webhook configuration resource is in place, the operator may not be able to reconcile the SMCP and reset istiod.

Here, we configure the webhook with an objectSelector that excludes resources with the `maistra-version` label. All resources created by the operator for the SMCP contain this label and thus don't trigger the webhook. This allows the operator to complete the reconciliation even if the webhook is offline.


luksa:
Whoever reviews this, please think about the impact of the users being able to bypass the webhook.

dgn (approved):
LGTM. I think this approach is fine as the validating webhook isn't really on the critical path.

PR #1160, opened 2023-04-11
OSSM-1352: Integrate with custom Kiali and external Prometheus

jewertow:
This change will make it possible to integrate Kiali with OpenShift Monitoring or a custom Prometheus Operator and at the same time, Kiali will still be reconciled by OSSM.

Example:
```
apiVersion: kiali.io/v1alpha1
kind: Kiali
metadata:
  name: kiali-user-workload-monitoring
  namespace: istio-system-1
spec:
  external_services:
    istio:
      config_map_name: istio-basic
      istio_sidecar_injector_config_map_name: istio-sidecar-injector-basic
      istiod_deployment_name: istiod-basic
      istiod_pod_monitoring_port: 15014
      url_service_version: 'http://istiod-basic.istio-system-1:15014/version'
    prometheus:
      auth:
        token: secret:thanos-querier-web-token:token
        type: bearer
        use_kiali_token: false
      query_scope:
        mesh_id: "basic_istio-system-1"
      thanos_proxy:
        enabled: true
      url: https://thanos-querier.openshift-monitoring.svc.cluster.local:9091
  version: v1.57
---
apiVersion: maistra.io/v2
kind: ServiceMeshControlPlane
metadata:
  name: basic
spec:
  addons:
    grafana:
      enabled: false
    kiali:
      enabled: true
      name: kiali-user-workload-monitoring
    prometheus:
      enabled: false
  ...
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ REPLACES_COMMUNITY_CSV ?= 2.3.0
 VERSION                ?= development
 CONTAINER_CLI          ?= docker
 COMMUNITY              ?= true
-TEST_TIMEOUT           ?= 5m
+TEST_TIMEOUT           ?= 10m
 TEST_FLAGS             ?=
 
 SOURCE_DIR          := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
--- a/pkg/apis/maistra/v2/servicemeshcontrolplane_types.go
+++ b/pkg/apis/maistra/v2/servicemeshcontrolplane_types.go
@@ -204,6 +204,11 @@ func (s ControlPlaneSpec) IsKialiEnabled() bool {
 		*s.Addons.Kiali.Enabled
 }
 
+func (s ControlPlaneSpec) IsCustomKialiConfigured() bool {
+	return s.Addons != nil && s.Addons.Kiali != nil &&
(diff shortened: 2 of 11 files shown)


Makefile
```diff
@@ -26,7 +26,7 @@ REPLACES_COMMUNITY_CSV ?= 2.3.0
 VERSION                ?= development
 CONTAINER_CLI          ?= docker
 COMMUNITY              ?= true
-TEST_TIMEOUT           ?= 5m
```
dgn, line 29: really? :cry:

dgn (approved):
lgtm except the timeout extension. is that really required? and I guess there needs to be a discussion whether we want to merge this, with the release just a few days away. It appears to be low-risk, so I'd be fine with it.

Makefile
jewertow: I didn't make our tests slower. Current implementation is very slow and I just increased the coverage by adding ~25 tests and therefore timeout is too short, so I had to increase it. We will have to improve it soon, but we can do it after the release.

jshaughn (commented):
I think this looks OK, a pass-thru for Kiali configuration to the Kiali CR, if I understand correctly.

PR #1161, opened 2023-04-12
[OSSM-1094] Use bcrypt for htpasswd

cam-garrison:
This PR makes it so that the htpasswd secret used to connect prometheus and grafana is hashed using `bcrypt` rather than `SHA-1`, as `SHA-1` is deprecated. In addition to changing from SHA-1 to bcrypt,  the Pod template is changed so only the htpasswd is mounted in the container instead of the htpasswd and the raw password as it was previously.

This change is only applied to v2.4 SMCPs created, to avoid breaking pre-existing deployments, and keep the `oauth-proxy` version consistent with older SMCP versions.

Tested:
Created v2.4 SMCP and run 
`oc get secret -n istio-system htpasswd -o json | jq .data.auth | tr -d \" | base64 -d | sed 's/}.*/}REDACTED\n/'`

Ensuring the secret begins with `$2a$`, indicating it's been hashed with bcrypt. Ensured that in `grafana-proxy` and `prometheus-proxy` containers, `/etc/proxy/htpasswd` only contains the hashed htpasswd, not the raw password as well. 

Created v2.3, v2.2 SMCPs and ensured that behavior remained the same as before.

### DIFF
--- a/manifests-servicemesh/2.4.0/servicemeshoperator.v2.4.0.clusterserviceversion.yaml
+++ b/manifests-servicemesh/2.4.0/servicemeshoperator.v2.4.0.clusterserviceversion.yaml
@@ -17,7 +17,7 @@ metadata:
       The OpenShift Service Mesh Operator enables you to install, configure, and manage an instance of Red Hat OpenShift Service Mesh. OpenShift Service Mesh is based on the open source Istio project.
 
     containerImage: ${OSSM_OPERATOR_IMAGE}
-    createdAt: 2023-04-06T13:11:56CEST
+    createdAt: 2023-04-11T11:59:34EDT
     support: Red Hat, Inc. 
     olm.skipRange: ">=1.0.2 <2.4.0-0"
     operators.openshift.io/infrastructure-features: '["Disconnected"]'
@@ -562,7 +562,7 @@ spec:
                 oauth-proxy.query: "true"
                 oauth-proxy.namespace: openshift
                 oauth-proxy.name: oauth-proxy
-                oauth-proxy.tag: v4.4
+                oauth-proxy.tag: v4.9
             spec:
               serviceAccountName: istio-operator
               containers:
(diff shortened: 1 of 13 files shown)


pkg/controller/servicemesh/controlplane/htpasswd.go
```diff
@@ -56,6 +67,29 @@ func (r *controlPlaneInstanceReconciler) patchHtpasswdSecret(ctx context.Context
 	return nil
 }
 
+func hashPassword(version versions.Version, rawPass string) (string, error) {
+	var auth, hashedPassword string
+	username := "internal"
+
+	// If the SMCP version is below v2.4, we use SHA-1 to keep behavior of old SMCP versions consistent,
+	// and to keep oauth-proxy version consistent at version 4.4 in those releases.
+	if versions.V2_4.Compare(version) > 0 {
```
jewertow:
```suggestion
	if version.AtLeast(V2_4) {
```
And then move code from `else` branch to this condition.

manifests-servicemesh/2.4.0/servicemeshoperator.v2.4.0.clusterserviceversion.yaml
```diff
@@ -17,7 +17,7 @@ metadata:
       The OpenShift Service Mesh Operator enables you to install, configure, and manage an instance of Red Hat OpenShift Service Mesh. OpenShift Service Mesh is based on the open source Istio project.
 
     containerImage: ${OSSM_OPERATOR_IMAGE}
-    createdAt: 2023-04-06T13:11:56CEST
+    createdAt: 2023-04-11T11:59:34EDT
```
jewertow, line 20: @jwendell when do we update `createdAt`?

jewertow:
I tested this PR and can confirm that it works.

jewertow (approved):

manifests-servicemesh/2.4.0/servicemeshoperator.v2.4.0.clusterserviceversion.yaml
luksa: only when there are actual changes in the file

pkg/controller/servicemesh/controlplane/htpasswd.go
cam-garrison: made change in [7bd48fc](https://github.com/maistra/istio-operator/pull/1161/commits/7bd48fc092911eb7f8746dc3c724debaab355351) :+1:

PR #1164, opened 2023-04-13
OSSM-3718 Revert CNI resource renaming for all versions except 2.4

dgn:
This reverts the name changes for all istio-cni versions < 2.4 and adds new templates for v2.4 resources that will use the new name prefix `ossm-cni` by default


luksa (approved):
Looks good. Tested bookinfo injection against 2.2, 2.3, and 2.4. All CNI pods come up fine.

Thanks @dgn, you taking care of this was really helpful!

PR #1170, opened 2023-04-16
[maistra-2.1] OSSM-3747: Remove duplicated env vars from gateway templates

jewertow:
Follow-up to #1169.


jewertow:
@luksa I forgot that before I merge #1169, I have to patch previous charts.

jewertow:
I forgot that 2.1 will no longer be fully supported after the release of 2.4, so we don't need this bug fix.

PR #1176, opened 2023-04-19
OSSM-3802 Implement SMCP.spec.meshConfig.discoverySelectors

mayleighnmyers:
https://issues.redhat.com/browse/OSSM-3802

This PR allows us to make discoverySelectors GA

### DIFF
--- a/pkg/apis/maistra/conversion/conversion_v1_to_v2.go
+++ b/pkg/apis/maistra/conversion/conversion_v1_to_v2.go
@@ -228,6 +228,11 @@ func Convert_v1_ControlPlaneSpec_To_v2_ControlPlaneSpec(in *v1.ControlPlaneSpec,
 		if err := populateExtensionProvidersConfig(values, out); err != nil {
 			return err
 		}
+
+		// Discovery Selectors
+		if err := populateDiscoverySelectorsConfig(values, out); err != nil {
+			return err
+		}
 	}
 
 	// Runtime
--- a/pkg/apis/maistra/conversion/conversion_v2_to_v1.go
+++ b/pkg/apis/maistra/conversion/conversion_v2_to_v1.go
@@ -218,6 +218,11 @@ func Convert_v2_ControlPlaneSpec_To_v1_ControlPlaneSpec(in *v2.ControlPlaneSpec,
 		if err := populateExtensionProvidersValues(in, values); err != nil {
 			return err
 		}
+
+		// Discovery Selectors
+		if err := populateDiscoverySelectorsValues(in, values); err != nil {
+			return err
+		}
 	}
 
 	// Runtime - must run last as this will add values to existing child maps
--- a/pkg/apis/maistra/conversion/discovery_selectors.go
+++ b/pkg/apis/maistra/conversion/discovery_selectors.go
@@ -1,0 +1,41 @@
+package conversion
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	v1 "github.com/maistra/istio-operator/pkg/apis/maistra/v1"
+	v2 "github.com/maistra/istio-operator/pkg/apis/maistra/v2"
+)
+
+func populateDiscoverySelectorsValues(in *v2.ControlPlaneSpec, out map[string]interface{}) error {
+	if in.MeshConfig == nil || in.MeshConfig.DiscoverySelectors == nil {
+		return nil
+	}
+
+	untypedSlice := make([]interface{}, len(in.MeshConfig.DiscoverySelectors))
+	for index, value := range in.MeshConfig.DiscoverySelectors {
+		untypedSlice[index] = value
+	}
+	if discoverySelectors, err := sliceToValues(untypedSlice); err == nil {
+		return setHelmValue(out, "meshConfig.discoverySelectors", discoverySelectors)
+	} else {
+		return err
+
+	}
+}
+
+func populateDiscoverySelectorsConfig(in *v1.HelmValues, out *v2.ControlPlaneSpec) error {
+	var discoverySelectors []*metav1.LabelSelector
+	if ds, ok, err := in.GetAndRemoveSlice("meshConfig.discoverySelectors"); ok {
+		if err := fromValues(ds, &discoverySelectors); err != nil {
+			return err
+		}
+		if out.MeshConfig == nil {
+			out.MeshConfig = &v2.MeshConfig{}
+		}
+		out.MeshConfig.DiscoverySelectors = discoverySelectors
+	} else if err != nil {
+		return err
+	}
+	return nil
+}
--- a/pkg/apis/maistra/conversion/discovery_selectors.go
+++ b/pkg/apis/maistra/conversion/discovery_selectors.go
@@ -20,7 +20,6 @@ func populateDiscoverySelectorsValues(in *v2.ControlPlaneSpec, out map[string]in
 		return setHelmValue(out, "meshConfig.discoverySelectors", discoverySelectors)
 	} else {
 		return err
-
 	}
 }
 
--- a/pkg/apis/maistra/conversion/discovery_selectors_test.go
+++ b/pkg/apis/maistra/conversion/discovery_selectors_test.go
@@ -1,0 +1,117 @@
+package conversion
+
+import (
+	"reflect"
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	v1 "github.com/maistra/istio-operator/pkg/apis/maistra/v1"
+	v2 "github.com/maistra/istio-operator/pkg/apis/maistra/v2"
+	"github.com/maistra/istio-operator/pkg/controller/versions"
+)
+
+var discoverySelectorsTestCases []conversionDiscoverySelectorsTestCase
+
+type conversionDiscoverySelectorsTestCase struct {
+	name       string
+	spec       *v2.ControlPlaneSpec
+	helmValues string
+}
+
+func init() {
+	for _, v := range versions.AllV2Versions {
+		if v.AtLeast(versions.V2_4) {
+			discoverySelectorsTestCases = append(discoverySelectorsTestCases, discoverySelectorsTestCasesV2(v)...)
+		}
+	}
+}
+
+func TestDiscoverySelectorsConversionFromV2(t *testing.T) {
+	for _, tc := range discoverySelectorsTestCases {
+		t.Run(tc.name, func(t *testing.T) {
+			specCopy := tc.spec.DeepCopy()
+			actualHelmValues := v1.NewHelmValues(make(map[string]interface{}))
+			if err := populateDiscoverySelectorsValues(specCopy, actualHelmValues.GetContent()); err != nil {
+				t.Errorf("error converting to values: %s", err)
+			}
+
+			expectedHelmValues := v1.HelmValues{}
+			if err := expectedHelmValues.UnmarshalYAML([]byte(tc.helmValues)); err != nil {
+				t.Fatalf("failed to parse helm values: %s", err)
+			}
+			if !reflect.DeepEqual(expectedHelmValues.DeepCopy(), actualHelmValues.DeepCopy()) {
+				t.Errorf("unexpected output converting v2 to values:\n\texpected:\n%#v\n\tgot:\n%#v", expectedHelmValues.GetContent(), actualHelmValues.GetContent())
+			}
+			specv2 := v2.ControlPlaneSpec{}
+			if err := populateDiscoverySelectorsConfig(expectedHelmValues.DeepCopy(), &specv2); err != nil {
+				t.Errorf("error converting from values: %s", err)
+			}
+			assertEquals(t, tc.spec.MeshConfig, specv2.MeshConfig)
+		})
+	}
+}
+
+func discoverySelectorsTestCasesV2(version versions.Version) []conversionDiscoverySelectorsTestCase {
+	ver := version.String()
+	return []conversionDiscoverySelectorsTestCase{
+		{
+			name: "nil." + ver,
+			spec: &v2.ControlPlaneSpec{
+				Version: ver,
+			},
+			helmValues: "{}",
+		},
+		{
+			name: "empty." + ver,
+			spec: &v2.ControlPlaneSpec{
+				Version: ver,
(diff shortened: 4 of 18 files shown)


pkg/apis/maistra/v2/meshconfig.go
```diff
@@ -1,9 +1,34 @@
...
+	//   - matchExpressions:
+	//     - key: app
+	//       operator: In
+	//       values:
+	//         - cassandra
+	//         - spark
+	// ```
+	// Refer to the [kubernetes selector docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors)
+	// for additional detail on selector semantics.
+	DiscoverySelectors []*v1.LabelSelector `json:"discovery_selectors,omitempty"`
```
jewertow:
```suggestion
	DiscoverySelectors []*v1.LabelSelector `json:"discoverySelectors,omitempty"`
```

resources/helm/v2.4/istio-control/istio-discovery/files/gateway-injection-template.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
     maistra-version: "2.4.0"
     service.istio.io/canonical-name: {{ index .ObjectMeta.Labels `service.istio.io/canonical-name` | default (index .ObjectMeta.Labels `app.kubernetes.io/name`) | default (index .ObjectMeta.Labels `app`) | default .DeploymentMeta.Name  | quote }}
     service.istio.io/canonical-revision: {{ index .ObjectMeta.Labels `service.istio.io/canonical-revision` | default (index .ObjectMeta.Labels `app.kubernetes.io/version`) | default (index .ObjectMeta.Labels `version`) | default "latest"  | quote }}
-    istio.io/rev: {{ index .ObjectMeta.Labels `istio.io/rev` | default .Revision | default "default" | quote }}
+    istio.io/rev: {{ .Revision | default "default" | quote }}
```
jewertow: Please revert this change.

jewertow (changes requested):
This PR lacks an implementation of v1 / v2 API conversion.

pkg/apis/maistra/conversion/discovery_selectors_test.go
```diff
@@ -0,0 +1,78 @@
...
+				MeshConfig: &v2.MeshConfig{
+					DiscoverySelectors: []*metav1.LabelSelector{},
+				},
+			},
+			helmValues: `
+meshConfig:
+  discoverySelectors: []
+`,
+		},
+	}
```
luksa, line 116: Looks like you lost the additional test that you added after our meeting.

luksa (changes requested):

pkg/apis/maistra/conversion/discovery_selectors.go
```diff
@@ -0,0 +1,51 @@
...
+	v2 "github.com/maistra/istio-operator/pkg/apis/maistra/v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func populateDiscoverySelectorsValues(in *v2.ControlPlaneSpec, out map[string]interface{}) error {
+	if in.MeshConfig == nil || in.MeshConfig.DiscoverySelectors == nil {
+		return nil
+	}
+
+	if len(in.MeshConfig.DiscoverySelectors) > 0 {
```
luksa: You need to remove this if statement (just this line, not the entire if block) to make the `empty` test pass.

luksa:
@mayleighnmyers FYI I ran `make gen` for you and commited the changes to your PR, so be sure to pull them before you continue.

pkg/apis/maistra/conversion/discovery_selectors_test.go
```diff
@@ -0,0 +1,117 @@
...
+				MeshConfig: &v2.MeshConfig{
+					DiscoverySelectors: []*metav1.LabelSelector{
+						{
+							MatchLabels: map[string]string{
+								"env":    "prod",
+								"region": "us-east1",
+							},
+						},
+						{
+							MatchExpressions: []*metav1.LabelSelectorRequirement{
```
jewertow:
```suggestion
							MatchExpressions: []metav1.LabelSelectorRequirement{
```

pkg/apis/maistra/conversion/discovery_selectors_test.go
```diff
@@ -0,0 +1,117 @@
...
+      region: us-east1
+  - matchExpressions:
+    - key: app
+      operator: In
+      values:
+        - cassandra
+        - spark
+`,
+		},
+	},
```
jewertow:
```suggestion
	}
```

pkg/apis/maistra/conversion/discovery_selectors_test.go
```diff
@@ -0,0 +1,117 @@
...
+							MatchLabels: map[string]string{
+								"env":    "prod",
+								"region": "us-east1",
+							},
+						},
+						{
+							MatchExpressions: []*metav1.LabelSelectorRequirement{
+								{
+									Key:      "app",
+									Operator: metav1.LabelSelectorOperatorIn,
```
jewertow:
```suggestion
									Operator: metav1.LabelSelectorOpIn,
```

jewertow (changes requested):
Tests fail on compilation.

luksa (commented):
LGTM, but I'd like @jewertow to double-check.

pkg/apis/maistra/conversion/discovery_selectors.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	if discoverySelectors, err := sliceToValues(untypedSlice); err == nil {
+		if err := setHelmValue(out, "meshConfig.discoverySelectors", discoverySelectors); err != nil {
+			return err
+		}
+	} else {
+		return err
+	}
+
+	return nil
```
jewertow: You can simplify these conditions as below:
```suggestion
	if discoverySelectors, err := sliceToValues(untypedSlice); err == nil {
		return setHelmValue(out, "meshConfig.discoverySelectors", discoverySelectors)
	} else {
		return err
	}
```

pkg/apis/maistra/conversion/discovery_selectors.go
```diff
@@ -0,0 +1,51 @@
...
+	} else if err != nil {
+		return err
+	}
+
+	if setDiscoverySelectors {
+		if out.MeshConfig == nil {
+			out.MeshConfig = &v2.MeshConfig{}
+		}
+		out.MeshConfig.DiscoverySelectors = discoverySelectors
+	}
```
jewertow: You can remove this boolean variable and simplify function as below:
```suggestion
        var discoverySelectors []*metav1.LabelSelector
	if ds, ok, err := in.GetAndRemoveSlice("meshConfig.discoverySelectors"); ok {
		if err := fromValues(ds, &discoverySelectors); err != nil {
			return err
		}
		if out.MeshConfig == nil {
			out.MeshConfig = &v2.MeshConfig{}
		}
		out.MeshConfig.DiscoverySelectors = discoverySelectors
	} else if err != nil {
		return err
	}
```

jewertow (changes requested):
Two blockers:
1. Functions `populateDiscoverySelectorsValues` and `populateDiscoverySelectorsConfig` are not used.
2. We need validation that will allow to set `discoverySelectors` only in cluster-wide mode. Look at this PR as reference: https://github.com/maistra/istio-operator/pull/1143.

pkg/apis/maistra/conversion/conversion_v1_to_v2.go
```diff
@@ -230,6 +230,13 @@ func Convert_v1_ControlPlaneSpec_To_v2_ControlPlaneSpec(in *v1.ControlPlaneSpec,
 		}
 	}
 
+	// Discovery Selectors
+	if version.AtLeast(versions.V2_4) {
+		if err := populateDiscoverySelectorsConfig(values, out); err != nil {
+			return err
+		}
+	}
```
luksa: You can merge this if statement with the one for extension providers. No need to have two if statements that check the version.

pkg/apis/maistra/conversion/conversion_v2_to_v1.go
```diff
@@ -220,6 +220,13 @@ func Convert_v2_ControlPlaneSpec_To_v1_ControlPlaneSpec(in *v2.ControlPlaneSpec,
 		}
 	}
 
+	// Discovery Selectors
+	if version.AtLeast(versions.V2_4) {
+		if err := populateDiscoverySelectorsValues(in, values); err != nil {
+			return err
+		}
+	}
```
luksa: Same here.

pkg/apis/maistra/conversion/discovery_selectors.go
```diff
@@ -0,0 +1,42 @@
...
+
+	untypedSlice := make([]interface{}, len(in.MeshConfig.DiscoverySelectors))
+	for index, value := range in.MeshConfig.DiscoverySelectors {
+		untypedSlice[index] = value
+	}
+	if discoverySelectors, err := sliceToValues(untypedSlice); err == nil {
+		return setHelmValue(out, "meshConfig.discoverySelectors", discoverySelectors)
+	} else {
+		return err
+
```
luksa: This empty line is not needed.

pkg/apis/maistra/conversion/discovery_selectors.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
```
luksa: This empty line is not needed.

pkg/apis/maistra/conversion/discovery_selectors.go
```diff
@@ -0,0 +1,42 @@
...
+			return err
+		}
+		if out.MeshConfig == nil {
+			out.MeshConfig = &v2.MeshConfig{}
+		}
+		out.MeshConfig.DiscoverySelectors = discoverySelectors
+	} else if err != nil {
+		return err
+	}
+
```
luksa: You need to `return nil` in this line. This will resolve the syntax error you're probably seeing here or in the next line.

luksa:
@mayleighnmyers You still need to implement #2 in @jewertow 's comment: 

> We need validation that will allow to set discoverySelectors only in cluster-wide mode. Look at this PR as reference: https://github.com/maistra/istio-operator/pull/1143.

luksa:
@mayleighnmyers You need to rebase to upstream/maistra-2.4 to get Jacek's changes from #1143 in. Then add a similar check that he made for extensionProviders, but for discoverySelectors. The reason you don't see his code is because you created your branch before his changes were merged.

To rebase, run the following command: `git pull --rebase upstream maistra-2.4`

jewertow:
Please reset to 10f45c7af0b395a0adaef96b0bdd1beccb6b2ad9 and then rebase.

pkg/controller/versions/util.go
```diff
@@ -58,6 +58,12 @@ func checkMeshConfigNotSet(spec *v2.ControlPlaneSpec, allErrors []error) []error
 	return allErrors
 }
 
+func checkDiscoverySelectorsNotSet(spec *v2.ControlPlaneSpec, allErrors []error) []error {
+	if spec.MeshConfig.DiscoverySelectors != nil {
+		return append(allErrors, fmt.Errorf("the spec.MeshConfig.DiscoverySelectors field is only supported in version 2.4 and above"))
```
jewertow:
```suggestion
		return append(allErrors, fmt.Errorf("the spec.meshConfig.discoverySelectors field is only supported in version 2.4 and above"))
```

luksa:
@mayleighnmyers I squashed all the commits and rebased to `maistra-2.4`, so this PR is now easier to work with.

pkg/controller/versions/util.go
```diff
@@ -92,6 +98,12 @@ func validateGlobal(ctx context.Context, version Ver, meta *metav1.ObjectMeta, s
 	}
 	return allErrors
 }
+func checkDiscoverySelectorsisClusterWide(spec *v2.ControlPlaneSpec, Version, allErrors []error) []error {
+	if spec.MeshConfig.DiscoverySelectors != nil {
+		return append(allErrors, fmt.Errorf("the spec.meshConfig.discoverySelectors must be set for ClusterWide"))
+	}
+	return allErrors
+}
```
luksa: No, this function is checking the wrong thing. It needs to return an error only when `mode` is `MultiTenant` and `discoverySelectors` are non-nil. If `mode` is `ClusterWide`, then `discoverySelectors` *may* be set (but don't need to be).

mayleighnmyers: So would I change ClusterWide to MultiTenant?

luksa: I'd simply name the function `checkDiscoverySelectors`. As for the actual check: it's not really simple, as checking `spec.Mode` is not enough, because `mode` can be set in an SMCP profile/template (i.e. `mode` might be nil in the `spec`). But we probably have similar bugs elsewhere. 

@jewertow Why do we need to disallow the use of discoverySelectors in MultiTenant mode? Even if we use xns-informer, the selectors would still work, wouldn't they? Also, we allowed users to use `techPreview.meshConfig.discoverySelectors`, did we not?

If anything, we should disallow the use of `techPreview.meshConfig.discoverySelectors` and tell users to use `meshConfig.discoverySelectors` instead.

mayleighnmyers: @luksa So I changed the name of the function, but a little confused on what I should do for the `spec.mode` or is this something that will eventually be changed later since we might have it used this way other places?

luksa: @mayleighnmyers let's wait for @jewertow to explain why we shouldn't allow discoverySelectors in multi-tenant mode. If it turns out we can allow them, we can just remove the function altogether.

luksa: Actually, it would be great if you could try this out (set mode: MultiTenant and specify spec.meshConfig.discoverySelectors). The best way would be to add another test case to maistra-test-tool (like the one in https://github.com/maistra/maistra-test-tool/pull/519).

luksa: I tested MultiTenant with discoverySelectors and realized we probably don't want to allow them, as Jacek suggested. So I implemented this in the validating webhook.

@jewertow Can you review this, so we can get it merged?

luksa (approved):

jewertow (approved):

PR #1191, opened 2023-05-11
OSSM-3937 Remove the 2.1 SMCP option from the 2.4 SM drop-down in OS UI

mayleighnmyers:


luksa:
@mayleighnmyers You need to run `make gen` and then commit the changes.

PR #1193, opened 2023-05-16
OSSM-2285: Remove support for v2.0

jewertow:
Apart from removing v2.0 containers from manifests I also refactored tests so that we only execute them for v2.2+, because v2.1 is no longer supported and will not be patched.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ test:
 # Helm charts generation and templates processing
 ################################################################################
 
-SUPPORTED_VERSIONS := 2.0 2.1 2.2 2.3 2.4
+SUPPORTED_VERSIONS := 2.1 2.2 2.3 2.4
 
 $(addprefix update-remote-maistra-,$(SUPPORTED_VERSIONS)): update-remote-maistra-%:
 	$(eval version:=$*)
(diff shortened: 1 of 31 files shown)


Makefile
```diff
@@ -87,7 +87,7 @@ test:
 # Helm charts generation and templates processing
 ################################################################################
 
-SUPPORTED_VERSIONS := 2.0 2.1 2.2 2.3 2.4
+SUPPORTED_VERSIONS := 2.1 2.2 2.3 2.4
```
dgn, line 90: Shouldn't 2.1 be removed from here as well?

jewertow: Oh, I forgot about it.

jewertow: We decided to not remove 2.1 related code entirely, so I will not change this line.

pkg/controller/versions/versions.go
```diff
@@ -38,6 +38,7 @@ const (
 
 var (
 	AllV2Versions  = []Version{V2_0, V2_1, V2_2, V2_3, V2_4}
+	TestedVersions = []Version{V2_2, V2_3, V2_4}
```
jewertow, line 41: I introduced a list of versions which we want to test and which we in fact support.

dgn (approved):

PR #1196, opened 2023-05-24
OSSM-4014 Add Make target that downloads charts

dgn:
I took the original script from the 2.4 branch and stripped it down a little bit to make it less confusing. Also, it will now download and extract into a tmp directory, so we avoid the need to run `make clean` everytime- if you have worked on the old operator, you might have encountered the situation that you were working with outdated charts because an old zip file was still around somewhere.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,15 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.1
+VERSION ?= 3.0.0
+MINOR_VERSION := $(shell v='$(VERSION)'; echo "$${v%.*}")
+
+# Istio repository to pull charts from
+ISTIO_REPOSITORY ?= maistra/istio
+# the branch to use when updating the commit hash below
+ISTIO_BRANCH_30 ?= maistra-3.0
+# the current commit of ${ISTIO_REPOSITORY} for v3.0. This will be used to copy charts
+ISTIO_COMMIT_30 ?= 16ff353b180a1f461bd380be6ef2d1809b8c1fba
 
 # CHANNELS define the bundle channels used in the bundle.
 # Add a new line here if you would like to change its default config. (E.g CHANNELS = "candidate,fast,stable")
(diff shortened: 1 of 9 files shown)


dgn:
unit test fix in https://github.com/openshift/release/pull/39693

Makefile
```diff
@@ -3,7 +3,11 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.1
+VERSION ?= 3.0.0
+MINOR_VERSION := $(shell v='$(VERSION)'; echo "$${v%.*}")
+
+MAISTRA_REPOSITORY ?= maistra/istio
```
jwendell: I'd rather rename MAISTRA here and in all places to ISTIO. Maistra is the org, might be confusing.

Makefile
```diff
@@ -3,7 +3,11 @@
...
+MAISTRA_30_COMMIT ?= 16ff353b180a1f461bd380be6ef2d1809b8c1fba
```
jwendell: Why not reference a branch and use the HEAD?

Then you can replace wget&unzip below with a shallow git clone

hack/download-charts.sh
```diff
@@ -0,0 +1,35 @@
...
+HELM_DIR="${REPO_ROOT}/resources/charts/${MAISTRA_RELEASE_STREAM}"
+
+ISTIO_FILE="${MAISTRA_COMMIT}.zip"
+ISTIO_URL="${MAISTRA_REPO}/archive/${MAISTRA_COMMIT}.zip"
+WORK_DIR=$(mktemp -d)
+EXTRACT_DIR="${MAISTRA_REPO##*/}-${MAISTRA_COMMIT}"
+EXTRACT_CMD="unzip -q ${ISTIO_FILE} ${EXTRACT_DIR}/manifests/charts/* ${EXTRACT_DIR}/manifests/addons/dashboards/*"
+
+function downloadIstioCharts() {
+  if [ -d "${HELM_DIR}" ] ; then
```
jwendell: superfluous. You can `rm -rf` directly without checking.

Makefile
dgn: this way we can do both. I would actually like to get away from using the branch name because that created inconsistencies in the past- whenever there was a new commit on maistra/istio, suddenly the istio-operator repo would break until charts were updated. If we're using the commit, that doesn't happen. It does require that the commit is updated when new charts are pulled in, but that's fine IMO, it will be done by prow anyway

hack/download-charts.sh
dgn: fixed

Makefile
dgn: good point, did that now. also added comments

dgn:
push-image job is fixed in https://github.com/openshift/release/pull/39722 -- that PR also adds the postsubmit job that updates the commit hash when maistra/istio is updated

Makefile
dgn: FYI I added the respective prow job, see https://github.com/openshift/release/pull/39722

jwendell (approved):

PR #1207, opened 2023-06-08
OSSM-3709 Use file for getting github token

bmangoen:


hack/operatorhub/publish-bundle.sh
```diff
@@ -41,8 +41,14 @@ done
 CUR_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 BUNDLE_DIR="${CUR_DIR}"/../../bundle/
 
-GITHUB_TOKEN="${GITHUB_TOKEN:-}"
 GIT_USER="${GIT_USER:-}"
+GITHUB_FILE="${GITHUB_FILE:-}"
+
+if [[ ! $dryRun || ! -f "${GITHUB_FILE}" ]]; then
```
jwendell: `&&` instead of `||`?

`-r` instead of `-f`?

bmangoen: Actually we even can delete the `! $dryRun` condition because we must have this variable `GITHUB_FILE` well set in any case. Agree for using `-r`

jwendell (changes requested):
What is wrong with the env variable approach?

Holding. I made a comment on the other PR

Makefile
```diff
@@ -283,7 +283,7 @@ bundle-push: ## Push the bundle image.
 .PHONY: bundle-publish
 bundle-publish: patch-istio-images ## Create a PR for publishing in OperatorHub
 	export GIT_USER=$(GITHUB_USER); \
-	export GITHUB_TOKEN=$(GITHUB_TOKEN); \
```
dgn, line 286: do we even need to re-export this?

PR #1208, opened 2023-06-08
OSSM-4016 Changes to better support maistra-assist

dgn:


config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/maistra-dev/istio-ubi9-operator
-  newTag: "3.0"
+  newName: registry.gitlab.com/dgrimm/istio/istio-ubi9-operator
```
jwendell: Is this right?

config/default/kustomization.yaml
```diff
@@ -1,12 +1,12 @@
...
-namespace: maistraoperator-system
+namespace: istio-operator
 
 # Value of this field is prepended to the
 # names of all resources, e.g. a deployment named
 # "wordpress" becomes "alices-wordpress".
 # Note that it should also match with the prefix (text before '-') of the namespace
 # field above.
-namePrefix: maistraoperator-
+# namePrefix: istiooperator-
```
jwendell, line 9: Intended change?

config/manager/kustomization.yaml
dgn: nope. thanks :)

config/default/kustomization.yaml
dgn: yeah, I needed to get rid of the prefix

jwendell: ok, let me put in a different way: Should we remove this field entirely (including the comments above it)? Or just commenting it is enough meaning this is a temporary change?

jwendell (approved):
minor nit, feel free to ignore

config/default/kustomization.yaml
dgn: yeah it's a good question. I decided to leave it in while we're figuring things out. We'll need a big cleanup anyway, there's a lot of commented out stuff, and a lot of boilerplate we don't need. I'll create a ticket to cover that

PR #1210, opened 2023-06-12
OSSM-4102 - Add Annotation For network-attachment-definitions

mattmahoneyrh:
Fix operator crash that was happening due to insufficient network-attachment-definitions permissions.

### DIFF
--- a/bundle/manifests/maistraoperator.clusterserviceversion.yaml
+++ b/bundle/manifests/maistraoperator.clusterserviceversion.yaml
@@ -24,7 +24,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2023-06-12T17:59:00Z"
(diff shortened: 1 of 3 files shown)


jwendell (commented):
Why do we need to read/write this? What's it being used for?

dgn:
> Why do we need to read/write this? What's it being used for?

This is for the global NetworkAttachmentDefinition that is part of the charts. The operator cannot create it unless it has this permission

dgn:
@mattmahoneyrh you should check your `git` setup, the commit author is `root`

dgn:
You'll need to run `make gen` again- it now also regenerates the OLM metadata which will include this new permission. This is something that I introduced in another PR.

dgn (approved):

mattmahoneyrh:
Suggested changes have been made.

PR #1216, opened 2023-06-15
OSSM-4185 Use a simple tag to reference new prometheus images

dgn:
Providing both a tag and digest is not supported by OSBS, causing the spec.relatedImages field to not be populated properly. Apart from that, the tag didn't even exist, it was missing the `v` prefix


FilipB:
Verified that both `registry.redhat.io/openshift4/ose-prometheus-config-reloader:v4.12` and `registry.redhat.io/openshift4/ose-prometheus:v4.12` exist and pulling is working.

jwendell (approved):

PR #1220, opened 2023-06-16
OSSM-4048 Properly manage owned resources

luksa:


controllers/istiohelminstall_controller_test.go
luksa, line 1: Note: I've changed how the operator is tested. Instead of calling the Reconcile function, the tests now fully run the controller via the controller manager. This was necessary to test whether the controller picks up changes to the owned resources. Also, this is the approach that a lot of controller developers use, as it allows you to test the actual behavior. If you call the Reconcile function directly, you miss things like the following:
- sometimes, multiple runs of the Reconcile function are needed to reconcile a resource
- a reconcile might trigger another reconcile that you don't expect (and don't catch if you test the Reconcile function directly)
- etc.

However, using this approach to testing the operator has its drawbacks. It's harder to grok what's happening in the test because there are many more moving parts. Also, when the test completes, we need to cancel the context used to run the manager. This causes some unwanted errors in the logs. 

I'm not yet 100% that we should test this way, but I'm nevertheless including this change in the PR. We'll figure it out later.

controllers/istiohelminstall_controller.go
```diff
@@ -126,25 +153,118 @@ func (r *IstioHelmInstallReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+	ownerReference := metav1.OwnerReference{
+		APIVersion:         maistrav1.GroupVersion.String(),
+		Kind:               maistrav1.IstioHelmInstallKind,
+		Name:               ihi.Name,
+		UID:                ihi.UID,
+		Controller:         pointer.Bool(true),
+		BlockOwnerDeletion: pointer.Bool(true),
+	}
+
+	err = helm.UpgradeOrInstallCharts(ctx, r.RestClientGetter, systemCharts, values, ihi.Spec.Version, ihi.Name, kube.GetOperatorNamespace(), ownerReference)
```
dgn, line 171: hmm. IIRC ownerReferences can't be cross-namespace. So for the istio-cni chart, we'll need to use a different ownerRef, e.g. one pointing to the operator deployment, right?

okay nevermind, I found your usage of annotations for this

controllers/istiohelminstall_controller.go
```diff
@@ -126,25 +153,118 @@ func (r *IstioHelmInstallReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 func (r *IstioHelmInstallReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	r.Config = mgr.GetConfig()
+	clusterScopedResourceHandler := handler.EnqueueRequestsFromMapFunc(mapOwnerAnnotationsToReconcileRequest)
+
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&v1.IstioHelmInstall{}).
+		For(&maistrav1.IstioHelmInstall{}).
+
+		// namespaced resources
+		Owns(&corev1.ConfigMap{}).
```
dgn, line 192: didn't know about `Owns()`. Good stuff

go.mod
```diff
@@ -2,17 +2,20 @@ module maistra.io/istio-operator
 
 go 1.19
 
+replace helm.sh/helm/v3 => github.com/luksa/helm/v3 v3.0.0-20230607112554-ceffc083f5a9
```
dgn, line 5: I'm not very happy about this... we haven't really gotten any feedback from the helm team whether they would accept that change. They seem to have a dev call on Thursdays, maybe we should attend this week and see if there is a chance that this will be accepted? Because if not we should be using the postRenderFn method, even though it's not as performant.

The performance impact should be minimal if we minimize the amount of times we do re-renders of the charts

hack/lint-watches.sh
```diff
@@ -0,0 +1,34 @@
...
+        for line in "${missing_kinds[@]}"; do
+            printf "  - %s\n" "$line"
+        done
+        exit 1
+    else
+        printf "Controller watches all kinds found in Helm charts.\n"
+    fi
+}
+
+check_watches
```
dgn: nit: missing newline

go.mod
```diff
@@ -135,8 +138,10 @@ require (
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
+	istio.io/api v0.0.0-20230217221049-9d422bf48675 // indirect
```
dgn, line 141: I suspect this is a remnant of the EnvoyFilter handling that is currently commented out? maybe we should remove the import while it's still unused

dgn (approved):
lgtm overall and works great, some comments

go.mod
luksa: I'll change this in a follow-up PR.

go.mod
luksa: I added the istio stuff to the schema, that's why this is here. 

Let's keep it, because I'll soon change the code to also watch istio resources.

hack/lint-watches.sh
luksa: fixed

PR #1226, opened 2023-06-20
Make sure the deploy-yaml target only outputs manifests

dgn:
In some cases we would get non-manifest output when calling `make deploy-yaml`, which in turn breaks `make deploy`. This now explicitly marks relevant targets as `.SILENT` (same as prefixing all their commands with `@` individually) and outputs user information that is unrelated to manifests to stderr, avoiding these problems.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -200,13 +200,13 @@ uninstall: kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube
 
 .PHONY: deploy
 deploy: kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -407,3 +407,5 @@ lint-bundle: operator-sdk ## runs linters against OLM metadata bundle
 
 .PHONY: lint
 lint: lint-scripts lint-go lint-yaml lint-helm lint-bundle ## runs all linters
+
+.SILENT: kustomize $(KUSTOMIZE) $(LOCALBIN) deploy-yaml
```
jwendell, line 412: cool, didn't know about it!

jwendell (approved):

PR #1231, opened 2023-06-28
OSSM-3979 Adding EnvoyExtAuthzGRPC to GA

mayleighnmyers:


jewertow (changes requested):
Please compare `ExtensionProviderEnvoyExternalAuthorizationGRPCConfig` with this [structure](https://github.com/istio/api/blob/master/mesh/v1alpha1/config.pb.go#L2521-L2547) and remove fields which don't exist there.

pkg/apis/maistra/v2/extension_providers.go
```diff
@@ -57,6 +60,32 @@ type ExtensionProviderEnvoyExternalAuthorizationHTTPConfig struct {
...
+	// The maximum duration that the proxy will wait for a response from the provider (default timeout: 600s).
+	// When this timeout condition is met, the proxy marks the communication to the authorization service as failure.
+	// In this situation, the response sent back to the client will depend on the configured `fail_open` field.
+	Timeout *string `json:"timeout,omitempty"`
+	// Sets a prefix to the value of authorization request header *Path*.
+	// For example, setting this to "/check" for an original user request at path "/admin" will cause the
+	// authorization check request to be sent to the authorization service at the path "/check/admin" instead of "/admin".	FailOpen *bool `json:"failOpen,omitempty"`
+	// Sets the HTTP status that is returned to the client when there is a network error to the authorization service.
+	// The default status is "403" (HTTP Forbidden).
+	FailOpen *bool `json:"fail_open,omitempty"`
```
jewertow: JSON schemas are camel case.
```suggestion
	FailOpen *bool `json:"failOpen,omitempty"`
```

jewertow (changes requested):

jewertow:
Could you rebase your branch?

build/patch-charts.sh
jewertow, line 1: Seems like an unrelated change...

resources/helm/v2.4/gateway/templates/poddisruptionbudget.yaml
jewertow, line 1: Same here.

resources/helm/v2.4/gateway/values.schema.json
jewertow, line 1: Same here.

resources/helm/v2.4/gateway/values.yaml
jewertow, line 1: Same here.

build/patch-charts.sh
mayleighnmyers: I mean they aren't related to Grpc, however, my `make gen` wasn't running through the test so Marko helped and we found this code to be an issue. I could take it out and put it in a new PR

resources/helm/v2.4/gateway/templates/poddisruptionbudget.yaml
mayleighnmyers: This must've came into the PR when I ran `make gen` since I don't remember doing this.

resources/helm/v2.4/gateway/values.schema.json
mayleighnmyers: This must've came into the PR when I ran make gen since I don't remember doing this.

resources/helm/v2.4/gateway/values.yaml
mayleighnmyers: This must've came into the PR when I ran make gen since I don't remember doing this.

resources/helm/v2.4/gateway/values.schema.json
luksa: This happened because some dependencies were upgrade but `make gen` wasn't run. You don't need to remove it from this PR, as it's already in #1257.  You could rebase once that PR is merged, but it's not necessary.

resources/helm/v2.4/gateway/values.yaml
luksa: This is now part of #1257. No need to remove, but you could rebase when that PR is merged.

jewertow:
@mayleighnmyers could you first rebase your branch and fix failed checks?

pkg/apis/maistra/v2/extension_providers.go
```diff
@@ -52,20 +55,44 @@ type ExtensionProviderEnvoyExternalAuthorizationHTTPConfig struct {
...
+	// REQUIRED. Specifies the port of the service.
+	Port int64 `json:"port"`
+	// The maximum duration that the proxy will wait for a response from the provider, this is the timeout for a specific request (default timeout: 600s).
+	// When this timeout condition is met, the proxy marks the communication to the authorization service as failure.
+	// In this situation, the response sent back to the client will depend on the configured `failOpen` field.
+	Timeout *string `json:"timeout,omitempty"`
+	// If true, the HTTP request or TCP connection will be allowed even if the communication with the authorization service has failed,
+	// or if the authorization service has returned a HTTP 5xx error.
+	// Default is false. For HTTP request, it will be rejected with 403 (HTTP Forbidden). For TCP connection, it will be closed immediately.
+	FailOpen bool `json:"failOpen,omitempty"`
```
mayleighnmyers: I noticed there were pointers @jewertow had used, is this fine to not use a pointer (following upstream). Or is there a reason for us to change it to a pointer?

pkg/apis/maistra/v2/extension_providers.go
```diff
@@ -52,20 +55,44 @@ type ExtensionProviderEnvoyExternalAuthorizationHTTPConfig struct {
...
+	// Sets the HTTP status that is returned to the client when there is a network error to the authorization service.
+	// The default status is "403" (HTTP Forbidden).
+	StatusOnError string `json:"statusOnError,omitempty"`
```
mayleighnmyers: Same here regarding the pointers

mayleighnmyers:
> @mayleighnmyers could you first rebase your branch and fix failed checks?

I rebased, made a few changes, and ran all tests including `make gen` which shows the only changes being the time stamp

pkg/apis/maistra/conversion/extension_providers_test.go
```diff
@@ -172,6 +172,46 @@ meshConfig:
...
+    envoyExtAuthzGrpc:
+      service: ext-authz.foo.svc.cluster.local
+      port: 8000
+      timeout: 1s
+      failOpen: true
+      statusOnError: "500"
+      includeRequestBodyInCheck:
+        maxRequestBytes: 100
+        allowPartialMessage: true
+        packAsBytes: true
```
luksa, line 215: Are these field names correct? Is it `failOpen` or `fail_open`? Both formats appear in the upstream type, but I'm not sure which one is actually used:
```
FailOpen bool `protobuf:"varint,4,opt,name=fail_open,json=failOpen,proto3" json:"fail_open,omitempty"`
                                           ^^^^^^^^^      ^^^^^^^^               ^^^^^^^^^
```

mayleighnmyers: We were following Jacek's work by doing the camel casing instead of snake casing. If we want to be more inline with upstream, I am more than happy to change! :)

luksa: Yes, I know, but I'm talking specifically about the helmValues, not about the SMCP (where, as you say, we followed Jacek's work). For helmValues, we need to use whatever istio expects.

luksa: Okay, it looks like it expects the fields to be in camel case, so we're good.

https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig-ExtensionProvider-EnvoyExternalAuthorizationGrpcProvider

jewertow: @luksa we use camel case, because we only support JSON schema.

pkg/apis/maistra/conversion/extension_providers.go
```diff
@@ -131,8 +160,46 @@ func populateExtensionProvidersConfig(in *v1.HelmValues, out *v2.ControlPlaneSpe
...
 	}
-
 	return nil
 }
+
+func convertEnvoyExtAuthzGrpc(rawExtAuthz interface{}, out *v2.ControlPlaneSpec, provider map[string]interface{}) {
+	if extAuthz, ok := rawExtAuthz.(map[string]interface{}); ok {
+		config := &v2.ExtensionProviderEnvoyExternalAuthorizationGRPCConfig{
+			Service: extAuthz["service"].(string),
+			Port:    extAuthz["port"].(int64),
```
jewertow: We should check casting result to avoid crashes.

pkg/apis/maistra/conversion/extension_providers.go
```diff
@@ -131,8 +160,46 @@ func populateExtensionProvidersConfig(in *v1.HelmValues, out *v2.ControlPlaneSpe
...
+		}
+		if rawTimeout, ok := extAuthz["timeout"]; ok {
+			config.Timeout = strPtr(rawTimeout.(string))
+		}
+		if rawFailOpen, ok := extAuthz["failOpen"]; ok {
+			config.FailOpen = rawFailOpen.(bool)
```
jewertow: Same here.

pkg/apis/maistra/conversion/extension_providers.go
luksa: Given that we need to fix these casts for ExtAuthzHTTP as well, I suggest we do it in a different PR.

pkg/apis/maistra/v2/extension_providers.go
jewertow: We use pointers in API structures to know whether a particular field was specified by a user or not. In case of value fields, you don't know whether `failOpen: false` comes from user input or is it just the default bool value.
However, we only need to know that for required fields. Optional fields don't have to be pointers.

pkg/apis/maistra/conversion/extension_providers.go
jewertow: I was also thinking about it, but if we decide for that, we cannot release this change without the follow-up change with safe casting.

pkg/apis/maistra/v2/extension_providers.go
jewertow: I was wrong, I'm sorry. We use pointers for optional values to know whether we should set them in helm values or not.

jewertow: But now I see that I didn't check default values in `validateExtensionProviders` for required fields, i.e. service and port. We can proceed this as is and improve that in a follow-up PR.

jewertow: So the following fields should be pointers: timeout, failOpen, statusOnError, includeRequestBodyInCheck.

jewertow: > to know whether we should set them in helm values or not.

Just for clarification: of course, we might use value fields and then set default values in helm templates, but then we would have to make sure that the default struct values are default API values. For example, default value for boolean field in Go is false, but an API might be implemented so that when it's not specified, then it's set to true.

pkg/apis/maistra/conversion/extension_providers.go
mayleighnmyers: Created a separate Jira to check casting https://issues.redhat.com/browse/OSSM-4628

pkg/apis/maistra/v2/extension_providers.go
mayleighnmyers: I changed failOpen and statusOnError back to pointers. includeRequestBodyInCheck was already one.

mayleighnmyers:
Marko's PR https://github.com/maistra/istio-operator/pull/1264, needs to be merged first before merging this PR.

pkg/controller/versions/strategy_v2_4.go
```diff
@@ -203,13 +203,13 @@ func (v *versionStrategyV2_4) validateExtensionProviders(spec *v2.ControlPlaneSp
...
 	}
 	for _, ext := range spec.MeshConfig.ExtensionProviders {
-		if ext.Prometheus == nil && ext.EnvoyExtAuthzHTTP == nil {
+		if ext.Prometheus == nil && ext.EnvoyExtAuthzHTTP == nil && ext.EnvoyExtAuthzGRPC == nil {
 			allErrors = append(allErrors, fmt.Errorf("extension provider '%s' does not define any provider - "+
-				"it must specify one of: prometheus or envoyExtAuthzHttp", ext.Name))
+				"it must specify one of: prometheus, envoyExtAuthzHttp, or envoyExtAuthzGrcp", ext.Name))
 		}
-		if ext.Prometheus != nil && ext.EnvoyExtAuthzHTTP != nil {
+		if ext.Prometheus != nil && ext.EnvoyExtAuthzHTTP != nil && ext.EnvoyExtAuthzGRPC == nil {
```
luksa: This if statement doesn't check what it's supposed to check. Maybe you should count the number of fields that are set and then rewrite this and the previous statement so they simply check if `count == 0` or `count > 1`.

pkg/controller/versions/strategy_v2_4.go
```diff
@@ -202,18 +202,31 @@ func (v *versionStrategyV2_4) validateExtensionProviders(spec *v2.ControlPlaneSp
...
-		if ext.Prometheus == nil && ext.EnvoyExtAuthzHTTP == nil && ext.EnvoyExtAuthzGRPC == nil {
-			allErrors = append(allErrors, fmt.Errorf("extension provider '%s' does not define any provider - "+
-				"it must specify one of: prometheus, envoyExtAuthzHttp, or envoyExtAuthzGrcp", ext.Name))
-		}
-		if ext.Prometheus != nil && ext.EnvoyExtAuthzHTTP != nil && ext.EnvoyExtAuthzGRPC == nil {
-			allErrors = append(allErrors, fmt.Errorf("extension provider '%s' must specify only one type of provider: "+
-				"prometheus, envoyExtAuthzHttp, envoyExtAuthzGrpc", ext.Name))
-		}
+
+		counter := 0
```
luksa: Better to move this to just before the if statements that do the counting (i.e. after the if statement that checks if the name is empty).

pkg/controller/versions/strategy_v2_4.go
```diff
@@ -202,18 +202,31 @@ func (v *versionStrategyV2_4) validateExtensionProviders(spec *v2.ControlPlaneSp
...
+			counter++
+		}
+		if ext.EnvoyExtAuthzGRPC != nil {
+			counter++
+		}
+		if counter == 0 {
+			allErrors = append(allErrors, fmt.Errorf("extension provider '%s' does not define any provider - "+
+				"it must specify one of: prometheus, envoyExtAuthzHttp, or envoyExtAuthzGrpc", ext.Name))
+		} else {
+			if counter > 1 {
```
luksa: This should be an `else if` instead of `else { if`.

pkg/controller/versions/strategy_v2_4.go
```diff
@@ -222,15 +235,8 @@ func (v *versionStrategyV2_4) validateExtensionProviders(spec *v2.ControlPlaneSp
...
 			}
 		}
-		if ext.EnvoyExtAuthzGRPC != nil {
-			if ext.EnvoyExtAuthzGRPC.Timeout != nil {
-				if _, err := time.ParseDuration(*ext.EnvoyExtAuthzGRPC.Timeout); err != nil {
-					allErrors = append(allErrors, fmt.Errorf("invalid extension provider '%s': envoyExtAuthzGrcp.timeout "+
-						"must be specified in the duration format - got '%s'", ext.Name, *ext.EnvoyExtAuthzGRPC.Timeout))
-				}
-			}
-		}
```
luksa: Why delete this?

pkg/apis/maistra/conversion/extension_providers.go
```diff
@@ -62,6 +62,40 @@ func populateExtensionProvidersValues(in *v2.ControlPlaneSpec, allValues map[str
...
+					includeRequestBodyInCheckValues["maxRequestBytes"] = *config.IncludeRequestBodyInCheck.MaxRequestBytes
+				}
+				if config.IncludeRequestBodyInCheck.AllowPartialMessage != nil {
+					includeRequestBodyInCheckValues["allowPartialMessage"] = *config.IncludeRequestBodyInCheck.AllowPartialMessage
+				}
+				if config.IncludeRequestBodyInCheck.PackAsBytes != nil {
+					includeRequestBodyInCheckValues["packAsBytes"] = *config.IncludeRequestBodyInCheck.PackAsBytes
+				}
+				values["includeRequestBodyInCheck"] = includeRequestBodyInCheckValues
+			}
```
luksa: This is duplicated code. Can you extract it into a function and call it for both HTTP and GRPC?

manifests-maistra/2.4.2/servicemeshcontrolplanes.crd.yaml
luksa, line 1: I believe you need to remove this file. This change will go into 2.4.3, so the only files that should be changed are the ones in the 2.4.3 directory.

manifests-servicemesh/2.4.2/servicemeshcontrolplanes.crd.yaml
luksa, line 1: Remove this file too.

pkg/apis/maistra/conversion/extension_providers.go
```diff
@@ -44,31 +44,57 @@ func populateExtensionProvidersValues(in *v2.ControlPlaneSpec, allValues map[str
...
+			if config.Timeout != nil {
+				values["timeout"] = *config.Timeout
+			}
+			if config.FailOpen != nil {
+				values["failOpen"] = *config.FailOpen
+			}
+			if config.StatusOnError != nil {
+				values["statusOnError"] = *config.StatusOnError
+			}
+
```
luksa: nit: please remove this line (and optionally add a blank line after lines 69 and 47).

luksa (approved):

jewertow (approved):

tareqmamari:
@jewertow @mayleighnmyers @luksa we have a small question, is there any plans to release this change soon?

mayleighnmyers:
> @jewertow @mayleighnmyers @luksa we have a small question, is there any plans to release this change soon?

@tareqmamari This will be release in the 2.4.3 release coming soon (a few days)

PR #1232, opened 2023-06-29
OSSM-4194 Made change to the host in the DestinationRule v2.4

mayleighnmyers:
https://issues.redhat.com/browse/OSSM-4194

- changed `"*.local"` to `".{{ .Values.global.proxy.clusterDomain }}"`
- Ran `make gen` and `make lint`

### DIFF
--- a/resources/helm/overlays/mesh-config/templates/enable-mesh-mtls.yaml
+++ b/resources/helm/overlays/mesh-config/templates/enable-mesh-mtls.yaml
@@ -27,7 +27,7 @@ metadata:
     app: istio
     release: {{ .Release.Name }}
 spec:
-  host: "*.local"
+  host: "*.{{ .Values.global.proxy.clusterDomain }}"
   {{- if .Values.global.defaultConfigVisibilitySettings }}
   exportTo:
   - '*'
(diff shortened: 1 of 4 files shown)


resources/helm/overlays/mesh-config/templates/enable-mesh-mtls.yaml
```diff
@@ -27,7 +27,7 @@ metadata:
     app: istio
     release: {{ .Release.Name }}
 spec:
-  host: "*.local"
+  host: "*.clusterDomain"
```
luksa: No. This should be something like `*.{{ .Values.global.proxy.clusterDomain }}`.

mayleighnmyers: Made the change, both make gen and make lint passed on my side!

luksa:
@mayleighnmyers did you check if there are any other charts where .local is hardcoded?

mayleighnmyers:
> @mayleighnmyers did you check if there are any other charts where .local is hardcoded?

@luksa I just changed the ones now. Do I need to change the version files that we no longer support?

luksa:
To make this change in the older versions, you need to submit separate PRs in each branch. Those changes will then be pulled into 2.4 automatically.

luksa (approved):

luksa:
@mayleighnmyers please make the same change in maistra-2.5. You'll have to do this manually, because cherry-picking won't work.

PR #1241, opened 2023-07-13
OSSM-4462 - Build Image Is Optional

mattmahoneyrh:


tests/integration/operator-integ-suite-ocp.sh
```diff
@@ -24,10 +24,12 @@ set -u
 # Print commands
 set -x
 
-echo "--------------------------------"
-echo "Build an operator image"
-echo "--------------------------------"
-"${WD}"/build-operator.sh
+if [ -v BUILD_IMAGE ]; then
```
jwendell, line 27: what does `-v BUILD_IMAGE` test?

mattmahoneyrh: If the BUILD_IMAGE environment var is set, then... build the image.

PR #1242, opened 2023-07-13
OSSM-4464 - remove image build

mattmahoneyrh:
Removing operator image build... because building the image does not belong within the scope of the test suite, but instead should be a dependency action that is done prior to running the tests.


jwendell:
why?

mattmahoneyrh:
Why should building the operator image have been included in an integration smoke test to begin with?

yxun:
Matt explained to me there is an existing operator build make target so he is trying to remove this duplicate operator build step.

yxun (approved):

mattmahoneyrh:
I am going to close this PR, because after talking through with @yxun the integration of these tests into the PROW job, there are efficiency improvements that are needed with the way the operator should be deployed and validated running.

PR #1244, opened 2023-07-17
Make the sample deployment work on OpenShift out of the box

jwendell:
These values are necessary when running on OpenShift. They are copied from upstream `openshift` profile.


leoieggli (approved):
LGTM

yxun (approved):

PR #1255, opened 2023-07-26
OSSM-4562 - Add CNI Test

mattmahoneyrh:
Add test to validate that CNI processes start successfully when the IstioHelmInstall control plane is created.


tests/integration/check-cni.sh
```diff
@@ -0,0 +1,40 @@
...
+set -e
+# Check unset variables
+set -u
+# Print commands
+set -x
+
+check-cni-ocp() { # Check that operator is running on OCP
+    local NS="${NS:-istio-operator}"
+    
+    local OPERATOR_NAME="${OPERATOR_NAME:-istio-operator}"
```
jwendell: Unsed variable

jwendell (commented):
Can we refactor these so we have all functions defined in a single file? Preferable within `operator-integ-suite-ocp.sh`. Let's try to keep it as simple as possible.

mattmahoneyrh:
Last commit included the following:
 - Simplified coded into one file
 - Improved pod 'Running' check
 - CNI pod validation

tests/integration/operator-integ-suite-ocp.sh
```diff
@@ -24,14 +24,47 @@ set -u
...
-echo "Check that istio operator is running in OCP"
-echo "--------------------------------"
-"${WD}"/check-operator.sh
-
-# create a Istio Helm Install CP and test httpbin
-echo "--------------------------------"
-echo "Create a IstioHelmInstall Control Plane and test httpbin"
-echo "--------------------------------"
-"${WD}"/control-plane-test.sh
+OPERATOR_NS="${OPERATOR_NS:-istio-operator}"
```
jwendell: We could probably reuse the variable `NAMESPACE` that's exported in the Makefile.

tests/integration/operator-integ-suite-ocp.sh
```diff
@@ -24,14 +24,47 @@ set -u
...
+OPERATOR_NAME="${OPERATOR_NAME:-istio-operator}"
+CONTROL_PLANE_NS="${CONTROL_PLANE_NS:-default}"
```
jwendell: Using `default` might hide some things. For instance, we don't need to apply the anyuid or privileged SCC's in the default namespace as it is not that strict.

Suggestion: Use anything else (e.g. istio-system).

tests/integration/operator-integ-suite-ocp.sh
```diff
@@ -24,14 +24,47 @@ set -u
...
+
+BRANCH="${BRANCH:-maistra-3.0}"
+ISTIO_HELM_INSTALL="https://raw.githubusercontent.com/maistra/istio-operator/${BRANCH}/config/samples/maistra.io_v1_istiohelminstall.yaml"
```
jwendell: Instead of pointing to a URL, just use the file from this repo. This allows the test to actually test current code, not code that's already deployed.

Suggestion: `ISTIO_HELM_INSTALL="${WD}/../../config/samples....."`

tests/integration/operator-integ-suite-ocp.sh
```diff
@@ -24,14 +24,47 @@ set -u
...
+
+TIMEOUT="3m"
+
+check_ready() {
+    local NS=$1
```
jwendell, line 37: Instead of checking for the pods *and* deployment, why not just check for deployment?

The pods might be `Running` even if they are `1/2`.

The deployment will only be `Available` (or `Ready`, don't recall right now) once all pods are running. So, just by checking the deployment should be enough.

tests/integration/operator-integ-suite-ocp.sh
```diff
@@ -24,14 +24,47 @@ set -u
...
+
+
+echo "Check that IstioHelmInstall Control Plane is running"
+check_ready "${CONTROL_PLANE_NS}" "istiod" "istiod"
+check_ready "${CONTROL_PLANE_NS}" "istio-egressgateway" "istio-egressgateway"
+check_ready "${CONTROL_PLANE_NS}" "istio-ingressgateway" "istio-ingressgateway"
+
+
+echo "Check that CNI processes are running"
+check_ready "${OPERATOR_NS}" "istio-cni" ""
```
jwendell: istio-cni uses a daemonset, not deployment.

tests/integration/operator-integ-suite-ocp.sh
jwendell: (and make sure to run the anyuid/privileged workaround)

tests/integration/operator-integ-suite-ocp.sh
mattmahoneyrh: Checking rollout status may be best way to check deployment because otherwise checking deployment Available would fail if the deployment has not yet been created at time of check. Thus, for time being checking pod and deployment Available is reliable.

tests/integration/operator-integ-suite-ocp.sh
mattmahoneyrh: Now checking daemonset rollout status.

tests/integration/operator-integ-suite-ocp.sh
mattmahoneyrh: Change made.

tests/integration/operator-integ-suite-ocp.sh
mattmahoneyrh: Change made.

tests/integration/operator-integ-suite-ocp.sh
mattmahoneyrh: Change made.

tests/integration/operator-integ-suite-ocp.sh
jwendell: My point was: There's no point in checking the running status when the check just does a grep for "Running". The command works even if the pod status is 1/2. Check the logs of the job. That's exactly what happened.

After checking this we also check for the deployment. We could simplify this script by just checking the deployment, which will have the same desired result.

But this can be done as an improvement in another PR. This one is good enough for the time being.

tests/integration/operator-integ-suite-ocp.sh
```diff
@@ -24,14 +24,49 @@ set -u
...
-# create a Istio Helm Install CP and test httpbin
-echo "--------------------------------"
-echo "Create a IstioHelmInstall Control Plane and test httpbin"
-echo "--------------------------------"
-"${WD}"/control-plane-test.sh
+NAMESPACE="${NAMESPACE:-istio-operator}"
+OPERATOR_NAME="${OPERATOR_NAME:-istio-operator}"
+CONTROL_PLANE_NS="${CONTROL_PLANE_NS:-istio-system}"
+
+BRANCH="${BRANCH:-maistra-3.0}"
```
jwendell, line 31: nit: unsed variable.

Hint: If you use vscode, install the `ShellCheck` extension. It will highlight these kind of "issues".

jwendell (approved):

PR #1264, opened 2023-08-09
OSSM-4559 Prevent panic in extensionProviders conversion

luksa:


jwendell (approved):

jewertow (commented):
It would be nice to have a test where invalid values are set in techPreview section, like this:
```
spec:
 techPreview:
   meshConfig:
     extensionProviders:
     - envoyExtAuthzHttp:
         service: "test"
         port: "80"
```
But it requires to refactor extension_provider_test.go to be executed like this: https://github.com/maistra/istio-operator/blob/a8fe5ce17d8370c44d28aff35b4b71949530f306/pkg/apis/maistra/conversion/security_test.go#L1015.

pkg/apis/maistra/conversion/extension_providers_test.go
```diff
@@ -176,3 +177,38 @@ meshConfig:
...
+	err := Convert_v1_ControlPlaneSpec_To_v2_ControlPlaneSpec(&specV1, &specV2, nil)
+	if err != nil {
+		t.Fatalf("failed to convert spec: %s", err)
+	}
+
+	errorMessage, found, err := specV2.TechPreview.GetString("errored.message")
+	if !found {
+		t.Fatalf("expected to find techPreview.errored.message field")
+	}
+	if err != nil {
```
jewertow: Shouldn't `err` be checked first? I guess this condition will never be true if executed in the current order.

pkg/apis/maistra/conversion/extension_providers_test.go
```diff
@@ -176,3 +177,38 @@ meshConfig:
...
+	}
+
+	specV1 := v1.ControlPlaneSpec{
+		Version: "v2.4",
+		Istio:   &helmValues,
+	}
+	var specV2 v2.ControlPlaneSpec
+	err := Convert_v1_ControlPlaneSpec_To_v2_ControlPlaneSpec(&specV1, &specV2, nil)
+	if err != nil {
+		t.Fatalf("failed to convert spec: %s", err)
```
jewertow: I don't understand why an error is unexpected... If `convertEnvoyExtAuthzHTTPValuesToConfig` returns error in case of failure during types casting, shouldn't an error be expected here?

pkg/apis/maistra/conversion/extension_providers_test.go
```diff
@@ -176,3 +177,38 @@ meshConfig:
...
+		t.Fatalf("failed to parse helm values: %s", err)
+	}
+
+	specV1 := v1.ControlPlaneSpec{
+		Version: "v2.4",
+		Istio:   &helmValues,
+	}
+	var specV2 v2.ControlPlaneSpec
+	err := Convert_v1_ControlPlaneSpec_To_v2_ControlPlaneSpec(&specV1, &specV2, nil)
+	if err != nil {
```
jewertow: nit: could we consistently initialize errors in `if` conditions, like in the previous one?

pkg/apis/maistra/conversion/extension_providers_test.go
luksa: Yup, you're right.

pkg/apis/maistra/conversion/extension_providers_test.go
luksa: I'm not a fan of using the `if err := doSth(); err != nil {` as it makes the code harder to scan (it hides the action in the middle of boilerplate code), so I don't think we should mandate the use of this pattern. But I changed this nevertheless.

pkg/apis/maistra/conversion/extension_providers_test.go
luksa: This [defer ](https://github.com/Maistra/istio-operator/blob/66ad6f6cac66c9fd277e239e2771b1dfce13ddcb/pkg/apis/maistra/conversion/conversion_v1_to_v2.go#L88-L121) function catches conversion errors and makes the function return nil instead of the error.

luksa: I guess I could've just called the ` convertEnvoyExtAuthzHTTPValuesToConfig` function in the unit test to make the test simpler.

jewertow: Oh... I didn't notice it. Thanks.

pkg/apis/maistra/conversion/extension_providers_test.go
jewertow: On the other hand, this approach enforces to check error and makes sure it's not overwritten then - what may cause not checking the first error. Anyway, I didn't want to mandate this approach in general. It's just a bit distracting when I see different conventions in the scope of a single function.

jewertow (approved):

pkg/apis/maistra/conversion/extension_providers_test.go
jewertow: > I guess I could've just called the  convertEnvoyExtAuthzHTTPValuesToConfig function in the unit test to make the test simpler.

Then techPreview values would not be tested.

jewertow:
I don't like current approach, because it decreases test coverage. `Convert_v1_ControlPlaneStatus_To_v2_ControlPlaneStatus` and  `Convert_v1_ControlPlaneSpec_To_v2_ControlPlaneSpec` are complex and call many functions which may affect previously set helm values.

luksa:
Yes, but the original issue was the panic in this particular function. That's the only thing that the test actually tests, so it doesn't make sense to make the test more complex than this (after all, it's a unit test and in this case, this function is the unit). 

What we really should do is add fuzz tests against the top-level conversion functions. This way, we'd test all the possible failure points.

PR #1265, opened 2023-08-10
OSSM-4639 fixing Unrecognized import path "vbom.ml/util"

mayleighnmyers:
https://issues.redhat.com/browse/OSSM-4639

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -43,7 +43,7 @@ require (
 	sigs.k8s.io/yaml v1.2.0
 )
 
-replace vbom.ml/util => github.com/fvbommel/util v0.0.2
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -43,6 +43,8 @@ require (
 	sigs.k8s.io/yaml v1.2.0
 )
 
+replace vbom.ml/util => github.com/fvbommel/util v0.0.2
```
luksa: I think we may need to reference the same commit that we referenced before instead of v0.0.2, since this appears to change some transitive dependencies.

Can you try referencing v0.0.0-20180919145318-efcd4e0f9787 instead of v0.0.2? Or the other one from go.sum. I don't know which one is correct.

mayleighnmyers: Done! :)

luksa (approved):

PR #1274, opened 2023-08-14
OSSM-4465 Changing host in DR for Jaeger to FQDN v2.4

mayleighnmyers:
https://issues.redhat.com/browse/OSSM-4465

### DIFF
--- a/resources/helm/overlays/istio-telemetry/tracing/templates/policy-jaeger.yaml
+++ b/resources/helm/overlays/istio-telemetry/tracing/templates/policy-jaeger.yaml
@@ -29,7 +29,7 @@ metadata:
     app: {{ .Values.tracing.provider }}
     release: {{ .Release.Name }}
 spec:
-  host: {{ .Values.tracing.jaeger.resourceName | default "jaeger" }}-collector.{{ .Release.Namespace }}.svc
+  host: {{ .Values.tracing.jaeger.resourceName | default "jaeger" }}-collector.{{ .Release.Namespace }}.svc.cluster.local
   trafficPolicy:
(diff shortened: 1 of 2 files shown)


jewertow (changes requested):
It will not work for custom trust domain. You can also try to just remove `.svc`.

mayleighnmyers:
> It will not work for custom trust domain. You can also try to just remove `.svc`.

@jewertow It was working for the customer to see the DR when they had it as the FQDN, which included `.svc`, plus it originally had the `.svc`, so unless we want to completely change it, that is fine too, but we need a fix for this bug right now. WDYT @luksa @rcernich

jewertow:
> but we need a fix for this bug right now

As far as it breaks some cases, I don't think so.

We need a generally correct solution, not a solution that fixes a specific case.

luksa:
We definitely shouldn't hard-code `cluster.local`. We could use `{{ .Values.global.proxy.clusterDomain }}` instead, but if @jewertow says it wouldn't work for custom trust domains, then we need a different solution.

Removing the `.svc` part would work, since the resolv.conf file in pods is configured to append `.svc.cluster.local` if omitted. But I don't know how a custom trust domain affects this. @jewertow can you explain what's the correct value when a custom trust domain is used?

jewertow:
@luksa I'm thinking about federation deployments. Let's say I created ServiceMeshPeer with `spec.security.trustDomain: west.local`. What FQDN should be set in `spec.host` in a destination rule then?

luksa (approved):

jewertow (approved):

PR #1276, opened 2023-08-15
Fix the update-istio target

jwendell:
It needs to update the make variable `ISTIO_COMMIT_30` in place, otherwise the new value will not be visible when invoking other targets, such as `gen`.

In particular, this command does not work as expected:

```sh
make update-istio gen
```

Because when the `gen` target is invoked, it is still using the old value of the `ISTIO_COMMIT_30` variable.


jwendell:
This is the reason this PR failed: https://github.com/maistra/istio-operator/pull/1275

yxun (approved):

PR #1287, opened 2023-08-29
OSSM-4758 OSSM-4759 Rename IstioHelmInstall to Istio and move to operator.istio.io/v1alpha1

luksa:


PROJECT
```diff
@@ -15,8 +15,8 @@ resources:
     crdVersion: v1
     namespaced: true
   controller: true
-  domain: maistra.io
-  kind: IstioHelmInstall
-  path: maistra.io/istio-operator/api/v1
-  version: v1
+  domain: operator.istio.io
+  kind: Istio
+  path: maistra.io/istio-operator/api/v1alpha1
```
jwendell, line 20: can we also change this path?

README.md
```diff
@@ -17,10 +17,10 @@ You’ll need a Kubernetes cluster to run against. You can use [KIND](https://si
 make deploy
 ```
 
-2. Create an IstioHelmInstall instance to install istiod:
+2. Create an instance of the Istio resource to install the Istio Control Plane:
 
 ```sh
-kubectl apply -f config/samples/maistra.io_v1_istiohelminstall.yaml
+kubectl apply -f config/samples/operator.istio.io_v1alpha1_istio.yaml
```
jwendell, line 23: not related to this pr, but... we could rename this file to something simpler, don't you think?

jwendell (approved):

PROJECT
luksa: To? I think we can only change it when we move the repo upstream.

README.md
luksa: I think we can. I can do it in a follow-up PR, where I'll also rename a few other things (e.g. `istio-editor-role` to `istio-editor`).

PR #1288, opened 2023-08-29
Remove privilege commands from test and doc

jwendell:
Plus, fix some markdown lint errors.


mattmahoneyrh (approved):
Installed operator and control plane as per instructions, all worked as expected.

PR #1295, opened 2023-09-06
Add arm64 to servicemesh manifests

yselkowitz:
This is needed for the operator to be shown in OperatorHub of the OpenShift console on ARM.  Without this, the operator can only be subscribed to ("installed") through the command line.

**DO NOT MERGE THIS** until Service Mesh on ARM is considered ready for public consumption.

### DIFF
--- a/build/generate-manifests.sh
+++ b/build/generate-manifests.sh
@@ -35,7 +35,7 @@ else
   DOCUMENTATION_URL="https://docs.openshift.com/container-platform/latest/service_mesh/v2x/servicemesh-release-notes.html"
   BUG_URL="https://issues.redhat.com/projects/OSSM"
   OLM_FEATURES="[\"Disconnected\",\"fips\"]"
-  ARCHITECTURE_LABELS=$(generateArchitectureLabels amd64 s390x ppc64le)
+  ARCHITECTURE_LABELS=$(generateArchitectureLabels amd64 s390x ppc64le arm64)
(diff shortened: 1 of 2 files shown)


build/generate-manifests.sh
```diff
@@ -35,7 +35,7 @@ else
   DOCUMENTATION_URL="https://docs.openshift.com/container-platform/latest/service_mesh/v2x/servicemesh-release-notes.html"
   BUG_URL="https://issues.redhat.com/projects/OSSM"
   OLM_FEATURES="[\"Disconnected\",\"fips\"]"
-  ARCHITECTURE_LABELS=$(generateArchitectureLabels amd64 s390x ppc64le)
+  ARCHITECTURE_LABELS=$(generateArchitectureLabels amd64 s390x ppc64le arm64)
```
jwendell, line 38: Should it be `aarch64` instead?

yselkowitz: No, arm64 is correct here.  General rule is whenever `amd64` is used instead of `x86_64`, `arm64` should be used instead of `aarch64`.

jwendell: cc @jmazzitelli

jwendell (approved):

build/generate-manifests.sh
jmazzitelli: yeah, definitely not aarch64 in the CSV. That value comes from the GOARCH env var... `arm64`.

Everything in our build infrastructure (CpaaS etc) should be `aarch64` though

PR #1298, opened 2023-09-12
OSSM-4175 Install CRDs together with the operator

luksa:


jwendell (approved):
Do you want to make the filenames `.gen` in a follow up PR?

luksa:
Yeah. I tried it, but the ones in `bundle/` end up losing the `.gen` suffix for some reason. I suggest we rename all the files in a separate PR.

PR #1299, opened 2023-09-12
OSSM-4050 type-safe helm values

bmangoen:


Makefile
```diff
@@ -23,6 +23,8 @@ VERSION ?= 3.0.0
 MINOR_VERSION := $(shell v='$(VERSION)'; echo "$${v%.*}")
 MINIMUM_SUPPORTED_VERSION ?= v3.0
 
+API_VERSION ?= v1alpha1
```
bmangoen: Should it be hardcoded here? Or should we get it dynamically from elsewhere?

PR #1302, opened 2023-09-13
Fix make lint-helm

luksa:


Makefile
```diff
@@ -421,7 +421,7 @@ lint-yaml: ## runs linters against YAML files
 .PHONY: lint-helm
 lint-helm: ## runs linters against helm charts
 	@echo "Helm version: `helm version`"
-	@${FINDFILES} -name 'Chart.yaml' -path './resources/helm/v3.?/*' \
+	@${FINDFILES} -name 'Chart.yaml' -path './resources/charts/v3.?/*' \
```
jwendell, line 424: do we need the v3 part?

luksa: only time will tell :)

jwendell: Thinking ahead, when we will support multiple versions, doesn't make sense to cover all versions?

jwendell (approved):

PR #1303, opened 2023-09-13
OSSM-4050 Type-safe helm values

bmangoen:


jwendell:
I think the tools should be in the builder image.

Check if the tools you need are already in the istio upstream builder image. We're going to switch to it soon anyway.

Makefile
```diff
@@ -351,6 +360,19 @@ envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(LOCALBIN)
 	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
+.PHONY: yq
+yq: $(YQ)
```
jwendell: no need of this. yq is present in the builder image

bmangoen: Maybe we can keep it for being able to execute `make gen` locally without installing `yq` manually into our PC. Just to add another test if `yq` is already installed and its version corresponds to the [builder image's one](https://github.com/istio/tools/blob/master/docker/build-tools/Dockerfile#L77). 
WDYT?

luksa:
@bmangoen can you add comments to patch-crd.sh? It's not clear what each function does. It's also not clear what the different sed and yq commands do. Please add comments explaining what each command does (at a high level) and why. People who look at this shell script for the first time should be able to get a basic idea of what the script does. Currently, this isn't the case - you have to study each sed/yq command in detail to figure out what's happening.

luksa:
How come there are no descriptions for the values fields?

bmangoen:
> How come there are no descriptions for the values fields?

@luksa From the [values_types.proto](https://github.com/istio/istio/blob/master/operator/pkg/apis/istio/v1alpha1/values_types.proto) file, not every value has an associated comment/description. Should I get them directly from the [charts](https://github.com/istio/istio/tree/master/manifests/charts)?

luksa:
@bmangoen We can get the descriptions from `values_types.proto`. Later we can add the missing comments from the charts and other documentation.

Makefile.core.mk
```diff
@@ -383,18 +384,26 @@ bundle-publish-nightly: OPERATOR_VERSION=$(VERSION)-nightly-$(TODAY)
...
-$(OPM): $(LOCALBIN)
-	@if test -x $(LOCALBIN)/opm && ! $(LOCALBIN)/opm version | grep -q $(OPM_VERSION); then \
-		echo "$(LOCALBIN)/opm version is not expected $(OPM_VERSION). Removing it before installing."; \
-		rm -f $(LOCALBIN)/opm; \
-	fi
-	test -s $(LOCALBIN)/opm || \
-	curl -sSLfo $(LOCALBIN)/opm https://github.com/operator-framework/operator-registry/releases/download/$(OPM_VERSION)/$(OS)-$(ARCH)-opm && \
-	chmod +x $(LOCALBIN)/opm;
+.PHONY: patch-istio-crd ## Update Istio CRD's openAPIV3Schema values
+patch-istio-crd:
```
luksa: The comment needs to be in the second line.

Makefile.core.mk
```diff
@@ -21,6 +21,7 @@
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
 VERSION ?= 3.0.0
 MINOR_VERSION := $(shell v='$(VERSION)'; echo "$${v%.*}")
+MINIMUM_SUPPORTED_VERSION ?= v3.0
```
luksa: Looks like you added this back when you had a merge conflict with one of my PRs.

hack/patch-istio-crd.sh
```diff
@@ -0,0 +1,288 @@
+#!/bin/bash
```
jwendell, line 1: I really don't like how this is written in shell. It's complex to read and fragile I guess.

:(

jwendell (commented):
Please add a minimum [integration?] test that validates the good and the bad path.

hack/patch-istio-crd.sh
bmangoen: What do you mean by complex to read and fragile?

config/samples/istio-sample-kubernetes.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
   name: istio-sample
 spec:
   version: latest
-  values:
+  rawValues:
```
luksa, line 11: Why is this change necessary? Seems wrong.

luksa:
I hate to say it, but I think we can implement this in a completely different way that's much simpler. I see that the `values_types.proto` file is already transformed into go types. Why don't we simply change the type of `Istio.spec.values` from `json.RawMessage` to the [Values](https://github.com/istio/istio/blob/master/operator/pkg/apis/istio/v1alpha1/values_types.pb.go#L4968) type from values_types.pb.go?

config/samples/istio-sample-kubernetes.yaml
luksa: Ah, I see. `Values.pilot.resources` is deprecated and you thus don't include it in the schema. So, changing `values` to `rawValues` is correct until we introduce `Istio.spec.components.pilot.k8s.resources`.

luksa (approved):
Let's get this merged. Later, we can rewrite this so that instead of adding the schema only in the CRD definition, we actually change the Go types in `IstioSpec`.

PR #1312, opened 2023-09-22
Use a string for the channels field

jwendell:
Because a value like 3.0 is interpreted as a number in annotations.yaml, and some validation tools think it differs from the value in the Dockerfile, then use quotes.


Makefile
```diff
@@ -81,7 +81,7 @@ ENVTEST_K8S_VERSION = 1.26.0
 # To re-generate a bundle for other specific channels without changing the standard setup, you can:
 # - use the CHANNELS as arg of the bundle target (e.g make bundle CHANNELS=candidate,fast,stable)
 # - use environment variables to overwrite this value (e.g export CHANNELS="candidate,fast,stable")
-CHANNELS ?= "3.0"
+CHANNELS ?= "\"3.0\""
 ifneq ($(origin CHANNELS), undefined)
 BUNDLE_CHANNELS := --channels=$(CHANNELS)
```
luksa: I think it's better to add the quotes to BUNDLE_CHANNELS so that CHANNELS can stay simple (this is important, since you can set it externally).

```suggestion
CHANNELS ?= "3.0"
ifneq ($(origin CHANNELS), undefined)
BUNDLE_CHANNELS := --channels=\"$(CHANNELS)\"
```

luksa (approved):

PR #1321, opened 2023-09-25
OSSM-3771 - Add routeConfig to additionalGateways in SMCP

eoinfennessy:
Provides the option to enable/disable creation of routes for each additionalGateway added to an SMCP.

Example SMCP spec that creates a route for the default `istio-ingressgateway` but does not create a route for the additional `test-ingress` gateway:

```yaml
spec: 
  gateways: 
    additionalIngress: 
      test-ingress: 
        enabled: true
        routeConfig: 
          enabled: false
        namespace: extra-routes
        service: 
          externalName: test-ingress
          metadata: 
            labels: 
              app: test
          ports: 
          - name: http
            port: 80
            protocol: TCP
            targetPort: 8080
          type: ClusterIP
    ingress: 
      routeConfig: 
        enabled: true
```

### DIFF
--- a/manifests-maistra/2.4.3/maistraoperator.v2.4.3.clusterserviceversion.yaml
+++ b/manifests-maistra/2.4.3/maistraoperator.v2.4.3.clusterserviceversion.yaml
@@ -15,7 +15,7 @@ metadata:
       The Maistra Operator enables you to install, configure, and manage an instance of Maistra service mesh. Maistra is based on the open source Istio project.
 
     containerImage: quay.io/maistra/istio-ubi8-operator:2.4.0
-    createdAt: 2023-09-25T16:35:59IST
+    createdAt: 2023-09-06T14:30:04EDT
     support: Red Hat, Inc. 
     olm.skipRange: ">=1.0.2 <2.4.3-0"
     operators.openshift.io/infrastructure-features: '[]'
--- a/manifests-maistra/2.4.3/maistraoperator.v2.4.3.clusterserviceversion.yaml
+++ b/manifests-maistra/2.4.3/maistraoperator.v2.4.3.clusterserviceversion.yaml
@@ -15,7 +15,7 @@ metadata:
       The Maistra Operator enables you to install, configure, and manage an instance of Maistra service mesh. Maistra is based on the open source Istio project.
 
     containerImage: quay.io/maistra/istio-ubi8-operator:2.4.0
-    createdAt: 2023-09-06T14:30:04EDT
+    createdAt: 2023-09-25T16:35:59IST
     support: Red Hat, Inc. 
     olm.skipRange: ">=1.0.2 <2.4.3-0"
     operators.openshift.io/infrastructure-features: '[]'
--- a/manifests-servicemesh/2.4.3/servicemeshoperator.v2.4.3.clusterserviceversion.yaml
+++ b/manifests-servicemesh/2.4.3/servicemeshoperator.v2.4.3.clusterserviceversion.yaml
@@ -18,7 +18,7 @@ metadata:
       The OpenShift Service Mesh Operator enables you to install, configure, and manage an instance of Red Hat OpenShift Service Mesh. OpenShift Service Mesh is based on the open source Istio project.
 
     containerImage: ${OSSM_OPERATOR_IMAGE}
-    createdAt: 2023-09-06T14:30:05EDT
+    createdAt: 2023-09-25T16:36:01IST
     support: Red Hat, Inc. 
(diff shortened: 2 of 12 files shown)


pkg/apis/maistra/v2/gateways.go
```diff
@@ -112,9 +115,6 @@ type ClusterIngressGatewayConfig struct {
 	// MeshExpansionPorts define the port set used with multi-cluster/mesh expansion
 	// +optional
 	MeshExpansionPorts []corev1.ServicePort `json:"meshExpansionPorts,omitempty"`
-	// RouteConfig specifies whether to create an OpenShift Route for istio-ingressgateway deployment
-	// +optional
-	RouteConfig *Enablement `json:"routeConfig,omitempty"`
```
jewertow, line 117: This removal should trigger removing `spec.gateways.ingress.routeConfig` from the CRD as well, so it seems that something went wrong and `make gen` ignores this change. I don't know why, but from the commit history, I see that you added `spec.gateways.additionalIngress.routeConfig` manually, instead of just modifying the Go struct and running `make gen`. I would try to remove the first commit and then run `make gen`.

jewertow (commented):
I see that the gencheck job fails, because zz_generated.deepcopy.go was not updated. It should be updated by `make gen`, but I didn't manage to setup tools properly, so I always add these changes manually accordingly to gencheck failures.

pkg/apis/maistra/v2/gateways.go
eoinfennessy: I did update the go struct and run `make gen` to update the CRDs, but somehow I failed to add `gateways.go` to the first commit.

`spec.gateways.ingress.routeConfig` remains in the CRDs because 'ClusterIngressGatewayConfig' contains (inherits?) the fields from `IngressGatewayConfig`, which is where `RouteConfig` was moved to.

Not sure why the gencheck job is failing... Thought the rebase would fix that. I'll try running `make gen` again.

luksa:
LGTM. You just need to run `make gen` to update the deepcopy code.

pkg/apis/maistra/v2/gateways.go
luksa: @eoinfennessy if you don't see any changes after running `make gen`, it's probably because you checked out the istio-operator repo somewhere other than under `GOPATH`. I believe `make gen` writes to `$GOPATH/src/github.com/maistra/istio-operator` instead of the current directory. We have to fix this.

jewertow: > spec.gateways.ingress.routeConfig remains in the CRDs because 'ClusterIngressGatewayConfig' contains (inherits?) the fields from IngressGatewayConfig, which is where RouteConfig was moved to.

Ah, right...

eoinfennessy: Thanks for your help @luksa @jewertow. `gencheck` is now succeeding. I'll squash the commits now.

luksa:
@eoinfennessy btw, there's no need to squash commits, as this happens automatically when the PR gets merged.

manifests-servicemesh/2.4.3/servicemeshoperator.v2.4.3.clusterserviceversion.yaml
```diff
@@ -18,7 +18,7 @@ metadata:
       The OpenShift Service Mesh Operator enables you to install, configure, and manage an instance of Red Hat OpenShift Service Mesh. OpenShift Service Mesh is based on the open source Istio project.
 
     containerImage: ${OSSM_OPERATOR_IMAGE}
-    createdAt: 2023-09-06T14:30:05EDT
+    createdAt: 2023-09-25T16:36:01IST
```
luksa: Please revert this change. The `make gen` command always updates this timestamp, but we usually revert this change when it's the only change in the CSV file.

manifests-maistra/2.4.3/maistraoperator.v2.4.3.clusterserviceversion.yaml
```diff
@@ -15,7 +15,7 @@ metadata:
       The Maistra Operator enables you to install, configure, and manage an instance of Maistra service mesh. Maistra is based on the open source Istio project.
 
     containerImage: quay.io/maistra/istio-ubi8-operator:2.4.0
-    createdAt: 2023-09-06T14:30:04EDT
+    createdAt: 2023-09-25T16:35:59IST
```
luksa: Please revert this one too.

luksa (approved):

jewertow (approved):
Tested, works as expected.

PR #1325, opened 2023-09-27
Fix naming

luksa:
Fixes some variable names that weren't renamed when we renamed IstioHelmInstall (ihi) to Istio and changed the package from maistra.io/v1 to operator.istio.io/v1alpha1.


bmangoen (approved):
LGTM

PR #1326, opened 2023-09-27
OSSM-4849 Add status to Istio resource

luksa:


controllers/istio_controller.go
```diff
@@ -150,15 +151,7 @@ func (r *IstioReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
-		appliedValues, err := json.Marshal(values)
-		if err != nil {
-			logger.Error(err, "failed to marshal status")
-			return
-		}
-		err = r.Client.Status().Patch(ctx, &istio, kube.NewStatusPatch(v1alpha1.IstioStatus{AppliedValues: appliedValues}))
-		if err != nil {
-			logger.Error(err, "failed to patch status")
-		}
+		err = r.updateStatus(ctx, logger, &istio, values, err)
```
jwendell: Does it make sense for `updateStatus()` to return anything? Return value is never used.

jwendell (commented):
Does it make sense to update the integration tests to rely on ready status?

controllers/istio_controller.go
luksa: Hmm, you're right. I thought it was returned, but I wasn't aware that `err` was defined within the function. You can only override the return value in a defer if the variable is defined in the function signature. See myfunc2 here: https://go.dev/play/p/QgMBrSxhWB9

I actually hate this pattern where the status is updated in a defer, so I'll rewrite this.

jwendell (approved):

PR #1330, opened 2023-10-02
OSSM-2446 Introduce profiles

luksa:


hack/download-charts.sh
```diff
@@ -1,34 +1,45 @@
...
 ISTIO_URL="${ISTIO_REPO}/archive/${ISTIO_COMMIT}.tar.gz"
 WORK_DIR=$(mktemp -d)
 EXTRACT_DIR="${ISTIO_REPO##*/}-${ISTIO_COMMIT}"
-EXTRACT_CMD="tar zxf ${ISTIO_FILE} ${EXTRACT_DIR}/manifests/charts ${EXTRACT_DIR}/manifests/addons/dashboards"
+EXTRACT_CMD="tar zxf ${ISTIO_FILE} ${EXTRACT_DIR}/manifests/charts ${EXTRACT_DIR}/manifests/profiles ${EXTRACT_DIR}/manifests/addons/dashboards"
 
-function downloadIstioCharts() {
-  rm -rf "${HELM_DIR}"
-  mkdir -p "${HELM_DIR}"
+YQ="go run github.com/mikefarah/yq/v4"
```
jwendell: YQ is already present is the istio upstream builder image. We should switch to it ASAP. @bmangoen is working on this.

hack/download-charts.sh
```diff
@@ -38,12 +49,24 @@ function patchIstioCharts() {
 - apiGroups: ["security.openshift.io"] \
   resources: ["securitycontextconstraints"] \
   resourceNames: ["privileged"] \
-  verbs: ["use"]/' "${HELM_DIR}/istio-cni/templates/clusterrole.yaml"
+  verbs: ["use"]/' "${CHARTS_DIR}/istio-cni/templates/clusterrole.yaml"
+}
+
+function convertIstioProfiles() {
```
jwendell, line 53: Do we really need to do that? I'm worried this operator becomes 2.x with lots of download/patch/sed/hack boilerplate code.

Can't we just define our own set of profiles, based on upstream?

api/v1alpha1/istio_types.go
```diff
@@ -28,6 +28,12 @@ const IstioKind = "Istio"
 // IstioSpec defines the desired state of Istio
 type IstioSpec struct {
 	Version string `json:"version,omitempty"`
+
+	// The built-in installation configuration profile to use. One of:
+	// default, demo, minimal, remote, empty, preview, ambient.
+	// When this field is left empty, the 'default' profile will be used.
+	Profile string `json:"profile,omitempty"`
```
jwendell, line 37: What about the openshift profile? Keep in mind that this is an upstream project, we can't default to openshift. Or at least we shouldn't, IMO.

hack/download-charts.sh
luksa: We do. We need to convert from `IstioOperator.install.istio.io` to `Istio.operator.istio.io` and remove the fields that we don't support (yet?), such as `components`, `meshConfig`, `hub`, and `tag`.

That's exactly what we're doing. We're defining our own set based on upstream.

api/v1alpha1/istio_types.go
luksa: We don't default to OpenShift, but we do override some values [here](https://github.com/Maistra/istio-operator/blob/fff1734ab6053c8a43b8624b386108d19dc3a14e/pkg/strategy/maistra_30_strategy.go#L20-L47). The plan is to eventually remove those overrides.

jwendell: this is wrong, no? We shouldn't be overriding values such as cni.enabled I think.

hack/download-charts.sh
jwendell: Sorry, I meant, should we just have static files in our repo instead of patching istio/istio files?

api/v1alpha1/istio_types.go
luksa: I agree. But this is how it was done initially and changing it isn't part of this PR.

hack/download-charts.sh
luksa: Why? Then we'd have to update them manually whenever they change upstream. IMHO we should support upstream profiles, not create our own.

api/v1alpha1/istio_types.go
jwendell: I got confused on why isn't openshift listed here. Plus, do we want to hardcode these values in the doc or just say that we accept any value present in istio/istio/manifests/profiles?

hack/download-charts.sh
jwendell: We are already getting only the spec.values map within `getValues()`. Can't we just use the raw upstream file as-is and ignore everything that's not `spec.values` in it?

luksa: Using the upstream file would be misleading and confusing because:
- the upstream file is an IstioOperator manifest
- ignoring fields will make things confusing and misleading; users will waste time figuring out why something isn't working

Also, this probably isn't the final state of these files. But we can't yet know what the final state will be. The way these files are transformed in this PR is what makes sense currently. We will most likely change them eventually.

hack/download-charts.sh
luksa: I'll put this PR on hold until @bmangoen's PR is merged.

Makefile
```diff
@@ -352,6 +354,19 @@ envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(LOCALBIN)
 	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
+.PHONY: yq
```
jwendell: With the changes in #1352 we don't need this. Any `make` commands will run within the build container by default, which already contains yq.

controllers/istio_controller.go
```diff
@@ -326,6 +336,79 @@ func (r *IstioReconciler) determineReadyCondition(ctx context.Context, istio *v1
 	}
 }
 
+func applyProfile(istio *v1alpha1.Istio, resourceDir string) error {
+	profileName := istio.Spec.Profile
+	if profileName == "" {
+		profileName = "default"
```
jwendell, line 341: Is it worth to make this fallback value configurable somehow? In community builds, it defaults to `default` but product builds would default to `openshift`?

controllers/istio_controller.go
```diff
@@ -326,6 +336,79 @@ func (r *IstioReconciler) determineReadyCondition(ctx context.Context, istio *v1
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func getValues(profile map[string]interface{}) (map[string]interface{}, error) {
+	val, found, err := unstructured.NestedFieldNoCopy(profile, "spec", "values")
+	if !found || err != nil {
+		return nil, err
```
jwendell, line 377: Does it play well with the `empty` profile?

controllers/istio_controller_test.go
```diff
@@ -295,3 +298,189 @@ func newCondition(conditionType v1.IstioConditionType, status bool, reason v1.Is
 		Reason: reason,
 	}
 }
+
+func TestApplyProfile(t *testing.T) {
+	resourceDir := t.TempDir()
+	profilesDir := path.Join(resourceDir, "v3.0", "profiles")
```
jwendell, line 305: perhaps out of the scope for this pr, but we should avoid hard coding `v3.0` everywhere

resources/v3.0/profiles/minimal.yaml
```diff
@@ -0,0 +1,4 @@
+# The minimal profile will install just the core control plane
+apiVersion: operator.istio.io/v1alpha1
```
jwendell, line 2: Because we got rid of everything !spec.values, this profile is identical to empty. Do we want to keep both? Does it make sense to have them at all?

jwendell (commented):
Do you think it's worth to add integration tests for each of the profiles? To make sure they are valid?

Makefile
luksa: removed

controllers/istio_controller.go
luksa: let's think about this when we clean up the openshift-specific stuff in the maistra30strategy.go file.

controllers/istio_controller.go
luksa: It should. For the empty profile, this returns a nil map. This nil map is then only used in `mergeValues()`, which can safely handle the nil value here: https://github.com/Maistra/istio-operator/blob/0d82b6ff8fca6e874e01c726efef03f791dc7124/controllers/istio_controller.go#L391

controllers/istio_controller_test.go
luksa: This "v3.0" instance is not a problem, since it's only used within the TestApplyProfile function. It could be any string, as long as the same string is used in the testcases within TestApplyProfile.

resources/v3.0/profiles/minimal.yaml
luksa: Let's keep them for now, since we'll most likely introduce the `components` field, at which point these two profiles will no longer be identical.

jwendell (approved):

jwendell:
We can think of adding integration tests for those profiles later

PR #1332, opened 2023-10-03
OSSM-4865 Patch

yxun:
Ref: https://nvd.nist.gov/vuln/detail/CVE-2021-32760
https://github.com/containerd/containerd/commit/22e9a70c71eff6507be71955947a611f2ed91e6c

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -80,3 +80,5 @@ replace bitbucket.org/ww/goautoneg => github.com/munnerz/goautoneg v0.0.0-201207
 replace github.com/operator-framework/operator-sdk => github.com/maistra/operator-sdk v0.0.0-20210824135520-3b25565223d4
 
 replace github.com/mikefarah/yaml/v2 => gopkg.in/yaml.v2 v2.4.0
+
+replace github.com/containerd/containerd => github.com/containerd/containerd v1.4.8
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -80,3 +80,5 @@ replace bitbucket.org/ww/goautoneg => github.com/munnerz/goautoneg v0.0.0-201207
 replace github.com/operator-framework/operator-sdk => github.com/maistra/operator-sdk v0.0.0-20210824135520-3b25565223d4
 
 replace github.com/mikefarah/yaml/v2 => gopkg.in/yaml.v2 v2.4.0
+
+replace github.com/containerd/containerd => github.com/containerd/containerd v1.4.8
```
jwendell: Why do you need a replace directive?

Running `go get github.com/containerd/containerd@v1.4.8`  is not enough?

yxun: Because our operator is not using containerd/containerd as a dependency directly. It was used as a dependency of a dependency in the vendor.

yxun: I think the ticket was scanning our vendor directory. I checked by running `go mod vendor` locally and then found containerd/containerd was in our vendor as version 1.3.2 and the patch was in version 1.4.8 .
Our operator go.mod does not include containerd/containerd package so it's a dependency of dependency issue.

jwendell: if you run the command above, you will see that go.mod is updated to reflect this - it adds the "indirect" keyword

yxun: that will change more dependencies in go.sum. I will follow your suggestion and push another commit and we can compare that.

jwendell: I think that using go get is cleaner than using a replaces. @luksa, any opinion here?

yxun: here are two commits:
first commit is using a "replace " in go.mod
second is reverting the first commit and then run `go get github.com/containerd/containerd@v1.4.8`

jwendell (approved):

PR #1342, opened 2023-10-05
OSSM-4881: Add support for SMCP v2.5

jewertow:


pkg/apis/maistra/v2/addons.go
```diff
@@ -16,7 +16,4 @@ type AddonsConfig struct {
 	// .Values.kiali.enabled, true if not null
 	// +optional
 	Kiali *KialiAddonConfig `json:"kiali,omitempty"`
-	// ThreeScale configures the 3scale adapter
-	// +optional
-	ThreeScale *ThreeScaleAddonConfig `json:"3scale,omitempty"`
```
luksa: We can't remove fields (we'd have to bump the API version of the resource).

jewertow: Ah, right... I will revert it.

luksa (approved):

PR #1346, opened 2023-10-07
Fix usage of oc in kind tests

jwendell:


luksa (approved):

fjglira:
integration-kind job is passing now, can we mark it again as required?

PR #1347, opened 2023-10-09
OSSM-4957 Move istio CNI plugin binaries to /var/lib/cni/bin

luksa:
OpenShift 4.14 no longer looks for CNI plugins in /opt/multus/bin.


fjglira (approved):
lgtm. This will solve all the failures that we are getting in the interop job

PR #1349, opened 2023-10-09
OSSM-4750 Bump Versions to 2.4.4 

mayleighnmyers:
https://issues.redhat.com/browse/OSSM-4750


Makefile
```diff
@@ -128,7 +128,7 @@ $(addprefix collect-templates-,$(SUPPORTED_VERSIONS)): collect-templates-%:
 	$(eval version:=$*)
 	mkdir -p ${TEMPLATES_OUT_DIR}/v${version}
 	cp ${RESOURCES_DIR}/smcp-templates/v${version}/${BUILD_TYPE} ${TEMPLATES_OUT_DIR}/v${version}/default
-	find ${RESOURCES_DIR}/smcp-templates/v${version}/ -maxdepth 1 -type f ! -name "maistra" ! -name "servicemesh" -exec cp {} ${TEMPLATES_OUT_DIR}/v${version}/ \;
+	find ${RESOURCES_DIR}/smcp-templates/v${version}/ -maxdepth 1 -type f ! -name "maistra" ! -name "servicemesh" | xargs cp -t ${TEMPLATES_OUT_DIR}/v${version}
```
luksa: Why are we changing this? I believe @fjglira recently replaced the `| xargs` with `-exec` so that the build would work on MacOS.

Makefile
```diff
@@ -18,11 +18,11 @@ THIS_FILE := $(lastword $(MAKEFILE_LIST))
 
 -include Makefile.overrides
 
-MAISTRA_VERSION        ?= 2.4.3
+MAISTRA_VERSION        ?= 2.4.4
 MAISTRA_BRANCH         ?= maistra-2.4
 ISTIO_VERSION          ?= 1.16.5
```
luksa: Does this need to be bumped as well? Haven't we updated `maistra/istio` to 1.16.7?

Makefile
fjglira: Yes, please do not change this again because will break the build on macOS. We need to keeped like this:
`find ${RESOURCES_DIR}/smcp-templates/v${version}/ -maxdepth 1 -type f ! -name "maistra" ! -name "servicemesh" -exec cp {} ${TEMPLATES_OUT_DIR}/v${version}/ \;`

resources/helm/v2.4/istio-control/istio-discovery/files/gen-istio.yaml
```diff
@@ -1810,7 +1809,6 @@ spec:
                 operator: In
                 values:
                 - "amd64"
-                - "arm64"
```
fjglira: Question: why we are deleting the arm64 from here? This will not break the support for the istiod to deploy on that arch?

Makefile
mayleighnmyers: @luksa @fjglira This was automatically changed with `make gen` so does that mean that there needs to be changes else where to ensure it is not changed again while running `make gen`?

resources/helm/v2.4/istio-control/istio-discovery/files/gen-istio.yaml
mayleighnmyers: @fjglira I will add it back in. It was automatically deleted with `make gen`

Makefile
mayleighnmyers: @luksa Hmmm. Yes we are using 1.16.7, but I dont think we updated it in this version. Would it break anything if I were to update it in this PR?

Makefile
fjglira: I do not know, I was running also make gen but from macOS. I do not know if has anything to do

jwendell (approved):

Makefile
luksa: Yeah, it looks like we forgot to set this to 1.16.7 in the maistra-2.4.2 release.

luksa (approved):

PR #1369, opened 2023-10-18
new update deps tool

jwendell:


jwendell:
Here's the output when we run it: https://gist.github.com/jwendell/5b2344f5bace7e2e053516e1e4411116

jwendell:
The plan is to create a nightly job that runs this script and creates a PR automatically.

edit: we already have a job, but it currently only update the common files: https://github.com/openshift/release/blob/fd9e1026c9c9249559d6f0ecdf5a5a0000b7d962/ci-operator/config/maistra/istio-operator/maistra-istio-operator-maistra-3.0.yaml#L71-L94

luksa (approved):

PR #1383, opened 2023-10-23
OSSM-4827 Support multiple versions

luksa:


Makefile.core.mk
```diff
@@ -264,10 +256,23 @@ gen-code: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
 
 .PHONY: gen-charts
 gen-charts: ## Pull charts from maistra/istio repository
-	hack/download-charts.sh v${MINOR_VERSION} https://github.com/${ISTIO_REPOSITORY} ${ISTIO_COMMIT_30}
+	@yq eval 'to_entries | .[] | "./hack/download-charts.sh " + .key + " " + .value.repo + " " + .value.commit + " " + ((.value.charts // []) | join(" "))' < versions.yaml | sh
```
jwendell, line 262: This is too complicated :/

Can you add comments and maybe splitting it into multiple steps to add clarity to the reader? ☺️

Makefile.core.mk
```diff
@@ -264,10 +256,23 @@ gen-code: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
...
+	@hack/copy-crds.sh "resources/$$(yq eval 'keys | .[0]' versions.yaml)/charts"
+
+.PHONY: update-version-list
```
jwendell, line 267: Generally speaking I think these recipes are become too complicated and maybe fragile.

Do you think it's better to merge all these complicated steps into a shell script? For instance, there's no need to `update-version-list` to be a target on its own. It's part of updating the charts...

Makefile.core.mk
```diff
@@ -264,10 +256,23 @@ gen-code: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
...
+		versions=$$(yq e 'keys | .[]' versions.yaml | tr '\n' ',' | sed -e 's/,/, /g' -e 's/, $$//g'); \
+		sed -i -E \
+			-e "s/(\/\/ \+operator-sdk:csv:customresourcedefinitions:type=spec,order=1,displayName=\"Istio Version\",xDescriptors=\{.*fieldGroup:General\")[^}]*(})/\1$$selectValues}/g" \
+			-e "s/(\/\/ \+kubebuilder:validation:Enum=)(.*)/\1$$versionsEnum/g" \
+			-e "s/(\/\/ \Must be one of:)(.*)/\1 $$versions./g" \
+			api/v1alpha1/istio_types.go
 
 .PHONY: gen ## Generate everything
-gen: controller-gen gen-manifests gen-code gen-charts bundle
+gen: update-version-list controller-gen gen-manifests gen-code gen-charts bundle
```
jwendell, line 280: If we are keeping this target (i.e., not merging it with `gen-charts`, I think it should be moved to after `gen-charts`.

controllers/istio_controller.go
```diff
@@ -75,13 +75,13 @@ func NewIstioReconciler(client client.Client, scheme *runtime.Scheme, restConfig
...
-	"istio-cni": "-cni",
+	"cni": "-cni",
 }
 
 // charts to deploy in the istio namespace (and their suffixes)
 var userCharts = map[string]string{
-	"base":                          "-base",
-	"istio-control/istio-discovery": "-istiod",
+	"base":   "-base",
+	"istiod": "-istiod",
```
jwendell, line 84: Does this work regardless if the charts are coming from GH or from the released zip files?

controllers/istio_controller_test.go
```diff
@@ -302,8 +304,9 @@ func newCondition(conditionType v1.IstioConditionType, status bool, reason v1.Is
 }
 
 func TestApplyProfile(t *testing.T) {
+	const version = "my-version"
```
jwendell, line 307: Are these changes from the other PR? (The one that applies the default profiles?)

versions.yaml
```diff
@@ -0,0 +1,34 @@
...
+v1.19.3:
+  repo: https://github.com/istio/istio
+  commit: 41218ffa66c12acf77ad08b3ef07d296218d0929
+  charts:
+    - https://istio-release.storage.googleapis.com/charts/base-1.19.3.tgz
+    - https://istio-release.storage.googleapis.com/charts/istiod-1.19.3.tgz
+    - https://istio-release.storage.googleapis.com/charts/gateway-1.19.3.tgz
+    - https://istio-release.storage.googleapis.com/charts/cni-1.19.3.tgz
+    - https://istio-release.storage.googleapis.com/charts/ztunnel-1.19.3.tgz
+v1.19.1:
```
jwendell, line 17: Any reason to have two 1.19 listed here?

versions.yaml
```diff
@@ -0,0 +1,34 @@
+# This file defines all the Istio versions supported by this operator.
+# The latest version must be listed first, so that the CRDs from the latest version
+# are copied to the config/crd directory.
+latest:
```
jwendell, line 4: Worth to add a `version` field to make it easier to change this format in the future?

versions.yaml
```diff
@@ -0,0 +1,34 @@
+# This file defines all the Istio versions supported by this operator.
```
jwendell, line 1: Perhaps we could have two main sections: `versions` and `aliases` where in `aliases ` we could say things like `stable` or `latest` or `stable-1.19`, etc... just thinking out loud...

jwendell (commented):
I think the third commit should be a separate PR;

Also, we could enhance the integration tests to deploy all versions in `versions.yaml` file, not only the latest.

Makefile.core.mk
luksa: this can't be split, since it merely generates a list of commands from the versions.yaml file, but I've now added comments.

Makefile.core.mk
luksa: The update-version-list target must be performed before the controller-gen target, because the former updates the comments in the go code (in istio_types.go), which are then parsed by controller-gen.

Makefile.core.mk
luksa: see my previous comment; update-version-list must be invoked before controller-gen.

controllers/istio_controller.go
luksa: Yes. The download-charts.sh script transforms the directory structure so that it matches the one in the published charts. See https://github.com/maistra/istio-operator/blob/8150bebec26d2c1b8de0209cc206679760e68dac/hack/download-charts.sh#L54-L58

controllers/istio_controller_test.go
luksa: Nope. These are more related to this PR, since we're getting rid of the v3.0 version in this PR. But instead of using "latest" here, I intentionally used a mock version name, since it doesn't actually need to be any of the actual versions.

versions.yaml
luksa: I'm going to need two patch releases of the same minor release when I implement control plane upgrades. Also, in the end, the operator will support multiple patch versions like this, so we might as well add some now.

versions.yaml
luksa: Not sure what you mean. What would that field be used for? The operator doesn't need it.

versions.yaml
luksa: Not sure why we'd need this. If the need arises, we can make these changes then. Let's try to add just the stuff that we're sure we need. Otherwise we're potentially wasting time.

jwendell (commented):
I'd love to see two things:

1) Remove the third commit. Or you can leave it but not the result of running `make update-istio`. Instead, you could put `make update-istio` directly in the `tools/update_deps.sh` script so it gets updated every night.

2) All the versions in versions.yaml being tested in integration tests. The tests we have today are very simple. They just deploy the sample resource and verify the operator and control plane are running. We could have a loop deploying different control plane versions and verifying they're successfully deployed.

versions.yaml
jwendell: Yeah, I was thinking if users could just reference `stable` in istio.spec.version and that could be translated to an actual version such as `v1.19.5`. But that would probably mean automatic upgrades...

versions.yaml
jwendell: I was just thinking about doing breaking changes in this file format in the future, and possibly having two operator versions reading the same file. Not sure it's worth the effort...

versions.yaml
jwendell: What's the plan as new versions are released? Is this list going to grow forever? Or we will keep a fixed number of releases? This file can be edited by the user for any purpose, right?

Will there be any distinction between supported and non-supported versions in this file?

luksa: > What's the plan as new versions are released? Is this list going to grow forever? Or we will keep a fixed number of releases? 

This is TBD. We can decide which versions to include in the operator every time we do a release. 

For product releases, we'll probably include the versions from the previous operator release + new patch releases for each minor version. We can drop versions from the operator just like we do in 2.x.

For the upstream operator, we can include all released Istio versions that are still supported by the community.

In any case, the decision on which versions should be included in the operator is not related to this PR. This PR merely allows us to include multiple versions in the operator.

> This file can be edited by the user for any purpose, right?

Not by the user, but by operator developers. But, yes, anyone that's building the operator can add whatever versions they want by editing this file and running `make gen`. 

> Will there be any distinction between supported and non-supported versions in this file?

There's currently no need to make a distinction. The operator allows users to deploy any version that's specified here (well, any version that's in the `resources/` dir). If needed, we can add the ability to mark supported vs unsupported versions later.

versions.yaml
luksa: This is an internal file that's only used by `make gen` to download charts, update the enum and docs in the Istio CRD, and collect all the istio CRDs. So, this file is only used by one version of the operator.

versions.yaml
luksa: For now, I think we'll even remove the `latest` version. At least from the product build, since it's not compatible with upgrades. Likewise, I don't think we'll want to add versions like `v1.19` or `stable` or anything that isn't one specific version.

We may add versions like this to the community operator, but with a warning that the upgrade behavior is undefined.

luksa:
@jwendell I removed the third commit and extended the integration test to test all versions in versions.yaml. But we need to convert the integration test from bash to go code if we want to extend it further.

jwendell (approved):

PR #1390, opened 2023-10-24
OSSM-4825 Fetch latest charts from github.com/istio/istio

luksa:


versions.yaml
```diff
@@ -2,33 +2,33 @@
...
 v1.19.3:
   repo: https://github.com/istio/istio
   commit: 41218ffa66c12acf77ad08b3ef07d296218d0929
   charts:
-  - https://istio-release.storage.googleapis.com/charts/base-1.19.3.tgz
-  - https://istio-release.storage.googleapis.com/charts/istiod-1.19.3.tgz
-  - https://istio-release.storage.googleapis.com/charts/gateway-1.19.3.tgz
-  - https://istio-release.storage.googleapis.com/charts/cni-1.19.3.tgz
-  - https://istio-release.storage.googleapis.com/charts/ztunnel-1.19.3.tgz
+    - https://istio-release.storage.googleapis.com/charts/base-1.19.3.tgz
```
jwendell, line 12: Are these spaces intentional? Asking because I made a similar mistake last week in sed :)

Makefile.core.mk
```diff
@@ -290,7 +290,7 @@ endif
 
 .PHONY: update-istio
 update-istio: ## Updates the Istio commit hash in the 'latest' entry in versions.yaml to the latest commit in the branch
-	$(eval COMMIT=$(shell yq eval '"git ls-remote " + .latest.repo + ".git | grep " + .latest.branch + " | cut -f 1"' versions.yaml | sh))
+	$(eval COMMIT=$(shell yq eval '"git ls-remote --heads " + .latest.repo + ".git " + .latest.branch + " | cut -f 1"' versions.yaml | sh))
```
jwendell, line 293: Please also add `make update-istio` to tools/update_deps.sh so we can get nightly updates for the `latest` version.

jwendell (commented):
Ideally I think it would be better if we split the commits. First one touching auxiliary files, preparing the soil. Second one would be the result of running `make update-istio gen`.

versions.yaml
luksa: yq adds these spaces when we process the file with it; yq doesn't have any option to keep the indentation as it was.

Makefile.core.mk
luksa: done

luksa:
@jwendell split into two commits

jwendell (approved):

PR #1399, opened 2023-10-26
Ensure older Istio versions can run on OpenShift

luksa:


resources/v1.18.5/charts/istiod/templates/deployment.yaml
```diff
@@ -75,7 +75,6 @@ spec:
       priorityClassName: "{{ .Values.global.priorityClassName }}"
 {{- end }}
       securityContext:
-        fsGroup: 1337
```
jwendell, line 78: securityContext becomes empty. Is this an issue at all?

hack/download-charts.sh
```diff
@@ -72,6 +72,12 @@ function patchIstioCharts() {
   resources: ["securitycontextconstraints"] \
   resourceNames: ["privileged"] \
   verbs: ["use"]/' "${CHARTS_DIR}/cni/templates/clusterrole.yaml"
+
+  # Remove fsGroup, runAsUser, runAsGroup from istiod deployment so that it can run on OpenShift
```
jwendell, line 76: Is there any risk these get reintroduced in Istio and we never notice it because of this?

resources/v1.18.5/charts/istiod/templates/deployment.yaml
luksa: Not an issue. It would be cleaner if we also removed `securityContext`, but it's hard to implement.

hack/download-charts.sh
luksa: Definitely possible, but I think we can remove this patch as soon as we stop packaging 1.19 and earlier (i.e. as soon as 1.21.0 is released, so that we have at least two different minor versions in the operator).

jwendell: ok, perhaps we should add a `#FIXME` comment so we don't forget...

jwendell (approved):

PR #1401, opened 2023-10-26
Fix lint due to the update of golangci-lint 1.55.0

bmangoen:


jwendell (approved):

luksa:
No, we have to revert this change. These ginkgo/gomega asserts are meant to be used with the dot import so that the tests read nicely. We should disable the dot import lint check instead.

jwendell:
-1 to revert.

+1 to do this change in istio/istio and then we can mimic their behavior

luksa:
What is "mimic their behavior"?

luksa:
Listen, the tests are hard to read as it is.. and with these imports, they are virtually impossible to read... if we disallow dot imports, then we should get rid of ginkgo/gomega and just use our own system... maybe something like in maistra-test-tool.

jwendell:
sorry I meant: we should follow the convention used in istio. We could propose this change there, I don't think there will be objections. Then, once they accept it and switch back to using dots, we could follow and do the same. I'm trying to avoid we deviate from them too much, especially if we have the goal of moving there soon.

luksa:
Upstream doesn't use ginkgo/gomega much. We'll use it a lot. 

Dot imports are useful, especially with DSLs. They initially wanted to remove dot imports from Go 2, but realized that it doesn't make sense to do so. This is a strong argument for *NOT* having a linter that disallows you from using dot imports.

jwendell:
I don't understand why this objection in submitting a pr in istio switching back to the dots and waiving this linter rule...

I give up. Do whatever you want.

luksa:
I don't want to waste time arguing with someone upstream to change something that doesn't make sense for them, but makes sense for us. It's clear that we need to have the ability to specify some overrides in our own project.

PR #1404, opened 2023-10-27
OSSM-5301 Support multiple default profiles

luksa:


main.go
```diff
@@ -60,12 +61,14 @@ func main() {
...
 	var configFile string
 	var resourceDirectory string
+	var defaultProfiles string
 	var logAPIRequests bool
 	var printVersion bool
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&configFile, "config-file", "/etc/istio-operator/config.properties", "Location of the config file, propagated by k8s downward APIs")
 	flag.StringVar(&resourceDirectory, "resource-directory", "/var/lib/istio-operator/resources", "Where to find resources (e.g. charts)")
+	flag.StringVar(&defaultProfiles, "default-profiles", "default", "Specifies one or more comma-separated profile names to be applied to each Istio resource")
```
jwendell: We should emphasize these profiles will *ALWAYS* be used and merged with whatever the user puts in spec.profile

main.go
```diff
@@ -122,7 +125,7 @@ func main() {
 	}
 
 	helm.ResourceDirectory = resourceDirectory
-	controller := controllers.NewIstioReconciler(mgr.GetClient(), mgr.GetScheme(), mgr.GetConfig(), resourceDirectory)
+	controller := controllers.NewIstioReconciler(mgr.GetClient(), mgr.GetScheme(), mgr.GetConfig(), resourceDirectory, strings.Split(defaultProfiles, ","))
```
jwendell, line 133: Does it play well if the user explicitly passes the empty string (i.e., they don't want any profiles to be applied automatically)? Do we have test for that?

jwendell (commented):
Do you think another alternative would be change spec.profile to spec.profiles and allow the user to pass an array? pros? cons?

main.go
luksa: I've inserted the word "always", but the description now sounds strange to me because of the "always" and "each". I think the previous version was pretty clear, since it stated that the profiles "applied to *each* Istio resource".

main.go
luksa: I've added a test and explicit handling of the empty value, even though it would be pretty hard to pass in an empty string to `--default-profiles`, since we set a default (you'd have to run the operator with `--default-profiiles=''` or `--default-profiles=`.

luksa:
> Do you think another alternative would be change spec.profile to spec.profiles and allow the user to pass an array? pros? cons?

It wouldn't fix the problem. The reason we're adding support for multiple *default* profiles is so that we can set default+openshift as the default on OpenShift. Changing `Istio.spec.profile` to `profiles` wouldn't fix the problem, as users would have to always add the "openshift" profile to their Istio resource. Also, let's keep profiles as simple as upstream instead of making them too complex like we did in 2.x. At least until there's an actual need for specifying multiple profiles in the Istio resource.

jwendell (approved):

PR #1407, opened 2023-10-27
Remove unused .golangci.yml and .yamllint.yml

luksa:


jwendell (approved):
Do you also want to remove .yamllint.yml?

PR #1409, opened 2023-10-27
OSSM-5302 Use profiles 'default' and 'openshift' as the default on OpenShift

luksa:


bundle/manifests/sailoperator.clusterserviceversion.yaml
```diff
@@ -363,6 +362,7 @@ spec:
               - args:
                 - --health-probe-bind-address=:8081
                 - --metrics-bind-address=127.0.0.1:8080
+                - --default-profiles=default,openshift
```
jwendell, line 365: I'm confused. This deployment is also used in kubernetes if OLM is used, no?

luksa: Does anyone really run OLM in k8s? Does it even work?

luksa: I mean.. does anyone run the sail operator via OLM in Kubernetes.

jwendell: Not yet. We have an issue created for that. To publish our operator in general operatorhub (not the current, ocp one).

luksa: So, that sounds the correct time to figure out how to properly configure the operator according to the underlying platform.

jwendell (approved):

PR #1410, opened 2023-10-27
OSSM-5293 Add rawValues field

bmangoen:


api/v1alpha1/istio_types.go
```diff
@@ -44,6 +44,12 @@ type IstioSpec struct {
 	// +kubebuilder:validation:Schemaless
```
bmangoen, line 48: @luksa Should we remove `+kubebuilder:pruning:PreserveUnknownFields` and `+kubebuilder:validation:Schemaless` markers for `values` field?

api/v1alpha1/istio_types.go
```diff
@@ -44,6 +44,12 @@ type IstioSpec struct {
...
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Helm Values"
 	Values json.RawMessage `json:"values,omitempty"`
+
+	// RawValues defines the values to be passed to the Helm chart when installing Istio and that do not need a CRD validation.
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Schemaless
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Helm RawValues"
+	RawValues json.RawMessage `json:"rawValues,omitempty"`
```
luksa, line 56: Should we call this `unvalidatedValues` (that's what upstream calls it). Maybe it's better than `rawValues` since it may not be clear what "raw" means in this case?

bundle/manifests/operator.istio.io_istios.yaml
```diff
@@ -55,6 +55,10 @@ spec:
                 description: The built-in installation configuration profile to use.
                   When this field is left empty, the 'default' profile will be used.
                 type: string
+              rawValues:
+                description: RawValues defines the values to be passed to the Helm
+                  chart when installing Istio and that do not need a CRD validation.
```
luksa: Maybe explain that the values here don't represent a stable API or something to that effect. And I don't think you should say "CRD validation"... Instead, just say something like "these values aren't validated".

controllers/istio_controller.go
```diff
@@ -153,6 +153,11 @@ func (r *IstioReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 
 	values := istio.Spec.GetValues()
 
+	if err := istio.Spec.SetValues(mergeValues(values, istio.Spec.GetRawValues())); err != nil {
```
luksa: You don't need to (and probably shouldn't) call `SetValues()` here. Simply merge values and rawValues and pass them to `installHelmCharts()`. 

On the other hand, you do want to store the merged values into `Istio.status.appliedValues`.

api/v1alpha1/istio_types.go
luksa: What's the effect? How does the OpenAPI schema change? How do these markers affect validation?

luksa:
Please also add a test(s). At least one test that checks that rawValues are passed to helm charts. The test should also test what happens if the same value is specified in both `values` and `rawValues`. Which of these fields takes precedence?

api/v1alpha1/istio_types.go
luksa: (but I don't really like "unvalidatedValues", because it's too long)

Maybe there's a better alternative than unvalidated or raw.

bmangoen: Maybe `unsupportedValues` but it's still too long.

luksa: Yeah. I'm thinkink maybe `rawValues` will make sense when we replace the `values` with actual types. On the other hand, outside of the operator, `values` already look like they're backed by actual types, so `rawValues` already makes sense.  

Maybe keep `rawValues` for now and then rename it later.

api/v1alpha1/istio_types.go
bmangoen: It looks like this [tag](https://book.kubebuilder.io/reference/markers/crd-validation) `+kubebuilder:validation:Schemaless` disables all type checking. So we may want to enable this check for the type-safe helm values (it may be done from this [PR](https://github.com/maistra/istio-operator/pull/1303))

api/v1alpha1/istio_types.go
```diff
@@ -68,6 +74,15 @@ func (s *IstioSpec) SetValues(values map[string]interface{}) error {
 	return nil
 }
 
+func (s *IstioSpec) GetRawValues() map[string]interface{} {
+	var rawVals map[string]interface{}
+	err := json.Unmarshal(s.RawValues, &rawVals)
+	if err != nil {
+		return nil
```
jwendell, line 81: We should print the error, or return it to the caller and handle it appropriately there. Same for `GetValues()` function above.

jwendell (commented):
Please add tests

api/v1alpha1/istio_types.go
luksa: Yes. But I don't think we need to do this right now, because I think we need to change how this field is handled altogether. We shouldn't be unmarshalling JSON at all. So when we fix this, we'll also fix the error issue. Trying to fix just the error issue would be a waste of time.

controllers/istio_controller.go
```diff
@@ -160,11 +160,13 @@ func (r *IstioReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 		return ctrl.Result{}, err
 	}
 
-	logger.Info("Installing components", "values", values)
-	err = r.installHelmCharts(ctx, istio, values)
+	mergedValues := mergeOverwrite(istio.Spec.GetRawValues(), values)
```
luksa: You need to rebase, but when you do, please make sure to use the existing `values` variable instead of introducing a new variable called `mergedValues`. Since we already make several other changes to the initial `values`, there's no need to use a new variable when we merge in the `rawValues` (this new variable might even cause bugs down the line if someone uses `values` instead of `mergedValues`).

api/v1alpha1/istio_types.go
luksa: Yes, let's remove this.

controllers/istio_controller.go
```diff
@@ -161,6 +161,8 @@ func (r *IstioReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 		return ctrl.Result{}, err
 	}
 
+	values = mergeOverwrite(istio.Spec.GetRawValues(), values)
```
luksa: I think this needs to go above `applyOverrides` so that the user can't change `global.istioNamespace` via `Istio.spec.rawValues` (we set `global.istioNamespace` in `applyOverrides` so that it always matches `Istio.metadata.namespace`).

luksa (approved):

PR #1411, opened 2023-10-27
Add gateway config instructions and sample config

longmuir:
A few notes:
- I debated only including the Gateway API config, as it does not require a new sample config. I was not sure everyone trying this would be able to enable Gateway API on their cluster though, so thought it best to also include a Gateway injection example. We will also be recommending OpenShift service mesh users adopt Gateway injection to make it easier to migrate in the future, so it's good to demonstrate.
- There may be a better way to include and reference the ingress-gateway.yaml file... I'm open to suggestions here.


bundle/README.md
```diff
@@ -79,11 +79,76 @@ By deploying the `reviews` virtual service, you can specify a different behavior
...
 The Sail Operator does not deploy Ingress or Egress Gateways. Gateways are not part of the control plane. As a security best-practice, Ingress and Egress Gateways should be deployed in a different namespace than the namespace that contains the control plane.
 
-You can deploy gateways using either the Gateway API or Gateway Injection methods. Both are well documented in the Istio documentation.
+You can deploy gateways using either the Gateway API or Gateway Injection methods. 
+
+### Option 1: Istio Gateway Injection
+
+Gateway Injection uses the same mechanisms as Istio sidecar injection to create a gateway from a `Deployment` resource that is paired with a `Service` resource that can be made accessible from outside the cluster. For more information, see [Installing Gateways](https://preliminary.istio.io/latest/docs/setup/additional-setup/gateway/#deploying-a-gateway).
+
+To configure gateway injection with the `bookinfo` application, we have provided a [sample gateway configuration](samples/ingress-gateway.yaml?raw=1) that should be applied in the namespace where the application is installed:
```
jwendell: I wonder if this file should live in the `config/samples` folder within the main directory...

bundle/README.md
```diff
@@ -79,11 +79,76 @@ By deploying the `reviews` virtual service, you can specify a different behavior
...
+
+1. Create the `istio-ingressgateway` deployment and service:
+
+    ```sh
+    $ oc apply -f ingress/gateway.yaml
```
jwendell: did you mean `ingress-gateway.yaml`?

Perhaps it's better to make the namespace explicit by specifying `-n <app-namespace>`?

bundle/README.md
```diff
@@ -79,11 +79,76 @@ By deploying the `reviews` virtual service, you can specify a different behavior
...
+Verify that the `productpage` is accessible from a web browser. 
+
+### Option 2: Kubernetes Gateway API
+
+Istio includes support for Kubernetes [Gateway API](https://gateway-api.sigs.k8s.io/) and intends to make it the default API for [traffic management in the future](https://istio.io/latest/blog/2022/gateway-api-beta/). For more information, see Istio's [Kubernetes Gateway API](https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/) page.
+
+As of Kubernetes 1.28 and OpenShift 4.14, the Kubernetes Gateway API CRDs are not available by default and must be enabled to be used. This can be done with the command:
+
+```sh
+$ oc get crd gateways.gateway.networking.k8s.io &> /dev/null ||  { oc kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.8.0" | oc apply -f -; }
```
jwendell, line 130: you're a shell hacker, nice one :)

bundle/README.md
```diff
@@ -79,11 +79,76 @@ By deploying the `reviews` virtual service, you can specify a different behavior
...
+```
+
+To configure `bookinfo` with a gateway using `Gateway API`:
+
+1. Create and configure a gateway using a `Gateway` and `HTTPRoute` resource:
+
+    ```sh
+    $ oc apply -f https://raw.githubusercontent.com/istio/istio/release-1.19/samples/bookinfo/gateway-api/bookinfo-gateway.yaml
```
jwendell: master branch?

jwendell (commented):
LGTM, just a few comments

bundle/README.md
longmuir: I didn't see that folder - I'll move it there.

bundle/README.md
longmuir: Good catch on the file name. Agreed on adding the namespace explicitely as well.

Originally I added a full example of installing Bookinfo into the`bookinfo` namespace, but decided it was better to just refer to the upstream docs, which use the Kubernetes default namespace.

bundle/README.md
longmuir: Good catch

jwendell (approved):

PR #1416, opened 2023-10-30
Wait until artifacts are available on GCS

jwendell:


luksa:
But won't this mean that this job will consume resources for an extended period of time and prevent other CI jobs from running?

jwendell:
The jobs in CI have a timeout.

luksa:
Not my point. If this job waits for 30 minutes, that's how long the CI capacity to run other jobs is affected. How long will this wait usually take?

jwendell:
Hard to say.

Note that (1) the nightly job that updates istio runs at night: https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/periodic-ci-maistra-istio-operator-maistra-3.0-update-deps; and (2) the job has to wait only when the latest commit at the time it runs is not yet available on GCS. In the worst case scenario, when this job is triggered right after a commit was merged, it has to wait the full time of the istio job that uploads the artifacts: https://prow.istio.io/job-history/gs/istio-prow/logs/release_istio_postsubmit

I don't think this is an issue at all. Istio upstream, for instance, has similar jobs (and wait times) for even post-submit tasks, which happens many times a day...

hack/update-istio.sh
```diff
@@ -2,14 +2,31 @@
 
 set -euo pipefail
 
+SLEEP_TIME=60
```
luksa, line 5: Can we reduce this to 10s or less, so that folks actually see that something's happening? 1minute seems excessive and will look like the script has hanged.

luksa (approved):
LGTM but I'd reduce the sleep time.

PR #1422, opened 2023-11-02
OSSM-1698: Don't install istio-cni on nodes with label `maistra.io/exclude-cni: 'true'`

jewertow:


luksa (approved):

jwendell:
We should submit this upstream (with `istio.io` namespace) and then cherry pick once it's approved there.

luksa:
Does upstream need this? I think the CNI node affinity can already be configured via `IstioOperator.components.cni.k8s` or something like this.

jewertow:
> We should submit this upstream (with istio.io namespace) and then cherry pick once it's approved there.

@jwendell we already did it one year ago :D

jewertow:
> then cherry pick once it's approved there

We can't cherry-pick it, because we don't use upstream istio-cni chart.

PR #1424, opened 2023-11-02
OSSM-5303 Show profiles in a drop-down list in the OpenShift console

luksa:


config/manifests/bases/sailoperator.clusterserviceversion.yaml
```diff
@@ -29,10 +29,23 @@ spec:
         - urn:alm:descriptor:com.tectonic.ui:select:v1.19.3
         - urn:alm:descriptor:com.tectonic.ui:select:v1.19.1
         - urn:alm:descriptor:com.tectonic.ui:select:v1.18.5
-      - description: The built-in installation configuration profile to use. When
-          this field is left empty, the 'default' profile will be used.
+      - description: 'The built-in installation configuration profile to use. The
+          ''default'' profile is always applied. On OpenShift, the ''openshift'' profile
+          is also applied on top of ''default''. Must be one of: ambient, default,
+          demo, empty, external, minimal, openshift, preview, remote.'
```
jwendell, line 35: Do we want to hide `openshift` and `default` from this list, since they are applied regardless?

luksa: I did that initially, but added them back after I tested the change in the console. Once you select a profile in the drop-down box, you can't deselect it, unless you refresh the form. This is why I then added back the two options.

jwendell (approved):

PR #1428, opened 2023-11-03
OSSM-5319 Clean up chart rendering

luksa:


jwendell (approved):

jwendell:
It looks like because the test doesn't deploy via OLM, the arguments for the default profiles are not passed to the operator, thus it didn't apply the openshift profile, making istio-cni pod not exist...

jwendell (approved):

PR #1434, opened 2023-11-07
Introduce HelmValues in place of map[string]interface{}

luksa:


pkg/helm/values.go
```diff
@@ -0,0 +1,21 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type HelmValues map[string]interface{}
+
+func (h HelmValues) GetString(key string) (string, bool, error) {
+	return unstructured.NestedString(h, toKeys(key)...)
+}
+
+func (h HelmValues) Set(key string, val string) error {
```
jwendell: Should this be called `SetString`?

pkg/helm/values.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"strings"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type HelmValues map[string]interface{}
+
+func (h HelmValues) GetString(key string) (string, bool, error) {
```
jwendell, line 13: Can we have comments for those exported functions?

pkg/helm/values.go
```diff
@@ -0,0 +1,21 @@
+package helm
+
+import (
+	"strings"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type HelmValues map[string]interface{}
```
jwendell: Should we have methods for other types (int, struct, etc) as well?

pkg/helm/values.go
luksa: Ah. No, `Set` is fine, but the second argument should be `val interface{}`. Thanks!

pkg/helm/values.go
luksa: Not yet. Would be dead code.

pkg/helm/values.go
luksa: Done.

pkg/helm/values.go
luksa: Fixed.

jwendell (approved):
should we replace `interface{}` with `any` in the future?

luksa:
> should we replace `interface{}` with `any` in the future?

No. Let's replace it now. :)  

Thanks!

pkg/helm/values.go
```diff
@@ -0,0 +1,25 @@
...
+
+// GetString returns the string value of a nested field.
+// Returns false if value is not found and an error if not a string.
+func (h HelmValues) GetString(key string) (string, bool, error) {
+	return unstructured.NestedString(h, toKeys(key)...)
+}
+
+// Set sets the value of a nested field to a deep copy of the value provided.
+// Returns an error if value cannot be set because one of the nesting levels is not a map[string]any.
+func (h HelmValues) Set(key string, val interface{}) error {
```
jwendell: still `interface{}`

luksa: Thanks. I missed this one.

PR #1437, opened 2023-11-08
Improve values aggregation code

luksa:
These changes make it easier to see how values are aggregated from default profiles, the selected profile, values, and rawValues.

### DIFF
--- a/controllers/istio_controller.go
+++ b/controllers/istio_controller.go
@@ -158,7 +158,7 @@ func (r *IstioReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 
 func (r *IstioReconciler) getAggregatedValues(istio v1alpha1.Istio) (helm.HelmValues, error) {
 	// 1. start with values aggregated from default profiles and the profile in Istio.spec.profile
-	values, err := r.getValuesFromProfiles(istio)
+	values, err := getValuesFromProfiles(r.getProfilesDir(istio), r.getProfiles(istio))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


controllers/istio_controller.go
```diff
@@ -171,15 +156,38 @@ func (r *IstioReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
+	// 1. start with values aggregated from default profiles and the profile in Istio.spec.profile
+	values, err := getValuesFromProfiles(r.getProfilesDir(istio), r.getProfiles(istio))
+	if err != nil {
+		return nil, err
 	}
 
+	// 2. apply values from Istio.spec.values, overwriting values from profiles
+	values = mergeOverwrite(values, istio.Spec.GetValues())
+
+	// 3. apply values from Istio.spec.rawValues, overwriting the current values
```
jwendell, line 169: so, rawValues has precedence over values... We should document this somewhere...

jwendell: BTW, do we have tests covering this?

jwendell (approved):

controllers/istio_controller.go
bmangoen: @luksa  Should it be `values` that should have precedence over `rawValues`?

luksa: Yes. in @bmangoen's initial iplementation, values had precedence over rawValues, but I've now purposefully changed this, because this is also the case in the IstioOperator resource (`unvalidatedValues` overrides `values`; see code [here](https://github.com/istio/istio/blob/57843ab90c3f1c97705166f72446d03b9773e213/operator/pkg/translate/translate.go#L563-L578)).

luksa: I've also added a test now.

jwendell (approved):

PR #1441, opened 2023-11-10
OSSM-5351 Only install istio-cni when it's enabled in values.istio_cni.enabled

luksa:


jwendell (commented):
I think you need to change the integration test to only check CNI in OCP. Maybe if it's not OCP check if CNI is *not* deployed.

luksa:
These tests don't run on OCP (or k8s). They use a fake API server.

I'll probably add with-CNI & without-CNI tests later (in a different PR) when I figure out how best to implement these tests.

jwendell:
What do you mean? The integration test runs in a OCP / Kind cluster. I think the kind test is going to fail because of this.

luksa:
This "[integration test](https://github.com/maistra/istio-operator/blob/97feba0298f6d3934cba17e9a2c4e77431a6821e/controllers/istio_controller_test.go#L36-L242)" does not run on OCP or k8s.

jwendell:
https://github.com/maistra/istio-operator/blob/041ad705a7b474ebb36cb6c5873c86fd473f9aa7/tests/integration/common-operator-integ-suite.sh#L186

luksa:
Ah, *that* integration test. :)

jwendell (approved):

PR #1443, opened 2023-11-10
Added instructions for deleting service mesh 3.0

rh-tokeefe:


bundle/README.md
```diff
@@ -255,4 +255,19 @@ To install Prometheus you will need to follow this steps:
 
         ```sh
         $ oc get route kiali -o jsonpath='{.spec.host}' -n istio-system
-        ```
\ No newline at end of file
+        ```
+## Deleting Service Mesh
```
jwendell: We have been calling it Istio in this document, not service mesh. Does it make sense to call this section "Removing Istio"?

rh-tokeefe: @jwendell I don't know what you see, but when I look, it says Istio. ;-)

Thanks for giving some thought to that while you were reviewing my work! I made the update, and squashed the PR.

jwendell (approved):

PR #1452, opened 2023-11-14
Added information for Validating the configuration

rh-tokeefe:


bundle/README.md
```diff
@@ -256,6 +256,13 @@ To install Prometheus you will need to follow this steps:
         ```sh
         $ oc get route kiali -o jsonpath='{.spec.host}' -n istio-system
         ```
+
+## Type-Safe Helm values API
```
jwendell: This title is so technical jargon that even I don't know what it means :)

Can we make it more human-friendly, e.g. `Input validation` or similar?

rh-tokeefe: @jwendell I've only heard it called Type-Safe Helm values, so I thought this was something the customer would know. Thank you for bringing this up for discussion. I'm definitely happy to change it to make it more helpful or usable for our customers. How about `Operator input validation`?

jwendell: cc @longmuir @bmangoen @luksa

longmuir: Yeah, I think there was an original intent (hope?) of type validations (ie validate if there's a string where there should be a number or vice versa)... but as I understand it, that wasn't feasible (it was a nice to have)... so we implemented a validation that would report an error if an unknown field was added - for example, if the user made a typo or tried to configure something that doesn't exist. We should have updated the related Jira issues to correct that direction when the pivot was made to better capture what was being added. Next best thing, let's try to describe it here :). Marko's demo was helpful today with his "foo: bar" example of a bogus field.

As for what to title the section... maybe "Configuration Validations", "Values Validations" or even just "Validations". I think it's also important to describe what the 'rawValues' field was in relation to the validations applied to "values".

longmuir: For what it's worth... starting to write about this in the blog, I am leaning toward "Validation of Configuration".

rh-tokeefe: I've updated the PR and changed the title to Validating the configuration, just to try and follow Red Hat's style guidelines.

bundle/README.md
```diff
@@ -256,6 +256,12 @@ To install Prometheus you will need to follow this steps:
         ```sh
         $ oc get route kiali -o jsonpath='{.spec.host}' -n istio-system
         ```
+## Validating the configuration
+
+The Operator validates the contents of the `spec.values` parameter. When parsing parameter contents, if the Operator encounters an unknown value, it generates an error. To force the Operator to pass an unknown value, use the `spec.rawValues` field.
```
jwendell: replace `parameter` with something like `field`.

`spec.values` is a field. When operator parses this field and encounters an unknown value, it will trigger an error.

bundle/README.md
```diff
@@ -256,6 +256,12 @@ To install Prometheus you will need to follow this steps:
...
+
+You can use the `oc explain` command to display the valid values for custom fields. For example, `oc explain istio.spec.values`.
```
jwendell: you can replace `custom fields` with something like `the fields in the Istio custom resource`

rh-tokeefe: @jwendell would it make sense to say custom resource definition?

Something like "...to display the valid values for the fields in the Istio custom resource definition."

jwendell: yeah, or just `CRD`, or just `Istio resource`

bundle/README.md
rh-tokeefe: Done

bundle/README.md
rh-tokeefe: Done

jwendell (approved):

PR #1453, opened 2023-11-15
Update versions.yaml with latest Istio versions

luksa:


hack/update-istio.sh
```diff
@@ -25,9 +25,9 @@ echo
 FULL_VERSION=$(curl -sSfL "${URL}")
 echo Full version: "${FULL_VERSION}"
 
-yq -i '.versions |
-    .latest.commit="'"${COMMIT}"'" |
-    .latest.charts=[
+yq -i '
```
jwendell, line 28: Thanks. The nightly job failed because of this.

jwendell (approved):

hack/update-istio.sh
luksa: Yup. I did run `make update-istio` when I made this change, but it didn't finish immediately, so I interrupted it. Then I forgot to test it again.

PR #1456, opened 2023-11-16
OSSM-5386 Add meshConfig to values schema

luksa:


bundle/manifests/operator.istio.io_istios.yaml
```diff
@@ -146,15 +146,17 @@ spec:
                           labelPods:
                             type: object
                           tag:
-                            type: string
+                            type: object
```
luksa, line 149: This is probably wrong, but is caused by the fact that `tag` is not defined as a string, but as a `google.protobuf.Value`:

https://github.com/istio/istio/blob/829b055e493511009d40e5f564c8d9ffca028917/operator/pkg/apis/istio/v1alpha1/values_types.proto#L51

bundle/manifests/operator.istio.io_istios.yaml
```diff
@@ -593,6 +599,10 @@ spec:
                       volumes:
                         type: object
                     type: object
+                  revision:
+                    type: string
+                  revisionTags:
+                    type: string
```
luksa, line 605: This should be an array. Will be fixed in a different PR.

hack/patch-istio-crd.sh
```diff
@@ -124,15 +122,15 @@ function set_fields() {
   local crd_file="${2}"
   local value_name="${3}"
 
-  set_values "${proto_file}"
-
```
luksa, line 128: Don't know why this was called. It's already called in the main function, so I don't think it needs to be called here (it gets called for each top-level value that the `set_values` function finds when it's called in the main function).

bundle/manifests/operator.istio.io_istios.yaml
bmangoen: It looks like a `google.protobuf.Value` can be any type. I don't really know which OpenAPIv3 type fits here

bmangoen (approved):

PR #1457, opened 2023-11-16
OSSM-5376 Use version v1.20.0 in the OpenShift example

luksa:


jwendell (approved):
update the k8s sample too?

luksa:
I've updated the kubernetes sample, too.

PR #1458, opened 2023-11-16
Remove old versions from resources/

luksa:


jwendell (approved):
Isn't this removed automatically when a version is removed from versions.yaml file and we run make gen?

luksa:
Not yet.

luksa:
@jwendell I've now implemented automatic removal. Please review again.

jwendell (commented):
Isn't it easier to just remove the entire parent directory and recreate them on make gen with current values from versions.yaml?

luksa:
> Isn't it easier to just remove the entire parent directory and recreate them on make gen with current values from versions.yaml?

Not sure. I don't really like the fact that we download everything every time you run `make gen`.

jwendell (approved):

PR #1459, opened 2023-11-16
OSSM-5374 Remove ingress and egress gateways from schema

luksa:


bmangoen (approved):

jwendell:
In our README there's a link to gateway chart values. Should we remove it? And maybe even mention we don't support values.gateways at all?

PR #1466, opened 2023-11-16
OSSM-5394 Fix values as array type

bmangoen:


hack/patch-istio-crd.sh
```diff
@@ -241,7 +281,7 @@ function set_nested_config_fields() {
         prefixToRemove=".spec.versions.0.schema.openAPIV3Schema.properties.spec.properties.values.properties."
         #shellcheck disable=SC2001
         echo "Changing $(echo "${nested_config_path#"$prefixToRemove"}" | sed "s/.properties//g").${name} type to ${openAPIType}"
```
luksa: Please move these log statements inside `format_type`

bmangoen: Done

luksa (approved):

PR #1473, opened 2023-11-17
Fix two different causes of infinite reconciliation loop

luksa:


controllers/istio_controller.go
```diff
@@ -285,6 +285,10 @@ func (r *IstioReconciler) updateStatus(ctx context.Context, log logr.Logger, ist
 	}
 	status.AppliedValues = appliedValues
 
+	if reflect.DeepEqual(istio.Status, *status) {
```
jwendell, line 288: worth adding a simple comment reminding readers why this is necessary?

tests/integration/common-operator-integ-suite.sh
```diff
@@ -158,6 +158,11 @@ check_ready() {
   ${COMMAND} wait deployment "${DEPLOYMENT_NAME}" -n "${NS}" --for condition=Available=True --timeout=${TIMEOUT}
 }
 
+fail() {
```
jwendell: nit: When I read a function called `fail` I expect it to fail (i.e., `exit 1`) which is not the case here. It's confusing.

tests/integration/common-operator-integ-suite.sh
```diff
@@ -172,13 +177,33 @@ main_test() {
     ${COMMAND} get ns "${CONTROL_PLANE_NS}" >/dev/null 2>&1 || ${COMMAND} create namespace "${CONTROL_PLANE_NS}"
     sed -e "s/version:.*/version: ${ver}/g" "${ISTIO_MANIFEST}" | ${COMMAND} apply -f - -n "${CONTROL_PLANE_NS}"
 
+    echo "Wait for Istio to be Reconciled"
```
jwendell, line 180: Do we need this? Or just wait for `Ready` is enough?

tests/integration/common-operator-integ-suite.sh
```diff
@@ -172,13 +177,33 @@ main_test() {
...
+    ${COMMAND} wait istio/istio-sample -n "${CONTROL_PLANE_NS}" --for condition=Reconciled=True --timeout=${TIMEOUT}
+
+    echo "Wait for Istio to be Ready"
+    ${COMMAND} wait istio/istio-sample -n "${CONTROL_PLANE_NS}" --for condition=Ready=True --timeout=${TIMEOUT}
+
+    echo "Give the operator some time to settle down"
```
jwendell: nit: s/some time/30s/

tests/integration/common-operator-integ-suite.sh
```diff
@@ -172,13 +177,33 @@ main_test() {
...
+    sleep 30
+
+    echo "Check that the operator has stopped reconciling the resource"
```
jwendell: nit: append sth like "Waiting 30s"

just so that people running this test manually knows what to expect

controllers/istio_controller.go
luksa: It's pretty clear from the code itself. Why update the status if it hasn't changed? Additionally, the commit message fully explains why this is needed. I wouldn't want to add that long explanation here.

tests/integration/common-operator-integ-suite.sh
luksa: renamed to `logFailure`

tests/integration/common-operator-integ-suite.sh
luksa: Although `Ready` should be enough, it makes sense to check `Reconciled`. Take this as an additional test assertion.

tests/integration/common-operator-integ-suite.sh
luksa: fixed

tests/integration/common-operator-integ-suite.sh
luksa: The tests waits in a few other places as well, and it doesn't say how long...   but I've made the change as suggested

jwendell (approved):

PR #1482, opened 2023-11-20
OSSM-5297: Added instructions for Installing Istioctl.

rh-tokeefe:


bundle/README.md
```diff
@@ -71,6 +71,36 @@ For a list of available configuration for the `values` field, refer to [Istio's
 - [CNI parameters](https://artifacthub.io/packages/helm/istio-official/cni?modal=values)
 - [ZTunnel parameters](https://artifacthub.io/packages/helm/istio-official/ztunnel?modal=values)
 
+## Installing the istioctl tool
+
+The istioctl tool is a configuration command line utility that allows service operators to debug and diagnose Istio service mesh deployments.
```
jwendell: surround `istioctl` with back quotes everywhere

bundle/README.md
```diff
@@ -71,6 +71,36 @@ For a list of available configuration for the `values` field, refer to [Istio's
...
+### Prerequisites
+
+Use an istioctl version that is the same version as the Istio control plane for the Service Mesh deployment. See [Istio Releases](https://github.com/istio/istio/releases) for a list of valid releases, including Beta releases.
+
+### Procedure
+
+1. Confirm if you have istioctl installed, and if so which version, by running the following command at the terminal:
+
+  ```sh
+    $ istioctl version
```
jwendell, line 87: I think we should verify the version of Istio the user is running with `oc -n istio-system get istio`

bundle/README.md
```diff
@@ -71,6 +71,36 @@ For a list of available configuration for the `values` field, refer to [Istio's
...
+    ```
+1. Upgrade istioctl by running the following command at the terminal. Replace <version> with a valid release version of Istio:
```
jwendell: `Install` instead of upgrade?

bundle/README.md
```diff
@@ -71,6 +71,36 @@ For a list of available configuration for the `values` field, refer to [Istio's
...
+1. Confirm if you have istioctl installed, and if so which version, by running the following command at the terminal:
+
+  ```sh
+    $ istioctl version
+    ```
+1. Upgrade istioctl by running the following command at the terminal. Replace <version> with a valid release version of Istio: 
+
+  ```sh
+    $ curl -sL https://istio.io/downloadIstioctl | ISTIO_VERSION=<version> sh -
+    ```
```
rh-tokeefe, line 100: @jwendell what path is being updated with the following syntax? Is it the path for istioctl or something else?

bundle/README.md
rh-tokeefe: Thank you, this was one of the questions I was going to ask

bundle/README.md
rh-tokeefe: Done, except in the title, which is the house style.

bundle/README.md
rh-tokeefe: Done

bundle/README.md
rh-tokeefe: Done

bundle/README.md
jwendell: `PATH` is a command line thing. It specifies which directories the system should look when running executable files.

For instance, you usually type `ls` instead of the full location of the `ls` binary, which is `/usr/bin/ls`. This is because `/usr/bin` is part of your `PATH`. Type `echo $PATH` to see the full list of directories. They are separated by colon.

What we do below is to add the directory of `istioctl` to the PATH so that we can invoke it by just typing `istioctl` rather than having to type `~/.istiovtl/bin/istioctl` every time.

rh-tokeefe: Thank you Professor Wendell!

rh-tokeefe: Done

jwendell (approved):

PR #1484, opened 2023-11-21
OSSM-4819 Use Helm's PostRenderer to add owner references to resources

luksa:


pkg/helm/ownerreferences.go
```diff
@@ -1,46 +1,100 @@
...
+		ownerReference: ownerReference,
+		ownerNamespace: ownerNamespace,
+	}
+}
+
+type OwnerReferencePostRenderer struct {
+	ownerReference metav1.OwnerReference
+	ownerNamespace string
+}
+
```
jwendell, line 34: nit: maybe add something like `var _ postrender.PostRenderer = &OwnerReferencePostRenderer{}` to guarantee we implement the interface?

jwendell (approved):

pkg/helm/ownerreferences.go
luksa: Not really needed, because `NewOwnerReferencePostRenderer` returns the interface - if `OwnerReferencePostRenderer` doesn't implement this interface, we get a compile-time error.

But maybe it makes sense to make it more explicit by adding a line like you suggest. Thoughts?

jwendell: Not a strong opinion, but it looks like this is good practice in go projects. I believe a compiler error on such a dummy variable declaration is easier to identify the reason than anywhere else.

luksa: Yeah, I agree it's easier to see the intention. Also, someone might change the return type of `NewOwnerReferencePostRenderer`, which would cause us to lose the interface implementation check.

Added.

PR #1487, opened 2023-11-22
OSSM-5419: Removing toolchain directive as it's breaking downstream builds

FilipB:
We can use it again when https://issues.redhat.com/browse/STONEBLD-1973 is fixed.


jwendell (commented):
I thought you were remove it in cpaas

jwendell (approved):

PR #1490, opened 2023-11-23
OSSM-5423 Fix istiod readiness check when revision is set

luksa:


jwendell (approved):
should we enhance the integration tests (in a follow up) to cover the installation and removal of a control plane with revision set?

luksa:
Yes. We should have a smoke test that sets the revision.

PR #1491, opened 2023-11-23
OSSM-5398 Add test checking spec.values validation

bmangoen:


controllers/istio_controller_test.go
```diff
@@ -78,10 +105,7 @@ var _ = Describe("IstioController", Ordered, func() {
 				},
 				Spec: v1.IstioSpec{
 					Version: istioVersion,
-					Values: []byte(`{
-						"pilot":{"image":"` + pilotImage + `"},
-						"istio_cni":{"enabled":true}
-					}`),
+					Values:  istioValues,
```
luksa, line 108: Not sure if it's a good idea to change the values of the existing test. Better to add a new test.

controllers/istio_controller_test.go
```diff
@@ -50,6 +50,33 @@ var _ = Describe("IstioController", Ordered, func() {
...
+		}
+	}`)
+
+	istioCNIValues := string(`{
+		"enabled":true,
+		"excludeNamespaces": ["openshift", "kube-system"]
+	}`)
+
+	istioValues := []byte(`{
+		"donotexist": "shouldbeomitted",
```
luksa, line 74: This value should throw an error instead of being silently ignored.

bmangoen: Actually, this value is omitted without throwing an error by the controller-runtime client but only as an info warning `INFO    KubeAPIWarningLogger    unknown field "spec.values.donotexist"`. Should we try to catch this as an error?

PR #1493, opened 2023-11-24
Reorg customize config section, add example config and version field …

longmuir:
A few things to close out the Istio operator documentation story. We should make sure to update this as the `Istio` CRD evolves.

### DIFF
--- a/bundle/README.md
+++ b/bundle/README.md
@@ -54,7 +54,11 @@ You have access to the OpenShift CLI (oc).
 
 ## Selecting the Istio version
 
-The `version` field of the `Istio` custom resource definition defines which version of Istio can be deployed. For a list of available versions, see the [versions.yaml](../versions.yaml) file. This can be set using the `Istio Version` when creating a new `Istio` with the OpenShift Container Platform web console.
+The `version` field of the `Istio` custom resource definition defines which version of Istio can be deployed. This can be set using the `Istio Version` drop down menu when creating a new `Istio` with the OpenShift Container Platform web console. For a list of available versions, see the [versions.yaml](versions.yaml) file or use the command:
+
+  ```sh
+  $ kubectl explain istio.spec.version
+  ```
 
 ## Customizing Istio configuration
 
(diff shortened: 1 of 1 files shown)


bundle/README.md
```diff
@@ -52,6 +52,10 @@ You have access to the OpenShift CLI (oc).
 
 1. When `State: Healthy` appears in the `Status` column, Istio is successfully deployed.
 
+## Selecting the Istio version
+
+The `version` field of the `Istio` custom resource definition defines which version of Istio can be deployed. For a list of available versions, see the [versions.yaml](../versions.yaml) file. This can be set using the `Istio Version` when creating a new `Istio` with the OpenShift Container Platform web console.
```
luksa: Users can also see the list of versions by running `kubectl explain istio.spec.version`.

bundle/README.md
```diff
@@ -71,6 +96,12 @@ For a list of available configuration for the `values` field, refer to [Istio's
 - [CNI parameters](https://artifacthub.io/packages/helm/istio-official/cni?modal=values)
 - [ZTunnel parameters](https://artifacthub.io/packages/helm/istio-official/ztunnel?modal=values)
 
+## Validating the configuration
+
+The Operator validates the contents of the `spec.values` field. When the Operator parses this field, and encounters an unknown value, it generates an error message. To force the Operator to pass an unknown value, use the `spec.rawValues` field.
```
luksa:
```suggestion
The Kubernetes API server validates the contents of the `spec.values` field. When it encounters an unknown value, it generates an error message. To pass an unknown value, use the `spec.rawValues` field.
```

bundle/README.md
```diff
@@ -71,6 +96,12 @@ For a list of available configuration for the `values` field, refer to [Istio's
...
+
+You can use the `oc explain` command to display the valid values for the fields in the Istio Custom Resoure Definition (CRD) file. For example, `oc explain istio.spec.values`.
```
luksa: Should we use `kubectl` instead of `oc` in all of our instructions, since we support vanilla k8s?

rh-tokeefe: @luksa based on today's team discussion, I created OSSM-5431 to change oc commands to kubectl commands.

bundle/README.md
longmuir: Thanks, I'll add that too.

luksa (approved):

PR #1504, opened 2023-12-01
Update deps

jwendell:


jwendell:
Replaces https://github.com/maistra/istio-operator/pull/1501

(difference is the 2nd commit)

luksa (approved):

PR #1508, opened 2023-12-04
OSSM-5409 Introduce IstioRevision

luksa:


controllers/istiorevision/istiorevision_controller_test.go
```diff
@@ -563,287 +545,6 @@ istio_cni:
...
 }
 
-// TestGetAggregatedValues tests that the values are sourced from the following sources
-// (with each source overriding the values from the previous sources):
-//   - default profile(s)
-//   - profile selected in Istio.spec.profile
-//   - Istio.spec.values
-//   - Istio.spec.rawValues
-//   - other (non-value) fields in the Istio resource (e.g. the value global.istioNamespace is set from Istio.metadata.namespace)
-func TestGetAggregatedValues(t *testing.T) {
```
jwendell, line 573: Is it right that this huge amount of tests got removed?

tests/integration/common-operator-integ-suite.sh
```diff
@@ -224,12 +224,9 @@ main_test() {
     echo "Undeploy Istio"
     ${COMMAND} delete -f "${ISTIO_MANIFEST}" -n "${CONTROL_PLANE_NS}"
 
-    echo "Check that Istio has been deleted"
-    if ${COMMAND} get deployment "istiod" -n "${CONTROL_PLANE_NS}" &>/dev/null; then
-      logFailure "Expected istiod deployment to have been deleted, but it still exists:"
-      ${COMMAND} -n "${CONTROL_PLANE_NS}" get deploy
-      exit 1
-    fi
+    echo "Check that istiod deployment has been deleted (waiting $TIMEOUT)"
```
jwendell, line 227: It would be nice to have this test enhanced to show the istiorevision functionality. It could start very simple, such as, creating a istio should also create a istiorevision with same spec.

jwendell (commented):
Should we mention this new resource in our README? Or it's only relevant when doing upgrades?

controllers/istiorevision/istiorevision_controller_test.go
luksa: Not removed. Moved to istio_controller_test.go.

tests/integration/common-operator-integ-suite.sh
luksa: No.

We have unit tests for that. The e2e test in this shell script should just be a smoke test. Later we can add additional tests, but these should only be proper e2e tests so that they test the actual behavior, not just parts of it. For everything else, unit tests are better, because they are simpler and execute faster.

jwendell (approved):

PR #1512, opened 2023-12-07
Remove global variable operatorNamespace

luksa:


controllers/istiorevision/suite_test.go
```diff
@@ -89,8 +88,6 @@ var _ = BeforeSuite(func() {
 
 	err = k8sClient.Create(context.TODO(), &corev1.Namespace{ObjectMeta: v1.ObjectMeta{Name: operatorNamespace}})
 	Expect(err).NotTo(HaveOccurred())
-	err = os.Setenv("POD_NAMESPACE", operatorNamespace)
```
jwendell, line 92: I'm not following... According to main.go having this variable set is mandatory, no?

main.go
```diff
@@ -84,12 +83,19 @@ func main() {
 		os.Exit(0)
 	}
 
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+
+	operatorNamespace := os.Getenv("POD_NAMESPACE")
+	if operatorNamespace == "" {
```
jwendell, line 89: Does k8s expose the namespace in the filesystem somehow (e.g. under /var/run/...)? If yes, should we fallback to it before exiting?

controllers/istiorevision/suite_test.go
luksa: The varable is only used to run the controller manager. These tests run the controller directly instead of running the manager.

Actually, this line here is exactly why I removed the global variable. If a controller needs to know the namespace, it should be passed in via a constructor argument.

main.go
luksa: Why would we need to do this? When the operator is running inside a pod, this variable is set, so there's no need for a fallback.

jwendell: not sure... just a thought...

jwendell (approved):

main.go
luksa: Hmmm, no, you actually have a point!

I'm so used to having the `POD_NAMESPACE` variable available in our operator, that I was convinced that it's set automatically by Kubernetes; But this isn't the case at all. We're the ones who set the variable in the Deployment.

So, yes, we should primarily be getting the namespace from the file-system. Kubernetes does expose the namespace in `/var/run/secrets/kubernetes.io/serviceaccount/namespace`, so we should get it from there, while still allowing folks to override it via the env var (when running the operator locally (outside of the pod)).

jwendell (approved):

PR #1517, opened 2023-12-09
OSSM-5555: Update `meshConfig.defaultConfig.tracing.zipkin.address` on SMCP reconciliation

jewertow:
I noticed that after a migration from the default Jaeger to a custom instance with different name traces disappear in Kiali and do not show up in new Jaeger instance. This is because the operator sets `meshConfig.defaultConfig.tracing.zipkin.address` only during the initial template rendering, so proxies are configured to send traces to the default Jaeger.

**How to reproduce:**
1. Create SMCP with default settings:
```
apiVersion: maistra.io/v2
kind: ServiceMeshControlPlane
metadata:
  name: basic
spec:
  tracing:
    type: Jaeger
```
2. Create custom Jaeger:
```
apiVersion: jaegertracing.io/v1
kind: Jaeger
metadata:
  name: custom-jaeger
spec:
  strategy: allInOne
  allInOne:
    options:
      query:
        base-path: /
  ingress:
    security: oauth-proxy
    openshift:
      htpasswdFile: /etc/proxy/htpasswd/auth
  volumeMounts:
  - name: secret-htpasswd
    mountPath: /etc/proxy/htpasswd
  - name: trusted-ca-bundle
    mountPath: /etc/pki/ca-trust/extracted/pem/
    readOnly: true
  volumes:
  - name: secret-htpasswd
    secret:
      secretName: htpasswd
  - name: trusted-ca-bundle
    configMap:
      defaultMode: 420
      items:
        - key: ca-bundle.crt
          path: tls-ca-bundle.pem
      name: trusted-ca-bundle
      optional: true
```
3. Configure custom Jaeger in the SMCP:
```
apiVersion: maistra.io/v2
kind: ServiceMeshControlPlane
metadata:
  name: basic
spec:
  addons:
    jaeger:
      name: custom-jaeger
  tracing:
    type: Jaeger
```

**Additional notes:**
1) Everything works fine when the custom instance is named as the managed Jaeger.
2) After updating Jaeger name, all proxies must be restarted to receive new default config.


jewertow:
After manual tests of this patch, I wonder if this is worth to merge, because I had to restart all proxies, so if users don't know abut that, they will not see traces anyway... So maybe we should not allow to update `spec.addons.jaeger.name`?

luksa:
I wouldn't disallow updates, because users would then be forced to delete and re-create the SMCP, which isn't great, since it may cause downtime.

We probably have other settings that require the proxies to be restarted, so this isn't something new. But we should document the fact that users need to restart them. cc @rh-tokeefe

jewertow:
Unfortunately, users have to recreate SMCP anyway, because SMCP controller never updates meshConfig.defaultConfig.tracing. But maybe we can ignore this problem, because Yuanlin created a PR with Zipkin extension provider, so users will be able to use that better API.

jewertow:
Closing in favor of https://github.com/maistra/istio-operator/pull/1531.

PR #1519, opened 2023-12-10
OSSM-3368: Allow changing SMCP mode to cluster-wide

jewertow:


luksa (approved):
LGTM, but we need tests for this (for both cases: changing from MultiTenant to ClusterWide and vice-versa). I assume you plan to add these tests to maistra-test-tool.

jewertow:
I tested these cases manually. I'm going to add proper tests to maistra-test-tool soon.

PR #1521, opened 2023-12-11
OSSM-5425 Allow two control planes to be deployed concurrently

luksa:


controllers/istiorevision/istiorevision_controller.go
```diff
@@ -154,25 +161,33 @@ func (r *IstioRevisionReconciler) installHelmCharts(ctx context.Context, rev *v1
...
 		BlockOwnerDeletion: ptr.Of(true),
 	}
+	ownerNamespace := rev.Namespace
 
 	if cniEnabled, err := isCNIEnabled(values); err != nil {
 		return err
 	} else if cniEnabled {
-		if err := helm.UpgradeOrInstallCharts(ctx, r.RestClientGetter, []string{"cni"}, values,
-			rev.Spec.Version, rev.Name, rev.Namespace, ownerReference); err != nil {
+		if shouldInstallCNI, err := r.isOldestRevisionWithCNI(ctx, rev); shouldInstallCNI {
```
jwendell, line 169: Why are we picking the oldest timestamp creation as the winner? Isn't it better to pick the newest CP version among all CP's? In theory newer Istios should keep some sort of backward compatibility?

jwendell (commented):
If I understand it correctly, we are deploying just one cni installation in the operator namespace, that's shared by all control plane installations in the cluster, right?

Is this a temporary solution while we don't have multiple CNI's like 2.x? I'm worried about compatibility between different CP versions (eg 1.20 and 1.21) sharing the same CNI.

controllers/istiorevision/istiorevision_controller.go
luksa: Trust me, I thought about both of these options and more. The reason I take the oldest CP is so that there's no way for the existing control plane to fail when the user starts an upgrade. If I took the newest CP, then CNI would get upgraded immediately, even before the user moves any workloads to the new control plane. If the new CNI plugin is broken, that means they immediately lose the ability to inject the sidecar.

And yes, this is a temporary solution until each control plane can use its own CNI plugin.

In the past, in OSSM 2.x, we actually forgot to upgrade CNI a few times. At one point, I think we actually had CNI at CPversion - 2.

jwendell (approved):

PR #1522, opened 2023-12-11
OSSM-5416 Building multi-platform container image

bmangoen:


jwendell:
@fjglira can you try this patch in your environment? Make sure you can build, push && pull for your platform (arm?)

fjglira:
@jwendell checking

Makefile.core.mk
```diff
@@ -175,20 +174,47 @@ docker-push-nightly: docker-build
...
+BUILDX_TAGS += --tag $(HUB)/$(IMAGE_BASE):$(MINOR_VERSION)-nightly-$(TODAY)
+endif
+
 # PLATFORMS defines the target platforms for  the manager image be build to provide support to multiple
 # architectures. (i.e. make docker-buildx IMAGE=myregistry/mypoperator:0.0.1). To use this option you need to:
 # - able to use docker buildx . More info: https://docs.docker.com/build/buildx/
 # - have enable BuildKit, More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 # - be able to push the image for your registry (i.e. if you do not inform a valid value via IMAGE=<myregistry/image:<tag>> then the export will fail)
 # To properly provided solutions that supports more than one platform you should use this option.
 PLATFORMS ?= linux/arm64,linux/amd64,linux/s390x,linux/ppc64le
```
bmangoen, line 200: @jwendell By default, it builds for these four architectures (arm64, amd64, s390x and ppc64le). I know that the first target platforms are `linux/amd64` and `linux/arm64`. Do you think we leave it at is or we should remove `linux/s390x` and `linux/ppc64le` for now?

jwendell: "by default" you mean if we run `make docker-buildx` or `make build-all`, right?

if I run `make build` it will build only linux/am64, correct?

Makefile.core.mk
```diff
@@ -175,20 +174,47 @@ docker-push-nightly: docker-build
 	docker tag ${IMAGE} $(HUB)/$(IMAGE_BASE):$(MINOR_VERSION)-latest
 	docker push $(HUB)/$(IMAGE_BASE):$(MINOR_VERSION)-latest
 
+# NIGHTLY defines if the nightly image should be pushed or not
+NIGHTLY ?= false
```
jwendell, line 178: Why do we need this? Can't we rely on existing `*-nightly` targets?

Makefile.core.mk
bmangoen: Correct:
- `make build` builds the manager binary for `linux/amd64` platform only
- `make build-all` builds the manager binaries for the listed platforms
- `make docker-buildx` builds and pushes the multi-platform container image

Makefile.core.mk
bmangoen: `docker buildx build` has the `--tag` option which it can be used multiple times to build/push multiple image tags. We may get rid of the `*-nightly` targets

fjglira:
Hey, after latest changes it's working on a cluster bot arm cluster but it was not possible to run it locally on my macOS, this is because TARGETOS is being set to the local OS which is darwin. I still need to do some small changes to be able to set the test job to be able to run against arm cluster. But I'll do it when this PR is merged

yxun:
Hi, does that multi arch build work in CI ? 
I thought the docker buildx is based on the OS arch. So in theory, our CI node has only x86(amd64) arch and then we can only build amd64 images in CI.

fjglira:
> Hi, does that multi arch build work in CI ? I thought the docker buildx is based on the OS arch. So in theory, our CI node has only x86(amd64) arch and then we can only build amd64 images in CI.

I'm changing the integration test to docker-buildx and right now it's failing: https://github.com/maistra/istio-operator/pull/1528. I'm checking the logs

fjglira:
> > Hi, does that multi arch build work in CI ? I thought the docker buildx is based on the OS arch. So in theory, our CI node has only x86(amd64) arch and then we can only build amd64 images in CI.
> 
> I'm changing the integration test to docker-buildx and right now it's failing: #1528. I'm checking the logs

Related this, it can't be tested until this PR is merged. I tested yesterday locally against an arm cluster an was working the test but on my local machine was not possible because of the TARGETOS as I mentioned before

bmangoen:
> Hi, does that multi arch build work in CI ? I thought the docker buildx is based on the OS arch. So in theory, our CI node has only x86(amd64) arch and then we can only build amd64 images in CI.

@yxun the build is not based on the host OS arch. The process is first to build the manager binaries for all architectures (with `go build ...`) and then `docker buildx` creates the container image which includes a docker manifest for each architecture. For example, the `3.0-latest` tag of this [image here](https://quay.io/repository/bmangoen/istio-operator?tab=tags) has 4 manifests (`linux on arm64`, `linux on amd64`, `linux on s390x` and `linux on ppc64le`) so it could run on these specific arch.

fjglira:
After some fights with my local configuration, I was able to run: `make BUILD_WITH_CONTAINER=1 HUB=quay.io/frherrer docker-buildx` with success build and pushing for all the arch the operator image:
```
 => => pushing layers                                                                                                                                                                                                              15.6s
 => => pushing manifest for quay.io/frherrer/istio-operator:3.0-latest@sha256:794cfb5b2330295be1295c794f0a99691a341834cbd15796c0d30a33e8aa0a0c                                                                                      5.6s
 => [auth] frherrer/istio-operator:pull,push token for quay.io
```
pushing the images for all these configurations: `--platform=linux/arm64,linux/amd64,linux/s390x,linux/ppc64le`

I tested by running the image using the integration test and the operator install without any error in the ARM cluster

fjglira (approved):
lgtm

PR #1529, opened 2023-12-13
Add gwControllerMode fork

jwendell:


bundle/manifests/sailoperator.clusterserviceversion.yaml
```diff
@@ -17,8 +17,8 @@ metadata:
       ]
     capabilities: Seamless Upgrades
     categories: OpenShift Optional, Integration & Delivery, Networking, Security
-    containerImage: quay.io/maistra-dev/istio-operator:3.0-latest
-    createdAt: "2023-12-13T03:14:21Z"
+    containerImage: quay.io/jwendell/istio-operator:gwAPIControllerMode
```
luksa, line 20: We need to fix the make targets so they don't make these changes in the source files.

PR #1534, opened 2023-12-14
OSSM-5414 Make Istio and IstioRevision cluster-scoped

luksa:


config/samples/istio-sample-kubernetes.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: istio-sample
 spec:
   version: v1.20.0
+  namespace: istio-system
```
luksa, line 7: Should this field default to `istio-system`? In production systems, customers will want to use a custom namespace, but for first-time dev users, it might make sense to default to `istio-system`.

jwendell (approved):

config/samples/istio-sample-kubernetes.yaml
jwendell: yes, this is just a sample. All istio docs refer to istio-system in their samples. We are good.

luksa: No, I mean in the CRD itself. Currently, the `namespace` field is required. But we could make it optional and default it to `istio-system`.

jwendell: Please no. Let users explicitly set it.

luksa: Well, they don't have to specify it when they use `istioctl install`. That's why I think it makes sense to default to `istio-system`.

jwendell: Just because upstream does something it doesn't mean we have to blindly do it as well. We have to chance to "start over" with best practices from the beginning. IMO it's better to be explicit than implicit for sensitive fields like this. Specially if we plan to support any sort of multi-tenancy since the beginning. Just my $0.02.

luksa: I'm not doing it blindly. I explain above why defaulting to `istio-system` would be a good idea.

PR #1535, opened 2023-12-14
Add v1.20.1 and v1.19.5, remove 1.19.4

luksa:


jwendell (approved):
Do we want to keep 1.20.0 as this stage? We're not committed yet to support customers, so, I think for now we could keep just the latest version for each stream...

I think you can also bump 1.19.5.

luksa:
I'd like to keep 1.20.0 so I can experiment with both minor and patch version upgrades.

PR #1536, opened 2023-12-14
Various fixes

luksa:
- The make gen command did not update the IstioRevision resource information in the CSV
- The IstioRevision resource was missing all the OCP console metadata in the CSV
- Some places referred to the "IstioRevision version" instead of "Istio version"
- The Version dropdown options weren't being updated for IstioRevision
- Improved a few field descriptions

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -19,4 +19,12 @@ resources:
   kind: Istio
   path: maistra.io/istio-operator/api/v1alpha1
   version: v1alpha1
+- api:
+    crdVersion: v1
+    namespaced: true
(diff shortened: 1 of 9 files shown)


bmangoen:
> * The make gen command did not update the IstioRevision resource information in the CSV

@luksa Should we have to fix this?

luksa:
Why wouldn't we need to fix this? When you made changes to any description in IstioRevision, the CSV wasn't updated. Now it is.

bmangoen (approved):

PR #1538, opened 2023-12-15
Improve Istio and IstioRevision descriptions

luksa:


api/v1alpha1/istio_types.go
```diff
@@ -200,7 +200,13 @@ const (
 // +kubebuilder:printcolumn:name="Version",type="string",JSONPath=".spec.version",description="The version of the control plane installation."
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="The age of the object"
 
-// Istio represents an Istio Service Mesh deployment
+// Istio represents an Istio Service Mesh deployment consisting of one or more
+// control plane instances (represented by one or more IstioRevision objects).
+// To deploy an Istio Service Mesh, a cluster-admin creates an Istio object with
+// the desired Istio version and configuration. The Istio operator then creates
+// an IstioRevision object, which in turn creates the underlying Deployment
+// objects for istiod and other control plane components, not unlike how a
```
jwendell: "not unlike" sounds confusing to me. Perhaps "similar to?"

jwendell (approved):

api/v1alpha1/istio_types.go
luksa: Yes, confusing to us non-native speakers, but I think native speakers usually use this form instead (even though it is a double negative). Let's wait for @rh-tokeefe to chime in.

rh-tokeefe: @jwendell and @luksa in a way, both of you are correct. The phrase "not unlike" is correct, but it isn't all that common. I think  Jonh's suggestion works well. "...control plane components, similar to how a Deployment object in Kubernetes creates ReplicaSets that create the Pods"

api/v1alpha1/istio_types.go
```diff
@@ -200,7 +200,13 @@ const (
 // +kubebuilder:printcolumn:name="Version",type="string",JSONPath=".spec.version",description="The version of the control plane installation."
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="The age of the object"
 
-// Istio represents an Istio Service Mesh deployment
+// Istio represents an Istio Service Mesh deployment consisting of one or more
```
rh-tokeefe, line 203: @luksa just a thought...when we say "a cluster-admin creates" should we say "a user account with the cluster-admin role"

luksa: Hm, my intent was not to specifically say that a cluster-admin role is needed, only that a cluster administrator creates the object. I started just with "user", but then realized I could make it more specific, but it's not completely true that only users with the cluster-admin role can create the object. Anyone that has the `create Istio` permission can do it (but initially only the cluster-admin has this permission). 

Anyway, the point of the description is to explain what the user does vs what the operator does, so maybe I should say "user" instead of "cluster-admin"?

rh-tokeefe: Ah, ok, thanks for the clarification. I'm fine if you want to say user instead of cluster-admin, and, in general, I like the distinction of what the user does vs what the operator does.

api/v1alpha1/istio_types.go
luksa: Oh, interesting. I thought it was common, but maybe that's because I always notice it because of the double negative. I'll change it to "similar to".

PR #1540, opened 2023-12-15
Add special version gwControllerMode

jwendell:


config/samples/gwControllerMode.yaml
```diff
@@ -0,0 +1,16 @@
...
+  namespace: istio-gw
+  values:
+    global:
+      hub: quay.io/maistra-dev
+      tag: gwAPIControllerMode
+    pilot:
+      env:
+        PILOT_ENABLE_GATEWAY_CONTROLLER_MODE: true
+        PILOT_GATEWAY_API_DEFAULT_GATEWAYCLASS_NAME: openshift-default
+        PILOT_GATEWAY_API_CONTROLLER_NAME: openshift.io/gateway-controller
```
luksa, line 16: Should this be part of the gw controller mode? This is not directly something that the gw controller needs but something that the ingress team needs.

jwendell: This sample is supposed to contain everything they need to deploy this modified istio. Right now this modified istio contains both changes (controller mode and overriding of gateway class names)

luksa: But they need some other settings as well. And still: gateway-controller mode is not the same as "the openshift gateway api implementation". Gateway controller mode is just one part of the OCP GW API implementation. The other part is the actual gateway class and gateway controller name.  So, I don't think these OCP ingress-specific settings should be in the profile.

luksa: (or, if we want to keep them in the profile, then let's call the profile "ocp-ingress" or "ocp-gateway-api" or something like that. But such a profile doesn't belong in "istio".

jwendell: >So, I don't think these OCP ingress-specific settings should be in the profile.

Which profile? Anyway, maybe we should switch to slack

luksa: Ah, sorry, I thought this file was a profile, but I see that it's just a sample.   Why do we need to provide this sample here?

luksa (approved):

config/samples/gwControllerMode.yaml
jwendell: Strictly speaking we don't need. I just thought it'd be useful to them to have it here, rather than just providing it in a slack thread.

PR #1544, opened 2023-12-18
OSSM-5410 Add update strategy to Istio resource

luksa:


controllers/istio/istio_controller.go
```diff
@@ -160,7 +161,25 @@ func (r *IstioReconciler) getActiveRevision(ctx context.Context, istio *v1alpha1
...
 
 func (r *IstioReconciler) getActiveRevisionKey(istio *v1alpha1.Istio) types.NamespacedName {
-	return types.NamespacedName{Name: istio.Name}
+	return types.NamespacedName{
+		Name: r.getActiveRevisionName(istio),
+	}
+}
+
+func (r *IstioReconciler) getActiveRevisionName(istio *v1alpha1.Istio) string {
+	var strategy v1alpha1.UpdateStrategyType
```
jwendell, line 170: nit: maybe just personal preference, but I'd rather use `strategy :=  v1alpha1.UpdateStrategyTypeInPlace` here. That would make the switch below simpler or even unnecessary...

api/v1alpha1/istio_types.go
```diff
@@ -78,6 +89,30 @@ func (s *IstioSpec) SetValues(values helm.HelmValues) error {
...
+	// +kubebuilder:validation:Enum=InPlace;RevisionBased
+	Type UpdateStrategyType `json:"type,omitempty"`
+
+	// Defines whether the workloads should be moved from one control plane instance to another
+	// automatically. If updateWorkloads is true, the operator moves the workloads from the old
+	// control plane instance to the new one after the new control plane is ready.
+	// If updateWorkloads is false, the user must move the workloads manually by updating the
+	// istio.io/rev labels on the namespace and/or the pods.
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,order=2,displayName="Update Workloads Automatically",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
+	UpdateWorkloads bool `json:"updateWorkloads,omitempty"`
```
jwendell, line 114: Default is false, right? Maybe state this explicitly in the doc above?

api/v1alpha1/istio_types.go
```diff
@@ -35,6 +42,10 @@ type IstioSpec struct {
 	// +kubebuilder:validation:Enum=v1.20.1;v1.20.0;v1.19.5;latest
 	Version string `json:"version"`
 
+	// Defines the update strategy to use when the version in the Istio CR is updated.
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Update Strategy"
+	UpdateStrategy *IstioUpdateStrategy `json:"updateStrategy,omitempty"`
```
jwendell, line 47: if omitted, can we still see the appliedValues somewhere?

jwendell (approved):

controllers/istio/istio_controller.go
luksa: Unfortunately, this wouldn't change anything. The switch statement would still have to have a `default` case with a `fallthrough`, or there would have to be an additional return statement.

api/v1alpha1/istio_types.go
luksa: Not sure what appliedValues have to do with this. If you omit `updateStrategy`, it currently defaults to `InPlace` (later we'll probably change the default to `RevisionBased`). In either case, the applied values are visible in the underlying IstioRevision object.

api/v1alpha1/istio_types.go
luksa: Done.

api/v1alpha1/istio_types.go
jwendell: sorry, not appliedvalues.... my question is: will it be possible to inspect what (default) value was applied for any omitted field?

luksa: Yes. But we have to add a mutatingwebhook, which we currently don't have yet. We'll also need the webhook to set other defaults.

PR #1545, opened 2023-12-18
OSSM-5412 Implement semi-automatic update strategy

luksa:


controllers/istio/istio_controller.go
```diff
@@ -154,19 +159,91 @@ func (r *IstioReconciler) reconcileActiveRevision(ctx context.Context, istio *v1
...
-func (r *IstioReconciler) getActiveRevisionKey(istio *v1alpha1.Istio) types.NamespacedName {
+func (r *IstioReconciler) getNonActiveRevisions(ctx context.Context, istio *v1alpha1.Istio) ([]v1alpha1.IstioRevision, error) {
+	revList := v1alpha1.IstioRevisionList{}
+	if err := r.Client.List(ctx, &revList); err != nil {
+		return nil, err
+	}
+
+	nonActiveRevisions := []v1alpha1.IstioRevision{}
+	for _, rev := range revList.Items {
+		if isRevisionOwnedByIstio(rev, istio) && rev.Name != getActiveRevisionName(istio) {
```
jwendell: nit: store the result of `getActiveRevisionName()` in a variable before the loop

controllers/istiorevision/istiorevision_controller.go
```diff
@@ -133,26 +126,44 @@ func (r *IstioRevisionReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-	if err != nil {
-		err = r.updateStatus(ctx, logger, &rev, values, err)
+	values := rev.Spec.GetValues()
+	if err := validateIstioRevision(rev, values); err != nil {
 		return ctrl.Result{}, err
 	}
 
 	logger.Info("Installing components", "values", values)
-	err = r.installHelmCharts(ctx, &rev, values)
+	err := r.installHelmCharts(ctx, &rev, values)
```
jwendell, line 141: are we ignoring errors here?

controllers/istiorevision/istiorevision_controller.go
```diff
@@ -372,6 +400,109 @@ func (r *IstioRevisionReconciler) determineReadyCondition(ctx context.Context,
...
+	if err != nil {
+		return v1alpha1.IstioRevisionCondition{}, err
+	}
+
+	if isReferenced {
+		return v1alpha1.IstioRevisionCondition{
+			Type:    v1alpha1.IstioRevisionConditionTypeInUse,
+			Status:  metav1.ConditionTrue,
+			Reason:  v1alpha1.IstioRevisionConditionReasonReferencedByWorkloads,
+			Message: "The IstioRevision is referenced by at least one pod or namespace",
```
jwendell: Maybe here and below we could simplify these messages by omitting "the istiorevision". It's kind of implicit that we're seeing the status of a istiorevision object. When inspecting the status, I'd completely understand the message if it was just "Referenced by at least one pod or namespace" or "Not referenced by any pod or namespace".

controllers/istiorevision/istiorevision_controller.go
```diff
@@ -372,6 +400,109 @@ func (r *IstioRevisionReconciler) determineReadyCondition(ctx context.Context,
...
+	return rev.Name == getReferencedRevisionFromNamespace(ns.Labels)
+}
+
+func podReferencesRevision(pod corev1.Pod, ns corev1.Namespace, rev *v1alpha1.IstioRevision) bool {
+	return rev.Name == getReferencedRevisionFromPod(pod.GetLabels(), pod.GetAnnotations(), ns.GetLabels())
+}
+
+func getReferencedRevisionFromNamespace(labels map[string]string) string {
+	if labels["istio-injection"] == "enabled" {
+		return "default"
```
jwendell: Should we start using constants for these values? istio.io/rev, istio-injection, default, etc?

controllers/istiorevision/istiorevision_controller.go
```diff
@@ -372,6 +400,109 @@ func (r *IstioRevisionReconciler) determineReadyCondition(ctx context.Context,
...
+	}
+	revision := labels["istio.io/rev"]
+	if revision != "" {
+		return revision
+	}
+	// TODO: if .Values.sidecarInjectorWebhook.enableNamespacesByDefault is true, then all namespaces except system namespaces should use the "default" revision
```
jwendell, line 485: Should we create tickets for these TODO's so we don't forget about them?

controllers/istiorevision/istiorevision_controller.go
luksa: no, we pass `err` to `updateStatus`, which reports the error via the Istio resource and also returns it, so that its also returned from Reconcile().

controllers/istiorevision/istiorevision_controller.go
luksa: +1 
Done

controllers/istiorevision/istiorevision_controller.go
luksa: Done.

controllers/istiorevision/istiorevision_controller.go
luksa: I created https://issues.redhat.com/browse/OSSM-5572 to address these and any future TODOs.

controllers/istio/istio_controller.go
luksa: Done

controllers/istiorevision/istiorevision_controller.go
```diff
@@ -372,6 +408,109 @@ func (r *IstioRevisionReconciler) determineReadyCondition(ctx context.Context,
...
+func (r *IstioRevisionReconciler) isRevisionReferencedByWorkloads(ctx context.Context, rev *v1alpha1.IstioRevision) (bool, error) {
+	logger := log.FromContext(ctx)
+	nsList := corev1.NamespaceList{}
+	nsMap := map[string]corev1.Namespace{}
+	if err := r.Client.List(ctx, &nsList); err != nil { // TODO: can we optimize this by specifying a label selector
+		return false, err
+	}
+	for _, ns := range nsList.Items {
+		if namespaceReferencesRevision(ns, rev) {
+			logger.Info("Revision is referenced by Namespace", "Namespace", ns.Name)
```
jwendell: nit: maybe these log.info are too much verbosity to have by default? maybe turn them debug messages instead?

config/samples/istio-sample-openshift.yaml
```diff
@@ -1,9 +1,10 @@
...
 kind: Istio
 metadata:
-  name: istio-sample
+  name: default
 spec:
   version: v1.20.1
   namespace: istio-system
   updateStrategy:
     type: InPlace
+    inactiveRevisionDeletionGracePeriodSeconds: 30
```
jwendell, line 10: should we have this field in the samples?

related question: Maybe this field too much implementation detail? What are the use cases users would need to tweak this value?

controllers/istio/istio_controller.go
```diff
@@ -154,19 +159,92 @@ func (r *IstioReconciler) reconcileActiveRevision(ctx context.Context, istio *v1
...
+		inUseCondition := rev.Status.GetCondition(v1alpha1.IstioRevisionConditionTypeInUse)
+		inUse := inUseCondition.Status == metav1.ConditionTrue
+		if inUse {
+			continue
+		}
+
+		pruneTimestamp := inUseCondition.LastTransitionTime.Time.Add(getPruningGracePeriod(istio))
+		expired := pruneTimestamp.Before(time.Now())
+		if expired {
+			err = r.Client.Delete(ctx, &rev)
```
jwendell, line 184: Is this worth a debug or even info log?

controllers/istiorevision/istiorevision_controller.go
```diff
@@ -372,6 +408,109 @@ func (r *IstioRevisionReconciler) determineReadyCondition(ctx context.Context,
...
+		nsMap[ns.Name] = ns
+	}
+
+	podList := corev1.PodList{}
+	if err := r.Client.List(ctx, &podList); err != nil { // TODO: can we optimize this by specifying a label selector
+		return false, err
+	}
+	for _, pod := range podList.Items {
+		if ns, found := nsMap[pod.Namespace]; found && podReferencesRevision(pod, ns, rev) {
+			logger.Info("Revision is referenced by Pod", "Pod", client.ObjectKeyFromObject(&pod))
```
jwendell: ditto

controllers/istiorevision/istiorevision_controller.go
```diff
@@ -372,6 +408,109 @@ func (r *IstioRevisionReconciler) determineReadyCondition(ctx context.Context,
...
+	nsMap := map[string]corev1.Namespace{}
+	if err := r.Client.List(ctx, &nsList); err != nil { // TODO: can we optimize this by specifying a label selector
+		return false, err
+	}
+	for _, ns := range nsList.Items {
+		if namespaceReferencesRevision(ns, rev) {
+			logger.Info("Revision is referenced by Namespace", "Namespace", ns.Name)
+			return true, nil
+		}
+		ns := ns
```
jwendell: 🧐
trick to avoid garbage collection or sth? maybe add a comment so future maintainers don't remove it?

jwendell (approved):
few nit, feel free to address them now, later or never :)

config/samples/istio-sample-openshift.yaml
luksa: Most of the fields in this sample are there because the OpenShift web console uses these values to pre-populate the Create form.

The `inactiveRevisionDeletionGracePeriodSeconds` is something users will probably want to set themselves, unless we find a good default. We should discuss this in the demo tomorrow.

controllers/istiorevision/istiorevision_controller.go
luksa: Oops, no. This one is a left-over and isn't needed anymore. I'll remove it.

controllers/istiorevision/istiorevision_controller.go
luksa: +1 done

controllers/istiorevision/istiorevision_controller.go
luksa: done

controllers/istio/istio_controller.go
luksa: Done. Info, because we also log IstioRevision creation as Info.

PR #1549, opened 2023-12-20
OSSM-5581 Improve & clean up logging

luksa:
* Show controller name instead of "reconciler"
* Remove clutter in log fields
* Don't repeat the same information
* Add more detailed logs to pruneInactiveRevisions()
* Don't show Helm values when installing components, as they make the logs unreadable
* Get logger from context instead of passing via function parameter
* Use log.FromContext() instead of our own implementation
* Use variable name "log" instead of "logger"


controllers/istio/istio_controller.go
```diff
@@ -70,25 +73,25 @@ func NewIstioReconciler(client client.Client, scheme *runtime.Scheme, resourceDi
...
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.1/pkg/reconcile
 func (r *IstioReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	logger := log.FromContext(ctx).WithName("reconciler")
+	log := logf.FromContext(ctx)
 
 	var istio v1alpha1.Istio
 	if err := r.Client.Get(ctx, req.NamespacedName, &istio); err != nil {
 		if errors.IsNotFound(err) {
-			logger.V(2).Info("Istio not found. Skipping reconciliation")
+			log.V(2).Info("Istio not found. Skipping reconciliation")
```
jwendell, line 81: Can you add instructions somewhere (README maybe?) on how to enable verbosity on operator logs?

jwendell (approved):

controllers/istio/istio_controller.go
luksa: Not in this PR.

PR #1552, opened 2023-12-20
OSSM-5584 Move integration tests into a separate package

luksa:
Integration tests should not be in the same package as unit tests, as this makes running only the unit tests much harder. Due to the difference in how fast unit tests execute vs. integration tests, we need to be able to run them separately (but this is out of scope for this PR).


jwendell (commented):
- I would love to see the reasoning for the change directly in the PR description / git commit.
- Why not moving to `tests/integration` or sth like that? That's the approach istio uses.

luksa:
Yaah, I was just in the process of moving them to test/integration.

jwendell (approved):

PR #1557, opened 2023-12-22
OSSM-5587 and OSSM-5585 - IstioRevision and Istio integration tests

luksa:


common/config/.golangci.yml
```diff
@@ -256,8 +256,8 @@ issues:
       linters:
         - errcheck
         - maligned
-    - path: _test\.go$
-      text: "dot-imports: should not use dot imports"
```
jwendell: This change will be overridden, this file is auto generated

luksa: Where's the source then? We made a similar change a few weeks ago.

jwendell:
istio/common-files

I suggest you submit this change upstream so we can benefit from it in the
next sync (could be as fast as today).

Jonh Wendell
https://jwendell.dev

luksa:
This change is not needed upstream.

common/config/.golangci.yml
luksa: Fixed.

tests/integration/istio_test.go
```diff
@@ -0,0 +1,496 @@
...
+
+var _ = Describe("Istio resource", Ordered, func() {
+	const (
+		istioName         = "test-istio"
+		istioNamespace    = "istio-test"
+		workloadNamespace = "istio-test-workloads"
+
+		istioVersion = "latest"
+		oldVersion   = "v1.20.0"
+		newVersion   = "v1.20.1" // TODO: get these versions from versions.yaml
```
dgn: nit, but: why not do it right away?

tests/integration/istio_test.go
```diff
@@ -0,0 +1,496 @@
...
+
+	BeforeAll(func() {
+		Step("Creating the Namespace to perform the tests")
+		Expect(k8sClient.Create(ctx, namespace)).To(Succeed())
+	})
+
+	AfterAll(func() {
+		// TODO(user): Attention if you improve this code by adding other context test you MUST
+		// be aware of the current delete namespace limitations.
+		// More info: https://book.kubebuilder.io/reference/envtest.html#testing-considerations
```
dgn, line 67: following this link, it states that we should suffix namespaces with `-%d` because namespaces can never be deleted properly and this avoids conflicts. We don't seem to be doing that, so I'm wondering how useful this comment is in the end

tests/integration/istio_test.go
```diff
@@ -0,0 +1,496 @@
...
+					},
+					"global":{
+						"configValidation":true,
+						"istioNamespace":"` + istio.Spec.Namespace + `"
+					},
+					"istio_cni": {
+						"enabled":true
+					},
+					"pilot":{
+						"image":"maistra.io/test:latest"
```
dgn: we should probably construct this string from the variables above- or at least `pilotImage` should be used, right?

tests/integration/istio_test.go
```diff
@@ -0,0 +1,496 @@
...
+					},
+					"revision":"` + revKey.Name + `"
```
dgn, line 129: oh I see we're doing it here. then we definitely should do the same for pilotImage

tests/integration/istio_test.go
```diff
@@ -0,0 +1,496 @@
...
+						},
+						"global":{
+							"configValidation":true,
+							"istioNamespace":"` + istio.Spec.Namespace + `"
+						},
+						"istio_cni": {
+							"enabled":true
+						},
+						"pilot":{
+							"image":"maistra.io/test:latest"
```
dgn: same here, use `pilotImage` instead of duplicating the image name

tests/integration/istio_test.go
```diff
@@ -0,0 +1,496 @@
...
+		gracePeriod = 30 * time.Second
+		pilotImage  = "maistra.io/test:latest"
+	)
+	istioKey := client.ObjectKey{Name: istioName}
+	istio := &v1alpha1.Istio{}
+
+	SetDefaultEventuallyTimeout(30 * time.Second)
+	SetDefaultEventuallyPollingInterval(time.Second)
+
+	SetDefaultConsistentlyDuration(30 * time.Second)
```
dgn: This long duration and the usage of `Consistently()` in 6 places really prolongs the tests... Currently this PR adds 5m(!) of execution time to `make test`. If we strip this duration down to `3s`, it is "only" 3m of additional execution time- which IMO is still too much and we should look into shortening it further.

Do we even really need to use `Consistently`? I only noticed this problem because during test execution the operator would be idle for 30s exactly 6 times (the amount of times we call `Consistently()`).

tests/integration/istio_test.go
luksa: done

tests/integration/istio_test.go
luksa: done

tests/integration/istio_test.go
luksa: done

tests/integration/istio_test.go
luksa: Unfortunately, we do, because we're checking that the reconciliation does not happen. There's no other way to confirm that the operator won't do something. The duration needs to be long enough for us to be reasonably sure that the fact that nothing has happened is a result of the operator actually not doing it and not because something caused the operator to stall. 

However, given that a reconciliation typically takes around 2 to 3 seconds, we can probably shorten the duration from 30s to maybe 10s? I don't think we can shorten it to anything less without affecting our confidence in the test. But we could also decide not to test for this at all. Or make the test configurable - for fast tests, we skip these checks; for long tests, we perform them.

tests/integration/istio_test.go
luksa: done

tests/integration/istio_test.go
luksa: We currently run the istio and istiorevision tests in different namespaces. We only create/delete each namespace once, so this isn't currently a problem. But let's improve this in another PR.

dgn: sounds good

tests/integration/istio_test.go
dgn: The change to 10s sounds like a good compromise. We can then see how we can improve testing time in the future. Thanks Marko

dgn (approved):

PR #1558, opened 2023-12-22
OSSM-5444: Fix mapping `spec.tracing.type: None` to helm values

jewertow:
When I was testing zipkin extension provider, I noticed that we incorrectly handle `None` tracing type and as a result istio proxies receive incorrect default config:
```
defaultConfig:
  discoveryAddress: istiod-basic.istio-system.svc:15012
  proxyMetadata:
    ISTIO_META_DNS_AUTO_ALLOCATE: "true"
    ISTIO_META_DNS_CAPTURE: "true"
    PROXY_XDS_VIA_AGENT: "true"
  tracing: null
```
and therefore pilot-agent overwrites `tracing: null` with the default zipkin address, while the expected config is:
```
defaultConfig:
  discoveryAddress: istiod-basic.istio-system.svc:15012
  proxyMetadata:
    ISTIO_META_DNS_AUTO_ALLOCATE: "true"
    ISTIO_META_DNS_CAPTURE: "true"
    PROXY_XDS_VIA_AGENT: "true"
  tracing: {}
```

Actually, it does not affect tracing, because when a `Telemetry` config is applied to a proxy, the default tracing address is not used by Envoy, but it might be confusing when debugging tracing issues.

### DIFF
--- a/pkg/apis/maistra/conversion/jaeger_test.go
+++ b/pkg/apis/maistra/conversion/jaeger_test.go
@@ -39,7 +39,7 @@ func jaegerTestCasesV2(version versions.Version) []conversionTestCase {
 					"meshExpansion": globalMeshExpansionDefaults,
 					"enableTracing": false,
 					"proxy": map[string]interface{}{
-						"tracer": "zipkin",
+						"tracer": "none",
 					},
 				},
 				"meshConfig": map[string]interface{}{
--- a/resources/helm/overlays/global.yaml
+++ b/resources/helm/overlays/global.yaml
@@ -402,15 +402,6 @@ meshConfig:
       ISTIO_META_DNS_AUTO_ALLOCATE: "true"
       # required to enable dns capture
       PROXY_XDS_VIA_AGENT: "true"
-    tracing:
-#      tlsSettings:
-#        mode: DISABLE # DISABLE, SIMPLE, MUTUAL, ISTIO_MUTUAL
-#        clientCertificate: # example: /etc/istio/tracer/cert-chain.pem
-#        privateKey:        # example: /etc/istio/tracer/key.pem
-#        caCertificates:    # example: /etc/istio/tracer/root-cert.pem
-#        sni:               # example: tracer.somedomain
-#        subjectAltNames: []
-        # - tracer.somedomain
 
 base:
   validationURL: ""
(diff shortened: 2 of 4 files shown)


pkg/apis/maistra/conversion/jaeger_test.go
```diff
@@ -39,7 +39,7 @@ func jaegerTestCasesV2(version versions.Version) []conversionTestCase {
 					"meshExpansion": globalMeshExpansionDefaults,
 					"enableTracing": false,
 					"proxy": map[string]interface{}{
-						"tracer": "zipkin",
+						"tracer": "none",
```
yxun, line 42: Is this "none" value case sensitive ? From UI, I see tracing provider type value option is "None"

resources/helm/overlays/global.yaml
```diff
@@ -402,15 +402,6 @@ meshConfig:
...
       PROXY_XDS_VIA_AGENT: "true"
-    tracing:
-#      tlsSettings:
-#        mode: DISABLE # DISABLE, SIMPLE, MUTUAL, ISTIO_MUTUAL
-#        clientCertificate: # example: /etc/istio/tracer/cert-chain.pem
-#        privateKey:        # example: /etc/istio/tracer/key.pem
-#        caCertificates:    # example: /etc/istio/tracer/root-cert.pem
-#        sni:               # example: tracer.somedomain
-#        subjectAltNames: []
-        # - tracer.somedomain
```
yxun, line 413: I feel it's better comment out the line 405 and keep the comment lines 406-413 for tlsSettings example references

pkg/apis/maistra/conversion/jaeger_test.go
yxun: nvm, I see the value lower case is correct one from: https://github.com/istio/istio/issues/45890

jewertow: https://github.com/maistra/istio/blob/maistra-2.5/operator/pkg/apis/istio/v1alpha1/values_types.pb.go#L118

> From UI, I see tracing provider type value option is "None"

This is only our value. Istio uses lowercase.

resources/helm/overlays/global.yaml
jewertow: I removed these values, because they are obsolete and no longer used in the charts.

yxun (approved):

PR #1576, opened 2024-01-12
OSSM-5558 Remove usage of kustomize and replace with helm

dgn:
This really simplifies our manifest creation (300 lines removed) and by using helm it'll also be much easier to introduce new things in the future. The biggest advantage is the ability to pass in a custom values.yaml on the product side to quickly create product manifests from the exact same templates as the community ones.


chart/values.yaml
```diff
@@ -0,0 +1,40 @@
...
+  longDescription: |-
+    This is an experimental operator for installing Istio service mesh.
+
+    This version of the operator supports the following Istio versions:
+
+    - v1.20.1
+    - v1.20.0
+    - v1.19.5
+    - latest (e6756e19)
+    - gwAPIControllerMode
```
luksa, line 20: This is generated from the versions.yaml file by [update-version-list.sh](https://github.com/Maistra/istio-operator/blob/b73f1a3146715bf9c37eb43fdce4bf470ae688b5/hack/update-version-list.sh#L44-L60), so you'll have to change that code so it updates the values.yaml file instead.

dgn: thanks! done

Makefile.core.mk
```diff
@@ -311,25 +339,25 @@ $(LOCALBIN):
...
-KUSTOMIZE ?= $(LOCALBIN)/kustomize
+HELM ?= $(LOCALBIN)/helm
 CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
 ENVTEST ?= $(LOCALBIN)/setup-envtest
 OPM ?= $(LOCALBIN)/opm
 
 ## Tool Versions
   OPERATOR_SDK_VERSION ?= v1.33.0
-  KUSTOMIZE_VERSION ?= v5.3.0
+  HELM_VERSION ?= v3.13.3
```
jwendell, line 349: do we need to install helm? It's already in the builder image.

jwendell (approved):

Makefile.core.mk
dgn: I guess we don't strictly need it here, but I don't use the builder image for local testing so it helps me

jwendell: ok, we can improve this in a follow up, e.g. by only installing it if it's not in $PATH. Not using the builder image is not an excuse to install every tool in the host. That's the whole purpose of the image :)

jwendell: We even have `oc` in the image nowadays :)

PR #1578, opened 2024-01-15
Disable IOR by default in v2.5

yannuil:


mayleighnmyers (approved):

jewertow:
What about existing SMCPs with enabled IOR? Will it be disabled after an upgrade? Did you test it?

rcernich:
As we don't automatically set the field in the SMCP, I'm concerned that a customer updating will go from having IOR enabled to having it disabled if they upgrade from 2.4 to 2.5.  It might be better to leave the charts as they are, but add some code to the mutating webhook to set IOR enablement to false (if unset) in the SMCP during creation.

yannuil:
Close the ticket in favor of https://github.com/maistra/istio-operator/pull/1579

PR #1579, opened 2024-01-18
OSSM-3654 Disable IOR by default in v2.5

yannuil:


pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -237,3 +275,15 @@ func (m *smcpv2mutator) OldVersion() string {
...
+func (m *smcpv2mutator) OldOpenShiftRoute() *bool {
+	return m.oldsmcp.Spec.Gateways.OpenShiftRoute.Enabled
+}
+
+func (m *smcpv2mutator) NewOpenShiftRoute() *bool {
+	return m.smcp.Spec.Gateways.OpenShiftRoute.Enabled
+}
+
+func (m *smcpv2mutator) SetOpenShiftRoute(value bool) {
+	m.patches = append(m.patches, jsonpatch.NewPatch("add", "/spec/gateways/openshiftRoute", value))
```
luksa: Shouldn't this be  `/spec/gateways/openshiftRoute/enabled`?

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -74,6 +74,31 @@ func (v *ControlPlaneMutator) Handle(ctx context.Context, req admission.Request)
...
+					mutator.SetOpenShiftRoute(false)
+				case admissionv1beta1.Update:
+					if oldVersion.LessThan(versions.V2_5) {
+						mutator.SetOpenShiftRoute(true)
+					} else {
+						mutator.SetOpenShiftRoute(false)
+					}
+				}
+			} else {
+				mutator.SetOpenShiftRoute(*oldOpenShiftRoute)
```
luksa: If openshiftRoute is enabled in the old object, why are we applying a patch to enable it?

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -74,6 +74,31 @@ func (v *ControlPlaneMutator) Handle(ctx context.Context, req admission.Request)
...
+			if oldOpenShiftRoute == nil {
+				switch req.AdmissionRequest.Operation {
+				case admissionv1beta1.Create:
+					mutator.SetOpenShiftRoute(false)
+				case admissionv1beta1.Update:
+					if oldVersion.LessThan(versions.V2_5) {
+						mutator.SetOpenShiftRoute(true)
+					} else {
+						mutator.SetOpenShiftRoute(false)
+					}
```
luksa: So, once users get this new operator, any update to the SMCP will set openShiftroute to true (for pre-2.5 versions)? Can't we keep it unset? 

Is there a unit test for this code? I'd like to see examples of what this code does and comments on why it does it, as it's not obvious.

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
yannuil: Because pre 2.5, IOR is enabled by default. If we update the 2.5 charts and the base chart, an user, who is using IOR, could have the routes disabled as described by Rob's [comment](https://github.com/maistra/istio-operator/pull/1578#issuecomment-1892415559).  Using the mutation hook, we can check if it is an upgrade from lower version. If it is, we need to make sure IOR should continue to work.

rcernich:
As I mentioned on the other PR, I think the appropriate thing to do is leave the charts untouched and set IOR to false when a new SMCP is created with version == 2.5.  IMO, this is the simplest approach and requires the smallest amount of changes.

rcernich:
Also note that this will mean IOR will be disabled in existing SMCP until they are modified, meaning somebody updated to 2.5 will have IOR disabled unless they had explicitly enabled it.

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -237,3 +256,23 @@ func (m *smcpv2mutator) OldVersion() string {
...
+		return nil
+	}
+
+	route := gateways.OpenShiftRoute
+
+	if route == nil {
+		return nil
+	}
+
+	return nil
```
rcernich: shouldn't this return route.Enabled?

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -74,6 +74,17 @@ func (v *ControlPlaneMutator) Handle(ctx context.Context, req admission.Request)
 		}
 	}
 
+	// As we are deprecating IOR, on creating a v2.5 SMCP we want to disable IOR if not specified explicitly
+	if req.AdmissionRequest.Operation == admissionv1beta1.Create {
+		newOpenShiftRoute := mutator.NewOpenShiftRoute()
+
+		if newOpenShiftRoute == nil {
+			if versions.V2_5.Version().String() == currentVersion {
```
eoinfennessy: If a new SMCP is created without providing a version, `currentVersion` here will be an empty string. Because of this, IOR will not be disabled when the version is not specified and the default version is 2.5.

Also, should versions greater than 2.5 be considered here too?

pkg/controller/servicemesh/webhooks/mutation/common_test.go
```diff
@@ -14,14 +15,21 @@ import (
...
+	"github.com/maistra/istio-operator/pkg/controller/versions"
 )
 
 var (
 	ctx        = common.NewContextWithLog(context.Background(), logf.Log)
 	testScheme = test.GetScheme()
 )
 
-var acceptWithNoMutation = admission.Allowed("")
+func getDefaultAdmissionResponse() admission.Response {
```
dgn: I think instead of this func we should change the tests to include the expected mutations for every test case. This `getDefaultAdmissionReponse` only works for test cases that don't touch `openshiftRoute.enabled` (which none of them currently do) but it's not really generic enough to warrant the name. Just supplying a spec that explicitly enables openshiftRoute will break this.

We will also adjust the tests properly, apparently the `no mutation 2` test case did actually have mutations on the profile (and now also on the openshiftRoute.enabled field) so I guess the name isn't appropriate anymore. We can add new test cases with no mutation

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
dgn: +1

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -237,3 +256,23 @@ func (m *smcpv2mutator) OldVersion() string {
 func (m *smcpv2mutator) GetProfiles() []string {
 	return m.smcp.Spec.Profiles
 }
+
+func (m *smcpv2mutator) NewOpenShiftRoute() *bool {
```
dgn: I think `func IsOpenShiftRouteEnabled() bool` would make more sense

yannuil: There are new and old SMCP in the mutator. I tried to follow the naming of other getter function to reflect getting value from the new SMCP.

pkg/controller/servicemesh/webhooks/mutation/controlplane_test.go
```diff
@@ -122,6 +131,13 @@ func TestCreate(t *testing.T) {
 
 				mutatedControlPlane := controlPlane.DeepCopy()
 				mutatedControlPlane.Spec.Profiles = []string{maistrav1.DefaultTemplate}
+				mutatedControlPlane.Spec.Gateways = &maistrav2.GatewaysConfig{
```
dgn: instead of comparing the admissionrequests, why not compare the `mutatedControlplane` with the result of applying the patch? you can apply a patch using the github.com/evanphx/json-patch module (we already import it - in fact, gomodules.xyz/jsonpatch uses it for testing, too)

yannuil: I have updated this test to be like others where we list expected patches as part of  the response.

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -148,6 +167,8 @@ type smcpmutator interface {
 	GetProfiles() []string
 	SetProfiles(profiles []string)
 	GetPatches() []jsonpatch.JsonPatchOperation
+	IsOpenShiftRouteEnabled() *bool
```
dgn, line 170: nit: I still think returning a `bool` makes more sense here. Either it is set, or it isn't, a third option isn't required

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -148,6 +167,8 @@ type smcpmutator interface {
...
+	SetOpenShiftRoute(bool)
```
dgn: nit: I prefer `SetOpenShiftRouteEnabled(bool)`

pkg/controller/servicemesh/webhooks/mutation/controlplane_test.go
```diff
@@ -89,54 +124,43 @@ func TestNoMutation(t *testing.T) {
 // Test if the webhook defaults Version to the existing Version on an update
 func TestCreate(t *testing.T) {
 	testCases := []struct {
-		name          string
-		controlPlanes func() (runtime.Object, runtime.Object)
+		name             string
```
dgn, line 127: as `TestCreate` and `TestNoMutation` now both follow the same pattern, maybe they should be merged into one?

dgn (approved):
overall LGTM. minor comments

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
yannuil: When it returns `nil`, it is the user doesn't specify any value for the field and it is also when we want to set the default. I want to know if it is the third option.

pkg/controller/servicemesh/webhooks/mutation/controlplane_test.go
yannuil: Because the test cases under both have short and not so descriptive test names, having a `TestNoMutaion` helps to understand what the tests do.

eoinfennessy (approved):
LGTM!

PR #1580, opened 2024-01-18
Update deps

jwendell:


jwendell:
replaced by https://github.com/maistra/istio-operator/pull/1582

PR #1581, opened 2024-01-19
Prevent nested docker build

asmigala:
When running integration tests with `BUILD_WITH_CONTAINER=1`, the image build would happen inside a nested docker which would fail when trying to mount the source dir.

Upstream has an analogous  solution here: https://github.com/istio/istio/blob/ef5e8019a4b1ccf3ad41522125fdca07bf93c12f/tools/docker-builder/main.go#L245-L246


dgn (approved):
good catch!

PR #1585, opened 2024-01-22
OSSM-5730: Allow creating 2 cluster-wide SMCPs when one of them is a gateway controller

jewertow:
This change aims to allow installing SMCP as a gateway controller and another one as a standard service mesh at the same time. What's more, service mesh should be allowed to be cluster-wide, even though the gateway controller is also cluster-wide, because if gateway controller would be pre-installed in the future, then cluster-wide SMCP could not be installed at all. Cluster-wide service mesh still makes sense, because users can narrow its scope using discovery selectors. There is a risk of overwriting istio-ca-root-cert config maps, so I added a validation which makes sure that 2 SMCPs can't be installed with the default config map name. Beside that we should require customizing the GatewayClass name, but it could be implemented in a follow-up.

### DIFF
--- a/pkg/controller/versions/util.go
+++ b/pkg/controller/versions/util.go
@@ -77,6 +77,7 @@ func validateGlobal(ctx context.Context, version Ver, meta *metav1.ObjectMeta, s
 	}
 
 	if isClusterScoped {
+		// TODO(jewertow): allow to create more than 1 control plane if one of them is a gateway-controller; do not allow to create more than 1 gateway-controllers
 		// allow SMCP create/update only when no SMCP exists or when a single SMCP exists and we're updating it
 		if len(smcps.Items) > 1 || len(smcps.Items) == 1 && smcps.Items[0].UID != meta.GetUID() {
 			return append(allErrors,
--- a/pkg/controller/versions/util.go
+++ b/pkg/controller/versions/util.go
@@ -65,7 +65,7 @@ func checkDiscoverySelectorsNotSet(spec *v2.ControlPlaneSpec, allErrors []error)
 	return allErrors
 }
 
-func validateGlobal(ctx context.Context, version Ver, meta *metav1.ObjectMeta, newSmcp *v2.ControlPlaneSpec, cl client.Client, allErrors []error) []error {
+func validateGlobal(ctx context.Context, meta *metav1.ObjectMeta, newSmcp *v2.ControlPlaneSpec, cl client.Client, allErrors []error) []error {
 	smcps := v2.ServiceMeshControlPlaneList{}
 	err := cl.List(ctx, &smcps)
 	if err != nil {
@@ -115,7 +115,9 @@ func validateGlobal(ctx context.Context, version Ver, meta *metav1.ObjectMeta, n
 	return append(allErrors, validateOverlappingCaCertConfigMapNames(meta, newSmcp, &smcps, allErrors)...)
 }
 
-func validateOverlappingCaCertConfigMapNames(meta *metav1.ObjectMeta, newSmcp *v2.ControlPlaneSpec, smcps *v2.ServiceMeshControlPlaneList, allErrors []error) []error {
+func validateOverlappingCaCertConfigMapNames(
+	meta *metav1.ObjectMeta, newSmcp *v2.ControlPlaneSpec, smcps *v2.ServiceMeshControlPlaneList, allErrors []error,
+) []error {
 	overlappingCaCertNameErr := fmt.Errorf("cannot create cluster-wide SMCP with overlapping caCertConfigMapName")
 	for _, smcp := range smcps.Items {
 		if meta.GetUID() == smcp.GetUID() {
--- a/pkg/controller/versions/util.go
+++ b/pkg/controller/versions/util.go
@@ -78,15 +78,12 @@ func validateGlobal(ctx context.Context, meta *metav1.ObjectMeta, newSmcp *v2.Co
 		if len(smcps.Items) == 1 && smcps.Items[0].UID != meta.GetUID() {
 			currentSmcp := smcps.Items[0].Spec
 			if currentSmcp.IsClusterScoped() {
-				// allow cluster-wide gateway controller when another cluster-wide non gateway controller already exists
+				// allow creating cluster-wide gateway controller when existing cluster-wide SMCP is not a gateway controller and vice versa;
 				// this is the case where openshift-ingress controller and cluster-wide mesh co-exist
-				if (newSmcp.IsGatewayController() && !currentSmcp.IsGatewayController()) || (!newSmcp.IsGatewayController() && currentSmcp.IsGatewayController()) {
-					return append(allErrors, validateOverlappingCaCertConfigMapNames(meta, newSmcp, &smcps, allErrors)...)
-				}
-				// do not allow more than 1 cluster-wide gateway controller
-				if newSmcp.IsGatewayController() && currentSmcp.IsGatewayController() {
+				if newSmcp.IsGatewayController() == currentSmcp.IsGatewayController() {
 					return append(allErrors, otherSmcpExists)
 				}
+				return append(allErrors, validateOverlappingCaCertConfigMapNames(meta, newSmcp, &smcps, allErrors)...)
 			}
 			// only cluster-wide gateway controller can be created when another SMCP exist
 			if !newSmcp.IsGatewayController() {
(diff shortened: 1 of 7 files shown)


pkg/controller/versions/util.go
```diff
@@ -65,41 +65,82 @@ func checkDiscoverySelectorsNotSet(spec *v2.ControlPlaneSpec, allErrors []error)
...
-				fmt.Errorf("a cluster-scoped SMCP may only be created when no other SMCPs exist"))
+	otherSmcpExists := fmt.Errorf("a cluster-scoped SMCP may only be created when no other SMCPs exist")
+	if newSmcp.IsClusterScoped() {
+		// an SMCP already exists and new one is created
+		if len(smcps.Items) == 1 && smcps.Items[0].UID != meta.GetUID() {
+			currentSmcp := smcps.Items[0].Spec
+			if currentSmcp.IsClusterScoped() {
+				// allow cluster-wide gateway controller when another cluster-wide non gateway controller already exists
+				// this is the case where openshift-ingress controller and cluster-wide mesh co-exist
+				if (newSmcp.IsGatewayController() && !currentSmcp.IsGatewayController()) || (!newSmcp.IsGatewayController() && currentSmcp.IsGatewayController()) {
```
dgn: same here, check negatives first, then happy path
```golang
if newSmcp.IsGatewayController() == currentSmcp.IsGatewayController() {
    return append(allErrors, otherSmcpExists)
}
return append(allErrors, validateOverlappingCaCertConfigMapNames(meta, newSmcp, &smcps, allErrors)...)
```

pkg/controller/versions/util.go
```diff
@@ -65,41 +65,82 @@ func checkDiscoverySelectorsNotSet(spec *v2.ControlPlaneSpec, allErrors []error)
...
+					return append(allErrors, otherSmcpExists)
+				}
+			}
+			// only cluster-wide gateway controller can be created when another SMCP exist
+			if !newSmcp.IsGatewayController() {
+				return append(allErrors, otherSmcpExists)
+			}
+		}
+		if len(smcps.Items) > 1 {
+			if newSmcp.IsGatewayController() && countGatewayControllers(smcps.Items) > 1 ||
```
dgn: how would `countGatewayControllers() > 1` ever happen? I'd say `if len(smcps.Items) > 1` we will never allow creation of a cluster-wide smcp, just return the error straight away

pkg/controller/versions/util.go
```diff
@@ -65,41 +65,82 @@ func checkDiscoverySelectorsNotSet(spec *v2.ControlPlaneSpec, allErrors []error)
...
+					return append(allErrors, validateOverlappingCaCertConfigMapNames(meta, newSmcp, &smcps, allErrors)...)
+				}
+				// do not allow more than 1 cluster-wide gateway controller
+				if newSmcp.IsGatewayController() && currentSmcp.IsGatewayController() {
+					return append(allErrors, otherSmcpExists)
+				}
+			}
+			// only cluster-wide gateway controller can be created when another SMCP exist
+			if !newSmcp.IsGatewayController() {
+				return append(allErrors, otherSmcpExists)
```
dgn, line 90: this whole block can be removed if we check above `if !currentSmcp.IsClusterScoped()`

pkg/controller/versions/util.go
```diff
@@ -65,41 +65,82 @@ func checkDiscoverySelectorsNotSet(spec *v2.ControlPlaneSpec, allErrors []error)
...
-		// allow SMCP create/update only when no SMCP exists or when a single SMCP exists and we're updating it
-		if len(smcps.Items) > 1 || len(smcps.Items) == 1 && smcps.Items[0].UID != meta.GetUID() {
-			return append(allErrors,
-				fmt.Errorf("a cluster-scoped SMCP may only be created when no other SMCPs exist"))
+	otherSmcpExists := fmt.Errorf("a cluster-scoped SMCP may only be created when no other SMCPs exist")
+	if newSmcp.IsClusterScoped() {
+		// an SMCP already exists and new one is created
+		if len(smcps.Items) == 1 && smcps.Items[0].UID != meta.GetUID() {
+			currentSmcp := smcps.Items[0].Spec
+			if currentSmcp.IsClusterScoped() {
```
dgn, line 80: I suggest starting with the negative cases, so:
```golang
	if newSmcp.IsClusterScoped() {
		// an SMCP already exists and new one is created
		if len(smcps.Items) == 1 && smcps.Items[0].UID != meta.GetUID() {
			currentSmcp := smcps.Items[0].Spec
			if !currentSmcp.IsClusterScoped() {
				return append(allErrors, otherSmcpExists)
			}
```

pkg/controller/versions/util.go
```diff
@@ -65,41 +65,82 @@ func checkDiscoverySelectorsNotSet(spec *v2.ControlPlaneSpec, allErrors []error)
...
+				// allow cluster-wide gateway controller when another cluster-wide non gateway controller already exists
+				// this is the case where openshift-ingress controller and cluster-wide mesh co-exist
+				if (newSmcp.IsGatewayController() && !currentSmcp.IsGatewayController()) || (!newSmcp.IsGatewayController() && currentSmcp.IsGatewayController()) {
+					return append(allErrors, validateOverlappingCaCertConfigMapNames(meta, newSmcp, &smcps, allErrors)...)
+				}
+				// do not allow more than 1 cluster-wide gateway controller
+				if newSmcp.IsGatewayController() && currentSmcp.IsGatewayController() {
```
dgn: this whole if-clause is not needed anymore if you check the negative cases first

dgn (commented):
thanks Jacek, especially for the good test coverage! I only had time for a first pass, will give it a more thorough look tomorrow. we don't need to allow multi-tenant gateway-controller modes, which simplifies the logic quite a bit; I left some comments to that extent

pkg/controller/versions/util.go
jewertow: Thanks, done.

pkg/controller/versions/util.go
jewertow: This condition is incorrect, because when existing SMCP is multi-tenant we have to check if the new one is cluster-wide gateway controller, so it could be:
```
	if newSmcp.IsClusterScoped() {
		// an SMCP already exists and new one is created
		if len(smcps.Items) == 1 && smcps.Items[0].UID != meta.GetUID() {
			currentSmcp := smcps.Items[0].Spec
			if !currentSmcp.IsClusterScoped() && !newSmcp.IsGatewayController() {
				return append(allErrors, otherSmcpExists)
			}
```

pkg/controller/versions/util.go
jewertow: As you can see in another comment, this condition would have to be merged with `!currentSmcp.IsClusterScoped()`.

pkg/controller/versions/util.go
jewertow: Thanks, done

pkg/controller/versions/util.go
jewertow: You're right. I added a test case, which cannot exist and then I implemented conditions, and I didn't notice that it does not make any sense.
```
{
        name: "creating cluster-wide gateway controller SMCP when another already exists - expected error (2nd execution)",
        smcp: clusterWideGatewayController,
        existingObjs: []*maistrav2.ServiceMeshControlPlane{
                NewV2SMCPResource("basic", "istio-system-1", clusterWideGatewayController),
                NewV2SMCPResource("basic", "istio-system-2", clusterWideGatewayController2),
        },
        expectedErr: fmt.Errorf("a cluster-scoped SMCP may only be created when no other SMCPs exist"),
},
```

dgn:
I'm not sure we really need to cover multitenant meshes- we could probably say they're incompatible with GW-API support. That would simplify things as we don't have to mix cluster-wide with multitenant in the same cluster

pkg/controller/versions/util.go
jewertow: I was wrong. We need that test case as well, because 2 gateway-controllers can exist when a user creates: gateway-controller and simple cluster-wide and then changes the 2nd to gateway-controller. That's why more than 1 gateway-controller can be found and in such a case we have to return an error and stop reconciliation.

jewertow: Actually, I could simplify this condition and implement this validation in [ValidateUpdate](https://github.com/maistra/istio-operator/pull/1519/files#diff-e668dab162c3d0f9417174bd4eecacac65846a7279393b7fe61d318c52a7a08b), which is called when SMCP mode is changed. That approach would be better, because currently, users will be able to update the SMCP without any error, but the reconciliation will fail. If I implemented this check in ValidateUpdate, users would get validation error.

PR #1590, opened 2024-01-24
OSSM-5685 Support offline installation through default image digests

dgn:
This adds the support for defining default image digests in the operator config, similar to how it worked in 2.x - however, in this new implementation it is much easier for a user to overwrite an image/hub/tag for one component and still use the default images for other components.

### DIFF
--- a/controllers/istio/istio_controller_test.go
+++ b/controllers/istio/istio_controller_test.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/kubectl/pkg/scheme"
 	v1alpha1 "maistra.io/istio-operator/api/v1alpha1"
+	"maistra.io/istio-operator/pkg/common"
 	"maistra.io/istio-operator/pkg/helm"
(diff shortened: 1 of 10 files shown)


controllers/istio/istio_controller_test.go
```diff
@@ -1328,6 +1329,182 @@ func TestMergeOverwrite(t *testing.T) {
...
+	testCases := []struct {
+		name         string
+		config       common.OperatorConfig
+		inputIstio   *v1alpha1.Istio
+		inputValues  helm.HelmValues
+		expectValues helm.HelmValues
+	}{
+		{
+			name: "no-user-values",
+			config: common.OperatorConfig{
```
jwendell, line 1363: can you add a test with no (or empty) `config`?

jwendell (commented):
LGTM (just a nit), will defer to Marko. Let me know if you want me to approve it right away

controllers/istio/istio_controller_test.go
dgn: done!

luksa (approved):

PR #1591, opened 2024-01-24
OSSM-5815: Adding VALUES_FILE var to be able to overwrite default values

FilipB:
default values

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -112,6 +112,13 @@ ifeq ($(USE_IMAGE_DIGESTS), true)
 	BUNDLE_GEN_FLAGS += --use-image-digests
 endif
 
+# Default flags used when rendering chart templates locally
+HELM_TEMPL_DEF_FLAGS = --include-crds --set image="${IMAGE}"
+# VALUES_FILE_OVERWRITE defines a values file to be used to overwrite default values from chart
+ifdef VALUES_FILE_OVERWRITE
+	HELM_TEMPL_DEF_FLAGS += --values $(VALUES_FILE_OVERWRITE)
+endif
+
 TODAY ?= $(shell date -I)
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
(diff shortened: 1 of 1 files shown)


dgn (approved):
exactly what I had in mind, thank you!

Makefile.core.mk
```diff
@@ -112,6 +112,13 @@ ifeq ($(USE_IMAGE_DIGESTS), true)
 	BUNDLE_GEN_FLAGS += --use-image-digests
 endif
 
+# Default flags used when rendering chart templates locally
+HELM_TEMPL_DEF_FLAGS = --include-crds --set image="${IMAGE}"
+# VALUES_FILE_OVERWRITE defines a values file to be used to overwrite default values from chart
+ifdef VALUES_FILE_OVERWRITE
```
asmigala: Maybe `VALUES_FILE_OVERRIDE` would be a slightly better name?

dgn: actually just `VALUES_FILE` would be sufficient IMO

FilipB: When using `VALUES_FILE` is it clear that those values will override default values? I'm new to helm so I'm not familiar with typical usage patterns.

dgn: yeah, a chart's default values are always taken into account even when supplying a values file or a set of custom values on the CLI. so I think it's clear enough

jwendell:
late to the game, but it would be helpful to have this covered in the integration tests.

dgn:
> late to the game, but it would be helpful to have this covered in the integration tests.

what do you want to cover? the helm values file in question will live in a downstream repo

jwendell:
We should make sure this feature (overriding a values yaml file) works fine. This could be achieved by having a simple test that makes use of an alternate values file, deploy the operator and verifies the values were applied as expected.

dgn:
I guess we could do that, could come in handy once we're migrating this to istio-ecosystem, to make sure no contribution accidentally breaks our product builds. I don't think it is required now, though- the --values flag is helm core functionality and will have enough coverage on their side

jwendell:
Yes, this is important not only to us, but to every potential user of this operator willing to provide their own values.

PR #1597, opened 2024-01-26
OSSM-5570 Add values and meshConfig types

bmangoen:


api/v1alpha1/values_types.go
```diff
@@ -0,0 +1,670 @@
...
+	// Controls whether the server-side validation is enabled.
+	ConfigValidation                bool     `json:"configValidation,omitempty"`
+	DefaultConfigVisibilitySettings []string `json:"defaultConfigVisibilitySettings,omitempty"`
+	// Specifies the docker hub for Istio images.
+	Hub string `json:"hub,omitempty"`
+	// Specifies the image pull policy for the Istio images. one of Always, Never, IfNotPresent.
+	// Defaults to Always if :latest tag is specified, or IfNotPresent otherwise. Cannot be updated.
+	//
+	// More info: https://kubernetes.io/docs/concepts/containers/images#updating-images
+	ImagePullPolicy  string   `json:"imagePullPolicy,omitempty"` // ImagePullPolicy             v1.PullPolicy                 `json:"imagePullPolicy,omitempty"`
```
luksa: Why not use the k8s type here?

api/v1alpha1/values_types.go
```diff
@@ -0,0 +1,670 @@
...
+	InjectionPath string `json:"injectionPath,omitempty"`
+}
+
+type Values struct {
+	Cni    *CNIConfig    `json:"cni,omitempty"`
+	Global *GlobalConfig `json:"global,omitempty"`
+	Pilot  *PilotConfig  `json:"pilot,omitempty"`
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Schemaless
+	Ztunnel map[string]string `json:"ztunnel,omitempty"`
```
luksa: Is there a schema for this? If not, can we create one?

At the very least, this should probably be a `map[string]any`.

api/v1alpha1/values_types.go
```diff
@@ -0,0 +1,670 @@
...
+	// Controls whether telemetry is exported for Pilot.
+	Telemetry              *TelemetryConfig       `json:"telemetry,omitempty"`
+	SidecarInjectorWebhook *SidecarInjectorConfig `json:"sidecarInjectorWebhook,omitempty"`
+	IstioCni               *CNIConfig             `json:"istio_cni,omitempty"`
+	Revision               string                 `json:"revision,omitempty"`
+	OwnerName              string                 `json:"ownerName,omitempty"`
```
luksa: This is used for the `install.operator.istio.io/owning-resource` label. Should we remove this value from the CRD and set it to some fixed value when we render the charts?

api/v1alpha1/values_types.go
bmangoen: I kept the same type as it is defined in [values_types.pb.go](https://github.com/istio/istio/blob/master/operator/pkg/apis/istio/v1alpha1/values_types.pb.go#L1508) (defined as a `string` in  [values_types.proto](https://github.com/istio/istio/blob/master/operator/pkg/apis/istio/v1alpha1/values_types.proto#L345))

bmangoen: We decided to remove all the kubernetes related fields

api/v1alpha1/values_types.go
bmangoen: We decided to remove all the Ambient related fields for now

luksa (approved):

PR #1617, opened 2024-02-07
OSSM-5849 - add infrastructure feature related annotations

DeanDoyle1502:
- Are the values for annotations correct?

- The Jira states that these new annotations are to replace: `operators.openshift.io/infrastructure-features` Should this be removed or is it required for older ocp versions?

### DIFF
--- a/build/generate-manifests.sh
+++ b/build/generate-manifests.sh
@@ -26,16 +26,16 @@ if [[ ${COMMUNITY} == "true" ]]; then
   OLM_FEATURES="[]"
   ARCHITECTURE_LABELS=$(generateArchitectureLabels amd64)
   OLM_SUBSCRIPTION_ANNOTATION=""
-  OLM_FEATURE_DISCONNECTED='"false"'
-  OLM_FEATURE_FIPS_COMPLIANT='"false"'
-  OLM_FEATURE_PROXY_AWARE='"true"'
-  OLM_FEATURE_TLS_PROFILES='"false"'
-  OLM_FEATURE_TOKEN_AUTH_AWS='"false"'
-  OLM_FEATURE_TOKEN_AUTH_AZURE='"false"'
-  OLM_FEATURE_TOKEN_AUTH_GCP='"false"'
-  OLM_FEATURE_CNF='"false"'
-  OLM_FEATURE_CNI='"true"'
-  OLM_FEATURE_CSI='"false"'
+  OLM_FEATURE_DISCONNECTED="false"
+  OLM_FEATURE_FIPS_COMPLIANT="false"
+  OLM_FEATURE_PROXY_AWARE="true"
+  OLM_FEATURE_TLS_PROFILES="false"
+  OLM_FEATURE_TOKEN_AUTH_AWS="false"
+  OLM_FEATURE_TOKEN_AUTH_AZURE="false"
+  OLM_FEATURE_TOKEN_AUTH_GCP="false"
+  OLM_FEATURE_CNF="false"
+  OLM_FEATURE_CNI="true"
+  OLM_FEATURE_CSI="false"
 else
   BUILD_TYPE="servicemesh"
   JAEGER_STORAGE="Memory"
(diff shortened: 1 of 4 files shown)


jewertow:
> Do these changes need to be added maistra-2.4 and maistra-2.3 branches

No, we release only the latest version of the operator.

@FilipB could you answer previous questions?

yxun:
You should run `make gen` and then that gencheck job should pass.

eoinfennessy:
To generate the annotations in the CSV you will need to do the following:
1. Add double-underscore placeholders in the  [clusterserviceversion.yaml](https://github.com/maistra/istio-operator/blob/maistra-2.5/build/manifest-templates/clusterserviceversion.yaml) template that will be replaced with each annotation e.g. `__OLM_FEATURE_FIPS_COMPLIANT__`
2. Set a value for each [here in generate-manifests.sh](https://github.com/maistra/istio-operator/blob/maistra-2.5/build/generate-manifests.sh#L30-L39) e.g. `OLM_FEATURE_FIPS_COMPLIANT="features.operators.openshift.io/fips-compliant: \"true\""`
3. If `$COMMUNITY == "true"` delete the lines with these placeholders; otherwise, [use sed to find and replace](https://github.com/maistra/istio-operator/blob/maistra-2.5/build/generate-manifests.sh#L108-L144) the placeholders with the annotation values from step two.

Run `make generate-manifests` to see the effect of each change you make reflected in the CSV.

yxun:
For reference: 
https://docs.openshift.com/container-platform/4.14/operators/operator_sdk/osdk-generating-csvs.html#osdk-csv-annotations-infra_osdk-generating-csvs

> These annotations are supported in OpenShift Container Platform 4.10 and later.

yxun:
> These annotations are supported in OpenShift Container Platform 4.10 and later.

So we should be fine by removing old `operators.openshift.io/infrastructure-features` annotation unless still running our operator in OCP prior to 4.10.

You may ask in team channel to confirm that we don't have any user still running OCP prior to 4.10

build/generate-manifests.sh
```diff
@@ -37,6 +37,17 @@ else
...
+  OLM_FEATURE_DISCONNECTED="features.operators.openshift.io/disconnected: \"true\""
+  OLM_FEATURE_FIPS_COMPLIANT="features.operators.openshift.io/fips-compliant: \"true\""
+  OLM_FEATURE_PROXY_AWARE="features.operators.openshift.io/proxy-aware: \"false\""
+  OLM_FEATURE_TLS_PROFILES="features.operators.openshift.io/tls-profiles: \"false\""
+  OLM_FEATURE_TOKEN_AUTH_AWS="features.operators.openshift.io/token-auth-aws: \"false\""
+  OLM_FEATURE_TOKEN_AUTH_AZURE="features.operators.openshift.io/token-auth-azure: \"false\""
+  OLM_FEATURE_TOKEN_AUTH_GCP="features.operators.openshift.io/token-auth-gcp: \"false\""
+  OLM_FEATURE_CNF="features.operators.openshift.io/cnf: \"false\""
+  OLM_FEATURE_CNI="features.operators.openshift.io/cni: \"false\""
+  OLM_FEATURE_CSI="features.operators.openshift.io/csi: \"false\""
```
luksa: To stay consistent, these variable names should end with `_ANNOTATION` just like the existing `OLM_SUBSCRIPTION_ANNOTATION` variable. 

Alternatively, instead of including the entire annotation, you could also just keep the annotation value, just like we do for `OLM_FEATURES`, which is used to set the value of the `operators.openshift.io/infrastructure-features` annotation.

build/generate-manifests.sh
```diff
@@ -143,6 +154,30 @@ $RELATED_IMAGES"
...
+    sed -i '/__OLM_FEATURE_DISCONNECTED__/d' "${csv_path}"
+    sed -i '/__OLM_FEATURE_FIPS_COMPLIANT__/d' "${csv_path}"
+    sed -i '/__OLM_FEATURE_PROXY_AWARE__/d' "${csv_path}"
+    sed -i '/__OLM_FEATURE_TLS_PROFILES__/d' "${csv_path}"
+    sed -i '/__OLM_FEATURE_TOKEN_AUTH_AWS__/d' "${csv_path}"
+    sed -i '/__OLM_FEATURE_TOKEN_AUTH_AZURE__/d' "${csv_path}"
+    sed -i '/__OLM_FEATURE_TOKEN_AUTH_GCP__/d' "${csv_path}"
+    sed -i '/__OLM_FEATURE_CNF__/d' "${csv_path}"
+    sed -i '/__OLM_FEATURE_CNI__/d' "${csv_path}"
+    sed -i '/__OLM_FEATURE_CSI__/d' "${csv_path}"
```
luksa: Do we really need to remove these from the community CSV? There's nothing wrong with including these annotations in the community CSV.

eoinfennessy: I suggested this because currently the `OLM_SUBSCRIPTION_ANNOTATION` is removed in the community CSV and `OLM_FEATURES` is set to an empty array, although maybe this would imply that the annotations should be provided with everything set to false? Should the values for these annotations be set differently in the community CSV?

luksa: I see. Yes, I believe the community version should have different values (e.g. it isn't FIPS compliant and doesn't support disconnected installs). 

But I've also just realized that the whole point of this PR is to add all these annotations, since they are required. If they are required, they are required in both the product and the community CSV, right?

eoinfennessy: Ah, okay! I'm sorry - I didn't realise these were required in the community version too. In that case it makes sense to set these as you suggested in your comment above, by having a hard-coded annotation key in the template for each annotation and a placeholder plus variable for its associated value.

eoinfennessy (approved):
Looks good! @luksa , can you please re-review and verify the annotation values provided are correct?

build/generate-manifests.sh
```diff
@@ -26,6 +26,16 @@ if [[ ${COMMUNITY} == "true" ]]; then
   OLM_FEATURES="[]"
   ARCHITECTURE_LABELS=$(generateArchitectureLabels amd64)
   OLM_SUBSCRIPTION_ANNOTATION=""
+  OLM_FEATURE_DISCONNECTED='"false"'
```
luksa: Any special reason why this is quoted twice (i.e. why is the double quote included here instead of the CSV template)?

This is inconsistent with `OLM_FEATURES` (see my next comment).

build/manifest-templates/clusterserviceversion.yaml
```diff
@@ -19,6 +19,16 @@ metadata:
     olm.skipRange: ">=1.0.2 <__SEMANTIC_VERSION__"
     operators.openshift.io/infrastructure-features: '__OLM_FEATURES__'
     __OLM_SUBSCRIPTION_ANNOTATION__
+    features.operators.openshift.io/disconnected: __OLM_FEATURE_DISCONNECTED__
```
luksa: The `operators.openshift.io/infrastructure-features` annotation is quoted, whereas `features.operators.openshift.io/disconnected` isn't. In other words, the `OLM_FEATURES` variable does not contain quotes, whereas  `OLM_FEATURE_DISCONNECTED` does. 

I suggest adding the quotes here instead of having them in `OLM_FEATURE_DISCONNECTED`.

Same for all other variables introduced in this PR.

DeanDoyle1502: Thank guys fixed in the next commit

luksa (approved):

luksa:
@DeanDoyle1502 LGTM. Don't forget to add the `okay to merge` label to actually merge this PR.

PR #1623, opened 2024-02-08
OSSM-5889 and OSSM-5886 Remove Istio.spec.rawValues and introduce values.pilot.resources

luksa:
This PR contains two changes:
- introduce Istio.spec.values.pilot.resources
- remove Istio.spec.rawValues

The first change is needed so that we can replace the use of `rawValues` in `istio-sample-kubernetes.yaml` with `values`.

I suggest reviewing each commit separately.


controllers/istio/istio_controller.go
```diff
@@ -288,27 +288,24 @@ func getActiveRevisionName(istio *v1alpha1.Istio) string {
 }
 
 func computeIstioRevisionValues(istio v1alpha1.Istio, defaultProfiles []string, resourceDir string) (helm.HelmValues, error) {
-	// 1. get userValues from Istio.spec.values
```
dgn, line 291: nit but why remove the numbering?

dgn (approved):

controllers/istio/istio_controller.go
luksa: just so we don't have to update all the lines every time we add/remove an item.

PR #1625, opened 2024-02-08
OSSM-5885 Fix the json patch to enable OpenShiftRoute

yannuil:


yxun (approved):

rcernich:
Will this work if the `gateways` field is specified?  Did you test that case?

yannuil:
Yes. I manually tested for both case, with `spec.gateways` absent and `spec.gateways` with some values.

PR #1628, opened 2024-02-09
OSSM-5887 Make IstioRevision.spec.values strongly typed

luksa:


hack/download-charts.sh
```diff
@@ -107,7 +107,24 @@ function convertIstioProfiles() {
       | del(.spec.components)
       | del(.spec.meshConfig)
       | del(.spec.hub)
-      | del(.spec.tag)' "$profile"
+      | del(.spec.tag)
```
dgn, line 110: why these changes?

luksa: The [ValuesFromHelmValues](https://github.com/maistra/istio-operator/pull/1628/files#diff-199134d8b522038233e83e025a767a73162d99ae7c9f96a198d810bd0055aa88R616) function throws an error if it finds a field that isn't in the schema, so I had to remove these fields. This change is temporary, as the reason most of these fields are invalid is because we haven't yet included k8s-related stuff in Values. Another reason is the fact that the current Values types are incorrect in a few places. The plan is to add the k8s fields and fix the invalid fields when we implement the automatic conversion from upstream values to our own Values. It doesn't make sense to fix these things before then.

luksa: Note: I already have a commit that adds all the k8s fields that appear in profiles. I'm waiting for this PR to be merged so I can submit it. In that PR, I'll remove a lot of these removals.

dgn (approved):
LGTM, Sorry for the holdup

PR #1630, opened 2024-02-12
OSSM-5848 Added feature annotations

DeanDoyle1502:
Added feature annotations

- Wondering if the values are correct

### DIFF
--- a/bundle/manifests/sailoperator.clusterserviceversion.yaml
+++ b/bundle/manifests/sailoperator.clusterserviceversion.yaml
@@ -23,8 +23,18 @@ metadata:
     capabilities: Seamless Upgrades
     categories: OpenShift Optional, Integration & Delivery, Networking, Security
     containerImage: quay.io/maistra-dev/istio-operator:3.0-latest
-    createdAt: "2024-02-08T11:24:58Z"
+    createdAt: "2024-02-12T10:02:23Z"
     description: Experimental operator for installing Istio service mesh
+    features.operators.openshift.io/cnf: "false"
+    features.operators.openshift.io/cni: "true"
+    features.operators.openshift.io/csi: "false"
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "true"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
     operators.operatorframework.io/builder: operator-sdk-v1.33.0
     operators.operatorframework.io/internal-objects: '["wasmplugins.extensions.istio.io","destinationrules.networking.istio.io","envoyfilters.networking.istio.io","gateways.networking.istio.io","proxyconfigs.networking.istio.io","serviceentries.networking.istio.io","sidecars.networking.istio.io","virtualservices.networking.istio.io","workloadentries.networking.istio.io","workloadgroups.networking.istio.io","authorizationpolicies.security.istio.io","peerauthentications.security.istio.io","requestauthentications.security.istio.io","telemetries.telemetry.istio.io"]'
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v4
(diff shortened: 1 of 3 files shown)


chart/values.yaml
```diff
@@ -32,6 +32,16 @@ csv:
...
+  olmFeatureDisconnected: true
+  olmFeatureFIPSCompliant: true
+  olmFeatureProxyAware: true
+  olmFeatureTLSProfiles: false
+  olmFeatureTokenAuthAWS: false
+  olmFeatureTokenAuthAzure: false
+  olmFeatureTokenAuthGCP: false
+  olmFeatureCNF: false
+  olmFeatureCNI: true
+  olmFeatureCSI: false
```
luksa: We should probably group these under a common parent (e.g. `olmFeatures`) and remove the `olmFeature` prefix.

bundle/manifests/sailoperator.clusterserviceversion.yaml
```diff
@@ -23,8 +23,18 @@ metadata:
...
     categories: OpenShift Optional, Integration & Delivery, Networking, Security
     containerImage: quay.io/maistra-dev/istio-operator:3.0-latest
-    createdAt: "2024-02-08T11:24:58Z"
+    createdAt: "2024-02-12T10:02:23Z"
     description: Experimental operator for installing Istio service mesh
+    features.operators.openshift.io/cnf: "false"
+    features.operators.openshift.io/cni: "true"
+    features.operators.openshift.io/csi: "false"
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "true"
```
luksa: This should probably be false, at least for now (Sail doesn't yet use our build of the proxy). Also note that the Sail operator is the equivalent of the community operator in 2.x. I believe we set the values for the product build outside of this repository. @FilipB should know more about this.

DeanDoyle1502: Thanks Marko changes made there

chart/values.yaml
```diff
@@ -32,6 +32,18 @@ csv:
...
+    disconnected: false
+    FIPSCompliant: false
+    proxyAware: false
+    TLSProfiles: false
+    tokenAuthAWS: false
+    tokenAuthAzure: false
+    tokenAuthGCP: false
+    CNF: false
+    CNI: false
+    CSI: false
```
luksa: @dgn do we want these to start with a lowercase character?

bundle/manifests/sailoperator.clusterserviceversion.yaml
FilipB: Better to check with OLM team, but I also think we should not include anything OCP specific to the community operator. On the other hand we will want to run community operator on both OCP and K8s so question for OLM team is how to do that.

dgn:
I would prefer to have these in values.yaml with their full names and values through a generic `csv.annotations` field

DeanDoyle1502:
-@dgn Added new commit there with what I think you were looking for , if its not i can change again.

-Still having trouble getting answer for values

chart/values.yaml
```diff
@@ -32,6 +32,17 @@ csv:
   - istio
   - servicemesh
   - envoy
+  annotations: 
+    olmFeatureDisconnected: false
```
dgn: what I meant is to have the actual annotations in here (e.g. `features.operators.openshift.io/disconnected: true`) and in the template just looping over the annotation key-value pairs and adding them

dgn: similar to how env vars are added in this upstream chart:
https://github.com/istio/istio/blob/ef08bf9850209ed8ecac3065a92b40427d9425f0/manifests/charts/istio-control/istio-discovery/templates/deployment.yaml#L151-L156

dgn:
For the actual values, we can set `disconnected` and `cni` to `true` for now and leave everything else at `false`

chart/values.yaml
DeanDoyle1502: Made those changes as per what i think your looking for, let me know if there's anything not correct 

-changed values of disconnected and cni to true also

luksa (approved):

dgn (approved):
Thank you @DeanDoyle1502!

FilipB (approved):
We have to remember to set fips to true when we are done with fips testing.

PR #1634, opened 2024-02-19
Bump Istio releases 1.19 and 1.20

jwendell:


dgn:
I wonder how much bigger our container images get each time we add a set of charts

dgn (approved):

PR #1643, opened 2024-02-21
[WIP]OSSM-5881 Do not require ingress and egress gateways when `spec.cluster.multiCluster` is enabled

mayleighnmyers:
https://issues.redhat.com/browse/OSSM-5881

Currently, users cannot disable ingress and egress gateways when multi-cluster is enabled, because setting spec.gateways.enabled is ignored when spec.cluster.multiCluster is enabled. 

Acceptance criteria:
When the SMCP above is applied, ingress and egress gateways are not deployed.


jewertow (commented):
Multi-cluster settings should not affect ingress and egress gateways at all, so you can remove these patches.

pkg/apis/maistra/conversion/common.go
mayleighnmyers, line 1: Changed valmap to val since there was a tautological condition.

pkg/apis/maistra/conversion/security_test.go
```diff
@@ -1034,3 +1034,79 @@ func TestSecurityConversionFromV2ToV1(t *testing.T) {
...
+multiClusterOverrides:
+  expansionEnabled:
+spec:
+  cluster:
+    multiCluster:
+      enabled: true
+  gateways:
+    enabled: false
+    useILB: false
+`),
```
jewertow: These values are obviously incorrect, so this test fails as expected.

pkg/apis/maistra/conversion/security_test.go
```diff
@@ -1034,3 +1034,79 @@ func TestSecurityConversionFromV2ToV1(t *testing.T) {
...
+var multiClusterTestCases = []multiClusterTestCase{
+	{
+		name: "MultiCluster_v2_4",
+		spec: &v2.ControlPlaneSpec{
+			Version: versions.V2_4.String(),
+			Security: &v2.SecurityConfig{
+				CertificateAuthority: &v2.CertificateAuthorityConfig{
+					Type: v2.CertificateAuthorityTypeCertManager,
+					CertManager: &v2.CertManagerCertificateAuthorityConfig{
+						Address: "",
```
jewertow: Security and cert-manager?

jewertow (changes requested):

pkg/apis/maistra/conversion/security_test.go
mayleighnmyers: When I run the test locally it is asking me to add a bunch of stuff, should I keep it to what you had put into the jira?

jewertow: > When I run the test locally it is asking me to add a bunch of stuff

You should never adjust expected test data to the actual result just to make the test pass. Your test and data should drive the implementation, so first fix the expected helm values and the input (SMCP spec) and then work on the implementation until it passes for the correct input and output.

> should I keep it to what you had put into the jira?

In JIRA I added only SMCP spec. You can use that YAML to properly configure input SMCP in your test.

pkg/apis/maistra/conversion/security_test.go
jewertow, line 1: You modified cluster.go, so I would expect to see appropriate tests in cluster_test.go.

jewertow (changes requested):
It's a good direction, but I can see that many tests fail, including the new one:
```
--- FAIL: TestMultiClusterGatewaysDisabled (0.00s)
    --- FAIL: TestMultiClusterGatewaysDisabled/MultiCluster_v2_4-v2_to_v1 (0.00s)
    --- FAIL: TestMultiClusterGatewaysDisabled/MultiCluster_v2_4-v1_to_v2 (0.00s)
```
Can you fix these tests?

jewertow:
Hey @mayleighnmyers, I am closing this PR, because we will release OSSM 3 soon with GA support for multi-cluster deployments, so it's no longer needed.

PR #1650, opened 2024-02-26
OSSM-5988 Skip go.mod updates until cachito has been updated

dgn:
This reverses some of the automator changes to go.mod. It's not a problem as we don't really use the istio.io dependency for a lot of things (mostly just their "ptr" package).

We might want to look at replacing its usage entirely if we don't need it. Or just pin it to a stable version.


FilipB (approved):
LGTM

jwendell (commented):
We should not limit / slowdown the development pace of operator because of a limitation in downstream build environment.

Hopefully we will be able to handle things like this in the fork.

PR #1657, opened 2024-02-28
OSSM-5894 Add tool for transforming upstream values types to the operator API

luksa:


luksa:
**The best way to review this is to review the 2nd commit instead of the entire PR.**

The other two commits are there to make the 2nd commit cleaner:
- The 1st commit minimizes the changes made by the 2nd commit.
- The 3rd commit merely merges the three types files into a single values_types.gen.go file (it contains no changes to the API or the tool).

Makefile.core.mk
```diff
@@ -302,6 +302,26 @@ deploy-example-kubernetes: ## Deploy an example Istio resource on Kubernetes
 gen-manifests: controller-gen ## Generate WebhookConfiguration and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) crd:allowDangerousTypes=true webhook paths="./..." output:crd:artifacts:config=chart/crds
 
+# TODO: move this to versions.yaml or get the files via go.mod instead of downloading them
+ISTIO_REPO_BASE=https://raw.githubusercontent.com/istio/istio/0e7ecbd31f9524b063ced1f49f1a6f6e063d2bf5
+API_REPO_BASE=https://raw.githubusercontent.com/istio/api/ccd5cd40965ccba232d1f7c3b0e4ecacd0f6ceda
+.PHONY: gen-api
+gen-api: ## Generate API types from upstream files
+	# TODO: should we get these files from the local filesystem by inspecting go.mod?
```
dgn, line 310: I don't think we should. we already had problems because of our dependency on `istio/istio` and we shouldn't enforce always using the latest just because we want to use the latest charts - it's not necessary and might force us to make unrelated changes (like in this case upgrade to golang 1.22- which in itself isn't bad but not strictly required for installing latest istio)

Makefile.core.mk
```diff
@@ -302,6 +302,26 @@ deploy-example-kubernetes: ## Deploy an example Istio resource on Kubernetes
 gen-manifests: controller-gen ## Generate WebhookConfiguration and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) crd:allowDangerousTypes=true webhook paths="./..." output:crd:artifacts:config=chart/crds
 
+# TODO: move this to versions.yaml or get the files via go.mod instead of downloading them
```
dgn, line 305: my first instinct is to use versions.yaml, I don't like using go.mod (see comment below).

Using versions.yaml would create the problem though that if we're tracking `master` here, but an older version downstream, the API will diverge. that would be correct- but we might want to discuss that more to see what the exact workflow would look like for downstream. it's also another reason why go.mod is maybe not a good choice- we'd have to downgrade the istio/istio dependency in the fork.

for a start, using a fixed commit is perfect though

hack/api_transformer/main.go
```diff
@@ -0,0 +1,884 @@
...
+					if toString(field.Type) == "*intstr.IntOrString" {
+						addTag(field, "// +kubebuilder:validation:XIntOrString")
+					}
+
+					if field.Doc != nil {
+						for _, comment := range field.Doc.List {
+							if strings.HasPrefix(comment.Text, "// REQUIRED.") {
+								addTag(field, "// +kubebuilder:validation:Required")
+								removeOmitemptyFromJSONTag(field)
+								// TODO: remove pointer?
```
dgn, line 251: I'd say no. IMO if it somehow gets past validation it's still better to have it as `nil` then an accidental default value

dgn (approved):
Thank you very much Marko, this was clearly a lot of work but the result looks fantastic.

I noticed a lot of `TODO:` docstrings in the upstream definitions which now get copied into our CRD. Do you think we'd want to filter them out?

PR #1664, opened 2024-03-03
Fix printing of operator logs

dgn:
Little typo in https://github.com/maistra/istio-operator/pull/1662 prevented it from actually printing operator logs


dgn:
Closing as fix is included in https://github.com/maistra/istio-operator/pull/1663

PR #1665, opened 2024-03-05
OSSM-5541: istio-operator Pod keeps waiting for leader lease

yxun:
This issue and workaround are described in OSSM-5541.
It was caused by the version of operator-sdk library we use in our operator.

The operator-sdk library leader selection method is waiting when the pod is not the leader.
However, in case a node stopped this causes a hanging of waiting.
reference: operator-sdk pkg/leader/leader.go Become  line 136

~I am adding a 5 minutes timeout case in our cmd/manager for that call.~

This PR replaces the leader election implementation with leader-with-lease option.  
reference: https://sdk.operatorframework.io/docs/building-operators/golang/advanced-topics/#leader-with-lease

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	"github.com/magiconair/properties"
@@ -26,6 +27,7 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -128,12 +130,48 @@ func main() {
 	}
 
 	ctx := context.Background()
-	// Become the leader before proceeding
-	err = leader.Become(ctx, "istio-operator-lock")
+	c1 := make(chan error, 1)
 
-	if err != nil {
-		log.Error(err, "")
-		os.Exit(1)
+	go func() {
+		// Become the leader before proceeding
+		c1 <- leader.Become(ctx, "istio-operator-lock")
+	}()
+
+	select {
+	case err := <-c1:
+		if err != nil {
+			log.Error(err, "")
+			os.Exit(1)
+		}
+		break
+
+	// OSSM-5541: Delete the existing configmap when leader selection timeout
+	case <-time.After(5 * time.Minute):
+		client, err := crclient.New(cfg, crclient.Options{})
+		if err != nil {
+			log.Error(err, "")
+		}
+
+		// Delete the existing lock from this pod, in case the node stopped
+		existing := &v1.ConfigMap{}
+		key := crclient.ObjectKey{Namespace: namespace, Name: "istio-operator-lock"}
+		err = client.Get(ctx, key, existing)
+		if err != nil {
+			log.Info("No pre-existing configmap istio-operator-lock was found.")
+		} else {
+			err = client.Delete(ctx, existing)
+			if err != nil {
+				log.Error(err, "Unknown error trying to delete existing ConfigMap.")
+			}
+		}
+
+		// Block until receiving a notification from leader.Become
+		err = <-c1
+
+		if err != nil {
+			log.Error(err, "")
+			os.Exit(1)
+		}
 	}
 
 	// Set default manager options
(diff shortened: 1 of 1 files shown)


yxun:
It's hard to mock a node failure and go routine cases. Here is a code snippet , I use to test my logic: 

```
func main() {

	c1 := make(chan string, 1)
	go func() {

		c1 <- f1()
	}()

	select {
	case res := <-c1:
		fmt.Println(res)
	case <-time.After(1 * time.Second):
		fmt.Println("timeout 1")
		res := <-c1
		fmt.Println(res)
	}

}

func f1() string {
	time.Sleep(2 * time.Second)
	return "result 1"
}
```

luksa:
I don't think this is a good solution. Imagine that an operator is running and holds the lock and then another operator is started. The second operator won't be able to acquire the lock; however, after 5 minutes, it deletes the ConfigMap and recreates it and thus becomes the leader. Because the first operator is still running at this point, there are now two active operators. @yxun have you tested this scenario?

Since we can't update operator-sdk, we should probably use a different library to handle leader election, preferably one that uses the `Lease` resource instead of `ConfigMaps`.

yxun:
> I don't think this is a good solution. Imagine that an operator is running and holds the lock and then another operator is started. The second operator won't be able to acquire the lock; however, after 5 minutes, it deletes the ConfigMap and recreates it and thus becomes the leader. Because the first operator is still running at this point, there are now two active operators. @yxun have you tested this scenario?
> 
> Since we can't update operator-sdk, we should probably use a different library to handle leader election, preferably one that uses the `Lease` resource instead of `ConfigMaps`.

Hi @luksa , what's a correct way of testing two active operators ?
Should I run maistra-assist with my local branch image and then create two operators in two different ns ?
Should I create one operator subscription and then modify the replica set to 2 deployment ?

luksa:
Why do you need two active operators? If you're testing leader election, then you just need to scale the deployment to 2 replicas and then check if only one instance of the operator is active, while the other one waits until the first one terminates.

yxun:
I tested leader election works fine with this change. image:
quay.io/maistra-dev/istio-ubi8-operator:PR-1665

This PR is ready for review

luksa (approved):
LGTM, but I have one small concern. Previously, when you ran the operator locally, leader election was automatically turned off. This is typically what you want, because you usually run the operator in debug mode, with breakpoints and everything. 

Now, if you try to run the operator locally, it will throw an error. This means, you'll have to disable leader election by setting the `manager.Options.LeaderElection` field to `false`. Ideally, this should be done automatically or at least configurable via an environment variable or command-line option.

cmd/manager/main.go
```diff
@@ -75,6 +75,9 @@ func main() {
 	var logAPIRequests bool
 	pflag.BoolVar(&logAPIRequests, "logAPIRequests", false, "Log API requests performed by the operator.")
 
+	var runLocalBuild bool
+	pflag.BoolVar(&runLocalBuild, "runLocalBuild", false, "Run the operator locally.")
```
luksa: Instead of calling the flag `runLocalBuild`, I suggest calling it `leader-elect`, since that's the standard name used across different controller managers (including the core [kubernetes controller manager](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/)).

cmd/manager/main.go
```diff
@@ -75,8 +75,8 @@ func main() {
 	var logAPIRequests bool
 	pflag.BoolVar(&logAPIRequests, "logAPIRequests", false, "Log API requests performed by the operator.")
 
-	var runLocalBuild bool
-	pflag.BoolVar(&runLocalBuild, "runLocalBuild", false, "Run the operator locally.")
+	var leaderElect bool
+	pflag.BoolVar(&leaderElect, "leader-elect", true, "Enable a leader election option in the operator manager.")
```
luksa:
```suggestion
	pflag.BoolVar(&leaderElect, "leader-elect", true, "Enable leader election for this operator. Enabling this will ensure there is only one active controller manager.")
```

cmd/manager/main.go
```diff
@@ -128,20 +130,15 @@ func main() {
...
-		os.Exit(1)
-	}
 
 	// Set default manager options
 	options := manager.Options{
 		Namespace:              namespace,
 		Port:                   admissionControllerPort,
 		MetricsBindAddress:     net.JoinHostPort(metricsHost, fmt.Sprint(metricsPort)),
 		HealthProbeBindAddress: healthProbeBindAddress,
+		LeaderElection:         true,
```
luksa:
```suggestion
		LeaderElection:         leaderElect,
```

cmd/manager/main.go
```diff
@@ -153,6 +150,11 @@ func main() {
 		options.NewCache = cache.MultiNamespacedCacheBuilder(strings.Split(namespace, ","))
 	}
 
+	// When you run the operator locally in debug mode, turn off the leader election.
+	if !leaderElect {
+		options.LeaderElection = false
+	}
+
```
luksa:
```suggestion
```

yxun:
related maistra-assist pr:
https://gitlab.cee.redhat.com/istio/maistra-assist/-/merge_requests/14

PR #1685, opened 2024-03-12
OSSM-6088 Leader election improvements

luksa:


cmd/main.go
```diff
@@ -63,13 +63,15 @@ func main() {
...
 	var printVersion bool
+	var leaderElectionEnabled bool
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&configFile, "config-file", "/etc/sail-operator/config.properties", "Location of the config file, propagated by k8s downward APIs")
 	flag.StringVar(&resourceDirectory, "resource-directory", "/var/lib/sail-operator/resources", "Where to find resources (e.g. charts)")
 	flag.StringVar(&defaultProfiles, "default-profiles", "default", "One or more comma-separated profile names that are always applied to each Istio resource")
 	flag.BoolVar(&logAPIRequests, "log-api-requests", false, "Whether to log each request sent to the Kubernetes API server")
 	flag.BoolVar(&printVersion, "version", printVersion, "Prints version information and exits")
+	flag.BoolVar(&leaderElectionEnabled, "leader-elect", true, "Enable leader election for this operator. Enabling this will ensure there is only one active controller manager.")
```
jwendell: Do we really need a flag? If it's useful only for debugging purposes, does it make sense to have it covered by an env variable?

Should we explain in the arg description above when it makes sense to disable this feature?

luksa: Other controller managers (including the kube-controller-manager) expose this as a flag, so this makes our operator consistent with them.

luksa: BTW you could argue the same for the other flags (e.g. `config-file`, `resource-directory`, etc.). Those are mostly there so that you can run the operator locally.

jwendell: Sure thing. I just wanted to confirm if that's what we want, long term, i.e., bloat our operator with tons of command line arguments that are barely used.

I don't have a strong opinion though, just a weak one :)

jwendell (approved):

cmd/main.go
luksa: Yeah, me neither. But I don't think we'll have many flags.

PR #1692, opened 2024-03-14
OSSM-6110 Check istiod version in e2e test

luksa:


fjglira (approved):
lgtm

versions.yaml
```diff
@@ -28,6 +30,7 @@ versions:
       - https://istio-release.storage.googleapis.com/charts/cni-1.20.3.tgz
       - https://istio-release.storage.googleapis.com/charts/ztunnel-1.20.3.tgz
   - name: latest
+    version: "1.22"
```
jwendell: `1.22-alpha` maybe?

Makefile.core.mk
```diff
@@ -276,7 +276,7 @@ deploy-olm: bundle bundle-build bundle-push ## Build and push the operator OLM b
 
 .PHONY: undeploy
 undeploy: ## Undeploy controller from an existing cluster. Call with ignore-not-found=true to ignore resource not found errors during deletion.
-	kubectl delete istios.operator.istio.io --all --all-namespaces --wait=true
+	kubectl delete istios.operator.istio.io --all --all-namespaces --wait=true || echo "istios.operator.istio.io not found"
```
jwendell: is this supposed to be an error message, warning or just info? 
IOW, would the user (me) be wondering if the make undeploy failed?

should we just `2>/dev/null || true` to make it fail silently?

tests/integration/operator/operator_test.go
```diff
@@ -253,6 +255,22 @@ spec:
 	})
 })
 
+func getVersionFromIstiodLogs() (string, error) {
```
jwendell: can't we just run something like `kubectl -n istio-system exec istiod-76569d4f64-st6fs -- pilot-discovery version` and parse the output?

parsing the logs seems fragile and expensive to me...

versions.yaml
luksa: No, unfortunately the actual istiod version is `1.22-alpha.44056bc33df10623819256d847848c27066b3279-44056bc33df10623819256d847848c27066b3279-Clean` and obviously changes every time the image is built. But we only update the versions.yaml file once a day. This means that the tests would fail if we used the full version here.

The stable releases report the version as `1.20.3-692e556046b48ebc471205211c68a2c69e74a321-Clean`. The two formats are inconsistent and it's not trivial to extract the important part of the version, so I decided to just use "1.22" as the version here. 

But yeah, I don't like this either. I'll see if I can come up with a regexp that extracts what we want in both cases.

jwendell: Sorry, I meant alpha.

tests/integration/operator/operator_test.go
jwendell: By using it we could even parse the output as json:

```
$ docker run --rm -it  istio/pilot:1.20.4 version
version.BuildInfo{Version:"1.20.4", GitRevision:"ff333a3512ba8321f40d90b6fc041db04fcb280b", GolangVersion:"go1.21.8", BuildStatus:"Clean", GitTag:"1.20.4"}
```

Just inspect the `.Version` field.

jwendell: perhaps you can even import this file to use this struct directly: https://github.com/istio/istio/blob/master/pkg/version/version.go#L37

luksa: Yeah, you're right. I don't know what I was thinking parsing the logs.

luksa: But we still have to use a regexp to extract the stable part of the version.

Makefile.core.mk
luksa: Nevermind. I'll remove this from this PR, because other commands in this target also fail in some scenarios.

tests/integration/operator/operator_test.go
```diff
@@ -253,6 +255,19 @@ spec:
...
 })
 
+func getVersionFromIstiod() (string, error) {
+	output, err := kubectl.Exec(controlPlaneNamespace, "deploy/istiod", "pilot-discovery version")
+	if err != nil {
+		return "", fmt.Errorf("error getting version from istiod: %v", err)
+	}
+
+	matches := istiodVersionRegex.FindStringSubmatch(output)
+	if len(matches) > 0 && matches[1] != "" {
```
jwendell: will this work (ie, not crash with array out of bounds) if len==1?
IOW, is it better to have `if len(matches) > 1 ...`?

luksa: It wouldn't crash because FindStringSubmatch would either return nil or a slice with exactly 2 elements. But I've fixed the condition anyway.

jwendell (approved):

PR #1695, opened 2024-03-15
Test improvements

luksa:


fjglira (approved):

common/scripts/kind_provisioner.sh
```diff
@@ -196,7 +196,7 @@ EOF
     return 9
   fi
   # Workaround kind issue causing taints to not be removed in 1.24
-  kubectl taint nodes "${NAME}"-control-plane node-role.kubernetes.io/control-plane- || true
+  kubectl taint nodes "${NAME}"-control-plane node-role.kubernetes.io/control-plane- 2>/dev/null || true
```
fjglira: This file is the same one that the upstream test has, maybe we should refer directly to that file and make the change there?

luksa: Argh. Thanks.

PR #1696, opened 2024-03-15
Ensure automatic update of version in Istio CR samples

luksa:


hack/update-version-list.sh
```diff
@@ -66,5 +66,14 @@ function updateVersionsInCSVDescription() {
     rm "$tmpFile"
 }
 
+function updateVersionInSamples() {
+    defaultVersion=$(yq '.versions[0].name' "${VERSIONS_YAML_FILE}")
```
jwendell, line 70: This assumes the first entry is the default. Is this enforced in any way? Maybe adding a note in versions.yaml to state this while we don't have a not-so-fragile way of enforcing it?

luksa: Done.

jwendell (approved):

PR #1700, opened 2024-03-18
Refactor helm

luksa:


pkg/helm/charts.go
```diff
@@ -17,55 +17,30 @@ package helm
...
-	}
-	for _, chartName := range charts {
-		releaseName := fmt.Sprintf("%s-%s", releaseNameBase, chartName)
-		_, err = uninstallChart(actionConfig, ns, releaseName)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
+type Client struct {
```
jwendell: does it make sense to rename this file to `client.go`? or move helmClient stuff to its own file?

pkg/helm/charts.go
```diff
@@ -17,55 +17,30 @@ package helm
...
-	if err != nil {
-		return err
-	}
-	for _, chartName := range charts {
-		releaseName := fmt.Sprintf("%s-%s", releaseNameBase, chartName)
-		_, err = upgradeOrInstallChart(ctx, actionConfig, chartName, chartVersion, ns, releaseName, ownerReference, values)
-		if err != nil {
-			return err
-		}
+func NewClient(cfg *rest.Config, driver string) *Client {
```
jwendell: could you add comments explaining what the arguments are?

pkg/helm/charts.go
```diff
@@ -159,3 +128,18 @@ func uninstallChart(cfg *action.Configuration, namespace, releaseName string) (*
 
 	return response, nil
 }
+
+func (h *Client) releaseExists(cfg *action.Configuration, namespace string, name string) (bool, error) {
```
jwendell: does it make sense to move this out of the Client class? This function doesn't use `h` at all...

pkg/helm/charts.go
luksa: Renamed to chartmanager.go, since I've also renamed the type to ChartManager (the name "Client" wasn't entirely correct).

pkg/helm/charts.go
luksa: Done.

pkg/helm/charts.go
luksa: Done.

pkg/helm/chartmanager.go
```diff
@@ -17,55 +17,34 @@ package helm
...
-		releaseName := fmt.Sprintf("%s-%s", releaseNameBase, chartName)
-		_, err = upgradeOrInstallChart(ctx, actionConfig, chartName, chartVersion, ns, releaseName, ownerReference, values)
-		if err != nil {
-			return err
-		}
+// NewChartManager creates a new Helm chart manager using cfg as the configuration
+// that Helm will use to connect to the cluster when installing or uninstalling
+// charts, and using the specified driver to store information about releases
+// (one of: memory, secret, configmap, sql).
+func NewChartManager(cfg *rest.Config, driver string) *ChartManager {
```
jwendell, line 39: I see usages with driver being empty string. What does it mean?

jwendell (approved):

pkg/helm/chartmanager.go
luksa: Default (secret). The reason I don't really want to document this further is that Helm itself doesn't have any documentation on this.

PR #1707, opened 2024-03-20
Simplify scheme usage by centralizing the scheme

luksa:


pkg/scheme/scheme.go
```diff
@@ -0,0 +1,36 @@
...
+
+var Scheme = runtime.NewScheme()
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(Scheme))
+	utilruntime.Must(multusv1.AddToScheme(Scheme))
+	utilruntime.Must(networkingv1alpha3.AddToScheme(Scheme))
+
+	utilruntime.Must(v1alpha1.AddToScheme(Scheme))
+	// +kubebuilder:scaffold:scheme
```
jwendell, line 35: do we need this?

jwendell (approved):

pkg/scheme/scheme.go
luksa: Not sure. This was in the original file, but we don't really use kubebuilder to add new CRDs (at least I prefer to add them manually).

PR #1710, opened 2024-03-20
OSSM-6126 Handle conflict when removing finalizer

luksa:


jwendell (commented):
this pr has the changes from the other pr (scheme)?

luksa:
@jwendell oh yeah, I forgot to remove them. I'll wait for the other PR to be merged and then I'll remove them from here.

pkg/kube/finalizers.go
```diff
@@ -44,12 +46,17 @@ func RemoveFinalizer(ctx context.Context, obj client.Object, cl client.Client) e
 
 	err := cl.Update(ctx, obj)
 	if errors.IsNotFound(err) {
-		// We're reconciling a stale instance. The object no longer exists, so we're done.
-		return nil
+		log.Info("Resource no longer exists; nothing to do")
```
jwendell, line 49: I just wonder if these logs should be reduced to debug, if they are too verbose...

jwendell (approved):

pkg/kube/finalizers.go
luksa: Yeah, I had the same thought and even started changing everything to `log.V(2)`, but then I realized that when I add a "Starting reconciliation" log message (which I plan to do in another PR soon), then the reconciliation that only adds or removes the finalizer, won't log anything, which would be confusing. So I now think that these finalizer-related log statements should remain at this Info level so that the logs will make sense (e.g. "Starting reconciliation", "Removing finalizer", "Resource no longer exists; nothing to do").

I think this is the only correct thing to do. All other options would result in strange (or no) logs being produced for reconciles that only add or remove the finalizer.

luksa: But maybe we could remove the "Removing finalizer" statement and just leave the others (e.g. "Couldn't remove due to error", "Conflict", "Nothing to do", and "Finalizer removed").

PR #1726, opened 2024-03-22
Run integration tests with ginkgo instead of go test and ensure logs go to GinkgoWriter instead of stdout

luksa:
Ginkgo displays progress, whereas go test doesn't.

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -152,7 +152,7 @@ test.unit:  ## Run unit tests.
 .PHONY: test.integration
 test.integration: envtest ## Run integration tests located in the tests/integration directory.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" \
-	go test --tags=integration $(VERBOSE_FLAG) ./tests/integration/...
+	go run github.com/onsi/ginkgo/v2/ginkgo --tags=integration $(VERBOSE_FLAG) ./tests/integration/...
 
 .PHONY: test.scorecard
 test.scorecard: operator-sdk ## Run the operator scorecard test.
--- a/pkg/test/environment.go
+++ b/pkg/test/environment.go
@@ -15,6 +15,7 @@
 package test
 
 import (
+	"io"
 	"path"
 
 	"github.com/istio-ecosystem/sail-operator/pkg/common"
@@ -26,8 +27,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
-func SetupEnv() (*envtest.Environment, client.Client, *rest.Config) {
-	logf.SetLogger(zap.New(zap.UseDevMode(true)))
+func SetupEnv(logWriter io.Writer) (*envtest.Environment, client.Client, *rest.Config) {
+	logf.SetLogger(zap.New(zap.WriteTo(logWriter), zap.UseDevMode(true)))
 
 	testEnv := &envtest.Environment{
 		CRDDirectoryPaths:     []string{path.Join(common.RepositoryRoot, "chart", "crds")},
--- a/tests/integration/api/suite_test.go
+++ b/tests/integration/api/suite_test.go
@@ -36,8 +36,6 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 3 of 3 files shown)


Makefile.core.mk
```diff
@@ -152,7 +152,7 @@ test.unit:  ## Run unit tests.
 .PHONY: test.integration
 test.integration: envtest ## Run integration tests located in the tests/integration directory.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" \
-	go test --tags=integration $(VERBOSE_FLAG) ./tests/integration/...
+	go run github.com/onsi/ginkgo/v2/ginkgo --tags=integration $(VERBOSE_FLAG) ./tests/integration/...
```
jwendell, line 155: just tried this locally and the output is too much verbose. :(

also, does this run ginkgo from master? This might break us at some point? Should we stick to a released version instead?

luksa: Yup, too much output due to a bug: https://issues.redhat.com/browse/OSSM-6159

luksa: As for running ginkgo from master. We use the same approach in the e2e tests. So if this need to be fixed, we should do it in both places (in a follow up PR).

luksa: Actually, it looks like `go run` honors whatever is in go.mod, so we're actually running v2.15.0:
```
[luksa@p1 sail-operator]$ go run github.com/onsi/ginkgo/v2/ginkgo version
Ginkgo Version 2.15.0

[luksa@p1 sail-operator]$ cd ..

[luksa@p1 maistra]$ go run github.com/onsi/ginkgo/v2/ginkgo version
no required module provides package github.com/onsi/ginkgo/v2/ginkgo: go.mod file not found in current directory or any parent directory; see 'go help modules'
```

luksa: @jwendell I've just pushed another commit that fixes the logs.

jwendell: Much better now. Now I see green dots.

Is it a matter of preference? Because I prefer 1000x the output of `go test`. It's standardized across different projects :)

luksa: For GInkgo tests, there's only one go test, so if you use `go test`, you don't see anything until the test completes. With ginkgo, you at least see the dots.

jwendell: Oh man, this is so anti pattern :/

Now when we run `make test` we have a mix of go and ginkgo outputs 😩

luksa: Why anti-pattern? Those are two very different sets of tests. Maybe the anti-pattern is that we run them in a single make target?

luksa: But that's not really a good argument. Other make targets also have multiple outputs (e.g. docker push + go test output).

jwendell: Sorry, when I said anti-pattern, I was referring to "For GInkgo tests, there's only one go test". I was expecting it to integrate well with the go test subsystem so all ginkgo tests would appear as go tests when run with `go test`...

luksa: Yeah, I agree, but as you say: "it is what it is" :)

Ginkgo has "specs" not tests. The list of specs gets built from a hierarchy of test "nodes". These specs are then executed (in a single test function that you must write). Unlike functions that are organized into packages, the specs are just one giant sequential list. And they are all executed by your test function.

jwendell: I see. Is there a way to show the progress a little more verbose, instead of just showing small dots?

luksa: I don't think so. There's only `-v`, but this displays everything that is logged by the test. I don't see any other option that would e.g. only print the specs with nothing else.

jwendell: ok. It is what it is, should we revert to using plain old go tests then? /me runs :D

pkg/test/environment.go
```diff
@@ -26,8 +27,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
-func SetupEnv() (*envtest.Environment, client.Client, *rest.Config) {
-	logf.SetLogger(zap.New(zap.UseDevMode(true)))
+func SetupEnv(logWriter io.Writer) (*envtest.Environment, client.Client, *rest.Config) {
+	logf.SetLogger(zap.New(zap.WriteTo(logWriter), zap.UseDevMode(true)))
```
jwendell, line 31: Funny story, in Brazil, "zap" is a short for "Whatsapp", the messenger app.

jwendell (approved):

Makefile.core.mk
luksa: We wouldn't be reverting. We had ginkgo tests from the very start. :)

PR #1727, opened 2024-03-22
OSSM-4840 Ensure chart is rendered even when existing helm release is in a weird state

luksa:


dgn (approved):
LGTM, thanks for the fix! We really need to add instrumentation to the operator so we can gauge how often these sorts of things happen.

PR #1750, opened 2024-03-28
Run all ginkgo tests with --no-color when running in CI

luksa:


luksa:
/override ci/prow/e2e-ocp-4-15-arm ci/prow/e2e-ocp-4-15 ci/prow/e2e-ocp-4-16

Makefile.core.mk
```diff
@@ -72,6 +72,7 @@ ENVTEST_K8S_VERSION ?= 1.29.0
 DOCKER_BUILD_FLAGS ?= "--platform=$(TARGET_OS)/$(TARGET_ARCH)"
 
 VERBOSE_FLAG := $(if $(VERBOSE),-v)
+NOCOLOR_FLAG := $(if [ -t 1 ],--no-color,)
```
jwendell: I'd rather join these two into a single flag, sth like `TEST_FLAGS`

jwendell (approved):

Makefile.core.mk
luksa: Can't. `go test` doesn't have `--no-color`

luksa: We could have two vars: `GOTEST_FLAGS` and `GINKGO_FLAGS`, but would that really be better than what we have now?

dgn: I'd personally prefer `GOTEST_FLAGS` and `GINKGO_FLAGS` but it's not important

dgn (approved):

Makefile.core.mk
luksa: Argh. I tested this and remember seeing it work correctly both attached and not attached, but when I tested it again, it doesn't behave correctly.

It turns out that when you run something in `$(cmd)`, it's never attached to a terminal. I can't get it to work, even when I add `$$( [ -t 1 ] ...)` to where we invoke the ginkgo command. The only two solutions that work are:
- have the make target call a shell script, which can then check whether it's attached
- use the CI env var instead of checking whether the terminal is attached

I don't like either solution. @dgn @jwendell your thoughts?

jwendell: check the `CI` env var.

We can always come up with better solutions when they are needed. Let's just fix this issue and move on

jwendell:
/override ci/prow/e2e-ocp-4-15-arm ci/prow/scorecard ci/prow/e2e-ocp-4-15

PR #1751, opened 2024-03-28
Small adjustements to the publish bundle script

jwendell:
In particular, do not force push to the fork and make curl fail when it receives an HTTP response code that indicates a failure (eg 404)

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -240,7 +240,7 @@ undefine BUILDX
 endif
 
 .PHONY: docker-buildx
-docker-buildx: test build-all ## Build and push docker image with cross-platform support.
+docker-buildx: build-all ## Build and push docker image with cross-platform support.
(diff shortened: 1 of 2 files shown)


Makefile.core.mk
```diff
@@ -240,7 +240,7 @@ undefine BUILDX
 endif
 
 .PHONY: docker-buildx
-docker-buildx: test build-all ## Build and push docker image with cross-platform support.
+docker-buildx: build-all ## Build and push docker image with cross-platform support.
```
bmangoen, line 243: Should we test somewhere else? Before publishing the bundle maybe?

jwendell: all merged code is already tested. why test again?

bmangoen (approved):

jwendell:
/override ci/prow/push-images ci/prow/e2e-kind ci/prow/e2e-ocp-4-15-arm

PR #1755, opened 2024-03-29
OSSM-6189 Update status even when status can't be determined

luksa:


controllers/istio/istio_controller.go
```diff
@@ -460,15 +461,15 @@ func (r *IstioReconciler) determineStatus(ctx context.Context, istio *v1alpha1.I
...
-	if err != nil {
-		return err
-	}
+	errs.Add(err)
 
-	if reflect.DeepEqual(istio.Status, *status) {
-		return nil
+	if !reflect.DeepEqual(istio.Status, *status) {
+		err := r.Client.Status().Patch(ctx, istio, kube.NewStatusPatch(*status))
+		errs.Add(err)
```
jwendell: nit: maybe `errs.Add(r.Client.....)`

jwendell (approved):

controllers/istio/istio_controller.go
luksa: Yeah, I tried it, but it didn't look right. I thought it made it harder to see what was actually happening. But now I've tried it again and it seems fine if wrapped in `errs.Add()`.

Done.

PR #1767, opened 2024-04-17
OSSM-6267 Fix Grafana 9.2 datasource query error

yxun:
The first commit is fixing a missing mount when deploying Grafana 9.2. Alerting is a new feature introduced in Grafana 9.
The second commit is using an additional custom header when configuring Prometheus data source in Grafana.
The previous basic auth config is not fully working in Grafana 9.2.10
   It works in data source metrics data and it is not working when Grafana sends a query request.
 So I replace the basic auth configmap field with a custom authorization header in data source.

### DIFF
--- a/pkg/controller/servicemesh/controlplane/htpasswd.go
+++ b/pkg/controller/servicemesh/controlplane/htpasswd.go
@@ -17,6 +17,8 @@ import (
 	"github.com/maistra/istio-operator/pkg/controller/versions"
 )
 
+var username = "internal"
+
 func (r *controlPlaneInstanceReconciler) patchHtpasswdSecret(ctx context.Context, object *unstructured.Unstructured) error {
 	var rawPassword, auth string
 	log := common.LogFromContext(ctx)
@@ -51,6 +53,7 @@ func (r *controlPlaneInstanceReconciler) patchHtpasswdSecret(ctx context.Context
 	b64Password := base64.StdEncoding.EncodeToString([]byte(rawPassword))
 	b64Auth := base64.StdEncoding.EncodeToString([]byte(auth))
 
+	// b/c Keep rawPassword for previous release Kiali, Prometheus and Grafana patching
 	// We store the raw password in order to be able to retrieve it below, when patching Grafana ConfigMap
 	err = unstructured.SetNestedField(object.UnstructuredContent(), b64Password, "data", "rawPassword")
 	if err != nil {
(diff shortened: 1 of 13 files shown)


yxun:
Ready for review. Need to merge #1772 and #1771 first and then push another commit about 2.3 and 2.4 helm changes

pkg/controller/servicemesh/controlplane/htpasswd.go
```diff
@@ -111,13 +136,14 @@ func (r *controlPlaneInstanceReconciler) patchGrafanaConfig(ctx context.Context,
...
+	authToken, err := r.getHtauthToken(ctx)
 	if err != nil {
 		return err
 	}
+	b64authToken := base64.StdEncoding.EncodeToString([]byte(authToken))
 
-	re := regexp.MustCompile("(?s)(basicAuthPassword:).*?\n")
-	dsYaml = re.ReplaceAllString(dsYaml, fmt.Sprintf("${1} %s\n", rawPassword))
+	re := regexp.MustCompile("(?s)({AuthToken}).*?\n")
+	dsYaml = re.ReplaceAllString(dsYaml, fmt.Sprintf("Basic %s'\n", b64authToken))
```
jewertow, line 143: Could we utilize `unstructured` API to set `secureJsonData.httpHeaderValue1` instead of this regex?

pkg/controller/servicemesh/controlplane/htpasswd.go
```diff
@@ -51,6 +53,7 @@ func (r *controlPlaneInstanceReconciler) patchHtpasswdSecret(ctx context.Context
 	b64Password := base64.StdEncoding.EncodeToString([]byte(rawPassword))
 	b64Auth := base64.StdEncoding.EncodeToString([]byte(auth))
 
+	// b/c Keep rawPassword for previous release Kiali, Prometheus and Grafana patching
```
jewertow: Could you somehow append this information to the existing comment so that they are one coherent piece of information? And what "b/c" means?

pkg/controller/servicemesh/controlplane/htpasswd.go
```diff
@@ -64,12 +67,21 @@ func (r *controlPlaneInstanceReconciler) patchHtpasswdSecret(ctx context.Context
 		return err
 	}
 
+	// OSSM-6267 Grafana 9.2: We store the username:rawPassword in order to be able to retrieve it below,
+	// when patching Grafana ConfigMap
```
jewertow: Could we not prepend comments with "OSSM-*"? I think that comments should be self-explanatory.
```suggestion
	// We store the username:rawPassword in order to be able to retrieve it below,
	// when patching ConfigMap for Grafana 9.2+
```

pkg/controller/servicemesh/controlplane/htpasswd.go
yxun: I googled an [Abbreviation](https://en.wiktionary.org/wiki/Appendix:Glossary#abbreviation) of [backward compatibility](https://en.wiktionary.org/wiki/backward_compatibility#English).

I will organize the comment.

pkg/controller/servicemesh/controlplane/htpasswd.go
yxun: Hi Jacek, I don't think the `SetNestedField` method can find that line directly. The whole file datasources.yaml is listed in the ConfigMap below:
We can only set nested level to the .data.datasources.yaml and the configurations are in the same field's value. We may set all the contents of the datasources.yaml using the `unstructured` API method but that approach is not as clear as having the datasources.yaml template in the values.yaml.

```
data:
  dashboardproviders.yaml: |-
    apiVersion: 1
    providers:
    - disableDeletion: false
      folder: istio
      name: istio
      options:
        path: /var/lib/grafana/dashboards/istio
      orgId: 1
      type: file
  datasources.yaml: |
    apiVersion: 1
    datasources:
    - access: proxy
      basicAuth: true
      basicAuthUser: 
      editable: true
      isDefault: true
      jsonData:
        timeInterval: 5s
        tlsSkipVerify: true
      name: Prometheus
      orgId: 1
      secureJsonData:
        basicAuthPassword:
       ```

jewertow (approved):

PR #1777, opened 2024-05-01
OSSM-6397: Revert OSSM-5541: istio-operator leader election change

yxun:
…(#1665)"

This reverts commit d387067ef47d27dc950bc5139ff4eca083fec537.

The Leader with lease implementation is only applicable in "sigs.k8s.io/controller-runtime v0.7.0 and above".
Upgrade that dependency from v0.6.3 needs major operator code refactoring. That's a not applicable in OSSM operator 2.5.x branch.
So I reverted the code change of #1665 in order to avoiding OSSM-6397 error.

Reference: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.7.0/pkg/manager

```
// Note: before controller-runtime version v0.7, the resource lock was set to "configmaps".
// Please keep this in mind, when planning a proper migration path for your controller.
LeaderElectionResourceLock [string](https://pkg.go.dev/builtin#string)
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/mitchellh/mapstructure"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	kubemetrics "github.com/operator-framework/operator-sdk/pkg/kube-metrics"
+	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 	"github.com/operator-framework/operator-sdk/pkg/metrics"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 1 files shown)


luksa:
@yxun I'm not sure I understand what you're trying to say. The comment that you reference from controller-runtime merely states that you should be careful when you migrate from using a `ConfigMap` to a `Lease` for leader election. Your change in https://github.com/maistra/istio-operator/commit/d387067ef47d27dc950bc5139ff4eca083fec537 doesn't do that. It keeps using a `ConfigMap` for the lock, so the comment doesn't apply in your case.

luksa:
I suggest closing this PR and potentially replacing it with a PR that allows users to configure them (or increase the defaults, but I'd like to avoid this because the defaults are already sufficiently high).

PR #1779, opened 2024-05-06
OSSM-4585 include must-gather image in csv

yxun:


dgn (approved):
lgtm- but do we need to keep separate versions? one global must-gather should be fine, right?

yxun:
> lgtm- but do we need to keep separate versions? one global must-gather should be fine, right?

I agree. Include only the latest one is sufficient.

PR #1780, opened 2024-05-07
OSSM-5556: append control plane namespace to discovery selectors

jewertow:
**Background context:**
When discovery selectors are configured, then must include the control plane namespace. Otherwise, gateways in the control plane namespace with unprivileged port, i.e. the default ingress and egress gateways, will not receive `Gateway` configurations.

Previously, I tried to change `gateways` charts to not use unprivileged ports, but that change broken many tests, so that issue still exists and users must be aware of that and properly configure `discoverySelectors`.

**Proposed solution:**
The operator can append the control plane namespace to `meshConfig.discoverySelectors` if this field is configured.

### DIFF
--- a/pkg/controller/servicemesh/controlplane/reconciler.go
+++ b/pkg/controller/servicemesh/controlplane/reconciler.go
@@ -185,6 +185,10 @@ func (r *controlPlaneInstanceReconciler) Reconcile(ctx context.Context) (result 
 			return
 		}
 
+		if r.Instance.Spec.Mode == v2.ClusterWideMode {
+			overrideDiscoverySelectors(&r.Instance.Spec, r.Instance.Namespace)
+		}
+
 		// initialize new Status
 		componentStatuses := make([]status.ComponentStatus, 0, len(r.Status.ComponentStatus))
 		for _, charts := range r.getChartsInInstallationOrder(version.Strategy().GetChartInstallOrder()) {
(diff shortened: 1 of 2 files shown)


pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -597,3 +601,43 @@ func componentFromChartName(chartName string) string {
...
+func overrideDiscoverySelectors(spec *v2.ControlPlaneSpec, controlPlaneNamespace string) {
+	var discoverySelectors []*metav1.LabelSelector
+	if spec.MeshConfig != nil {
+		discoverySelectors = spec.MeshConfig.DiscoverySelectors
+	}
+	if discoverySelectors == nil {
+		return
+	}
+	for _, s := range discoverySelectors {
+		if s.MatchExpressions != nil {
```
luksa, line 614: What about `s.MatchLabels`? Don't we need to check those as well?

pkg/controller/servicemesh/controlplane/reconciler.go
```diff
@@ -597,3 +601,43 @@ func componentFromChartName(chartName string) string {
...
+	}
+	discoverySelectors = append(discoverySelectors, &metav1.LabelSelector{
+		MatchExpressions: []metav1.LabelSelectorRequirement{
+			{
+				Key:      "kubernetes.io/metadata.name",
+				Operator: metav1.LabelSelectorOpIn,
+				Values:   []string{controlPlaneNamespace},
+			},
+		},
+	})
```
luksa, line 632: Hmm, aren't discovery selectors ANDed together (and not ORed)? This means that by adding this selector, discovery will _only_ be performed in the control plane namespace.

Also, what if there's an existing `kubernetes.io/metadata.name` selector? In that case, you have to append to it.

luksa: Ah, nevermind. We have an array of selectors that are ORed, but each selector contains an array of requirements that are ANDed together, so your solution should be okay.

pkg/controller/servicemesh/controlplane/reconciler.go
jewertow: We would have to know namespace labels - this is easy during reconciliation, but what about removing relevant label from the namespace after reconciliation? Then we would have to watch namespace and add proper `matchExpressions` on a change. That's too complicated, so I think we can just add single `matchExpressions` rule for the namespace name. If `matchLabels` matches the control plane namespace and `matchExpressions` as well - that's fine. Rules won't be optimal, but still correct.

luksa (approved):

dgn (commented):
I'm a little torn on this. I think this is a great solution to 'copy' multitenancy behavior (ie control plane namespace is always included by default) but it is a deviation from upstream behavior. I honestly think proper documentation is better than patching the discoverySelectors, because in the long run (3.0), we will converge with upstream behaviour anyway

jewertow:
@dgn I do not agree with "I think this is a great solution to 'copy' multitenancy behavior", because this is an OpenShift-related bug, not multi-tenant behavior. For some reason proxies with UNPRIVILEGED_PORT setting are incorrectly discovered on OpenShift when `meshConfig.discoverySelectors` are configured. I was able to reproduce this bug with community Istio on OpenShift, but not on KinD. I didn't find the root cause yet. This is in my backlog, but it's very complicated and I will need more time.

> I honestly think proper documentation is better than patching the `discoverySelectors`

Why to bother users and requires them them to know all hidden issues and teach workarounds, instead of automating and patching problematic configs?

> because in the long run (3.0), we will converge with upstream behaviour anyway

I'm going to investigate the root cause more and I hope I will be able to fix this problem directly in istio/istio, but in the long term I wanted to fix this issue to ensure seamless migration between tenancy modes.

Does it convince you?

dgn:
I see. In that case, I think my preferred approach would be investigating the issue first before making this change. Maybe it's something we can fix in istio/istio. But if there's an urgent customer case behind this, feel free to merge as-is. I don't think there's a lot of risk attached to this change, it's just a behaviour change when compared with upstream so we'll need to document that.

jewertow:
I installed Istio built from latest master and I noticed that now (since 1.22) Istio does not create `istio-ca-root-cert` config map in namespaces which do not match discovery selectors. This means that now users have to include istio-system in the discovery selectors if they deploy gateways in the control plane namespace. So I'm closing this PR to keep maistra compliant with the latest istio.

PR #1788, opened 2024-05-16
[maistra-2.5] make gen

jewertow:


jwendell (approved):
I thought we had automation for this. Did it fail or did we forget to add this automator job for 2.5?

jewertow:
It didn't fail, so I guess we don't have this job

PR #1795, opened 2024-06-13
OSSM-6411: Copy appropriate istio-cni binary based on host OS [maistra-2.5]

eoinfennessy:


dgn:
We usually only support one version of the operator at a time - so when 2.6 comes out, that will be 2.6.0, for all control planes versions down to 2.4, so you should not have to cherry-pick these changes back.

eoinfennessy:
> We usually only support one version of the operator at a time - so when 2.6 comes out, that will be 2.6.0, for all control planes versions down to 2.4, so you should not have to cherry-pick these changes back.

@dgn Thanks Daniel. Should I close the PRs targeting `maistra-2.4` and `maistra-2.3`? I thought I needed to also update these so that changes made in those branches would be pulled in to the latest branch when running `make update-charts`.

dgn:
> I thought I needed to also update these so that changes made in those branches would be pulled in to the latest branch when running `make update-charts`.

In most cases that is true, however CNI is sort of a special case as we keep the daemonsets for all supported versions in the `overlays` directory which is always applied on top of the charts from the branches, so if you make these changes in the maistra-2.6 branch and apply to all relevant daemonsets, all versions will be covered.

dgn:
And for CNI we always install the latest version (which contains the daemonsets for the older versions) so you don't have to change the old CNI charts - the 2.6 operator will be the only one available once we release and it always installs the 2.6 CNI chart, backwards-compatibility being ensured through the separate daemonsets

dgn:
So you can close all these PRs and open a new one against maistra-2.6

eoinfennessy:
@dgn Thanks Daniel. Closed these three PRs and opened #1806 targeting `maistra-2.6`.

PR #1796, opened 2024-06-14
OSSM-6339: Bump moby pkg used by helm

unsortedhashsets:


luksa (approved):

unsortedhashsets:
Hi, @luksa
Thanks for the review.
I am going to merge it and check if the downstream build pass with that change

PR #1802, opened 2024-06-17
OSSM-6391 - 2.6 Disable the OSSM operator's Jaeger and ES installation by default

DeanDoyle1502:
[https://issues.redhat.com/browse/OSSM-6391](url)

### DIFF
--- a/build/manifest-templates/clusterserviceversion.yaml
+++ b/build/manifest-templates/clusterserviceversion.yaml
@@ -40,10 +40,6 @@ metadata:
           },
           "spec": {
             "version": "v2.6",
-            "tracing": {
-              "type": "Jaeger",
-              "sampling": 10000
-            },
             "policy": {
                "type": "Istiod"
             },
@@ -51,13 +47,6 @@ metadata:
                "type": "Istiod"
             },
             "addons": {
-              "jaeger": {
-                "install": {
-                  "storage": {
-                    "type": "__JAEGER_STORAGE__"
-                  }
-                }
-              },
               "prometheus": {
                 "enabled": true
               },
--- a/pkg/controller/servicemesh/webhooks/mutation/controlplane.go
+++ b/pkg/controller/servicemesh/webhooks/mutation/controlplane.go
@@ -334,5 +334,12 @@ func (m *smcpv2mutator) IsTracingTypeSpecified() bool {
 }
 
 func (m *smcpv2mutator) SetTracingType(value v2.TracerType) {
-	m.patches = append(m.patches, jsonpatch.NewPatch("add", "/spec/tracing/type", string(value)))
+	tracing := m.smcp.Spec.Tracing
+
+	if tracing == nil {
+		tracing = &v2.TracingConfig{}
+
+	}
+	tracing.Type = value
+	m.patches = append(m.patches, jsonpatch.NewPatch("add", "/spec/tracing", *tracing))
 }
--- a/pkg/controller/servicemesh/webhooks/mutation/controlplane.go
+++ b/pkg/controller/servicemesh/webhooks/mutation/controlplane.go
@@ -60,55 +60,34 @@ func (v *ControlPlaneMutator) Handle(ctx context.Context, req admission.Request)
 	// on create we set the version to the current default version
 	// on update, if the version is removed we reset it to what was previously set
 	currentVersion := mutator.NewVersion()
+	effectiveVersion, _ := versions.ParseVersion(currentVersion)
 	if currentVersion == "" {
 		switch req.AdmissionRequest.Operation {
 		case admissionv1beta1.Create:
 			log.Info("Setting .spec.version to default value", "version", versions.DefaultVersion.String())
 			mutator.SetVersion(mutator.DefaultVersion())
+			effectiveVersion, _ = versions.ParseVersion(mutator.DefaultVersion())
 		case admissionv1beta1.Update:
 			oldVersion := mutator.OldVersion()
 			if currentVersion != oldVersion && oldVersion != versions.InvalidVersion.String() {
 				log.Info("Setting .spec.version to existing value", "version", oldVersion)
 				mutator.SetVersion(oldVersion)
+				effectiveVersion, _ = versions.ParseVersion(oldVersion)
 			}
 		}
 	}
 
-	// As we are deprecating IOR, on creating a v2.5 SMCP we want to disable IOR if not specified explicitly
 	if req.AdmissionRequest.Operation == admissionv1beta1.Create {
+		// As we are deprecating IOR, on creating a v2.5 SMCP we want to disable IOR if not specified explicitly
 		newOpenShiftRoute := mutator.IsOpenShiftRouteEnabled()
 
-		if newOpenShiftRoute == nil {
-			var ver versions.Version
-			var err error
-			// If version is not specified
-			if currentVersion == "" {
-				ver, err = versions.ParseVersion(mutator.DefaultVersion())
-			} else {
-				ver, err = versions.ParseVersion(currentVersion)
-			}
-			if err == nil && ver.AtLeast(versions.V2_5.Version()) {
-				mutator.SetOpenShiftRouteEnabled(false)
-			}
+		if newOpenShiftRoute == nil && effectiveVersion.AtLeast(versions.V2_5.Version()) {
+			mutator.SetOpenShiftRouteEnabled(false)
 		}
-	}
 
-	// removing tracing by default on creation of v2.6 SMCP
-	if req.AdmissionRequest.Operation == admissionv1beta1.Create {
-		newOpenShiftTracing := mutator.IsTracingTypeSpecified()
-
-		if !newOpenShiftTracing {
-			var ver versions.Version
-			var err error
-			// If version is not specified
-			if currentVersion == "" {
-				ver, err = versions.ParseVersion(mutator.DefaultVersion())
-			} else {
-				ver, err = versions.ParseVersion(currentVersion)
-			}
-			if err == nil && ver.AtLeast(versions.V2_6.Version()) {
-				mutator.SetTracingType(v2.TracerTypeNone)
-			}
+		// removing tracing by default on creation of v2.6 SMCP
+		if !mutator.IsTracingTypeSpecified() && effectiveVersion.AtLeast(versions.V2_6.Version()) {
+			mutator.DisableTracing()
 		}
 	}
 
@@ -189,7 +168,7 @@ type smcpmutator interface {
 	IsOpenShiftRouteEnabled() *bool
 	SetOpenShiftRouteEnabled(bool)
 	IsTracingTypeSpecified() bool
-	SetTracingType(v2.TracerType)
+	DisableTracing()
 }
 
 type smcppatch struct {
@@ -325,7 +304,7 @@ func (m *smcpv1mutator) IsTracingTypeSpecified() bool {
 	return false
 }
 
-func (m *smcpv1mutator) SetTracingType(value v2.TracerType) {}
+func (m *smcpv1mutator) DisableTracing() {}
 
 func (m *smcpv2mutator) IsTracingTypeSpecified() bool {
 	tracing := m.smcp.Spec.Tracing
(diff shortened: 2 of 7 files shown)


luksa:
@DeanDoyle1502 are you still not a member of the maistra team? I'm sure I sent you the invite at least once.

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -93,6 +93,25 @@ func (v *ControlPlaneMutator) Handle(ctx context.Context, req admission.Request)
...
+			var ver versions.Version
+			var err error
+			// If version is not specified
+			if currentVersion == "" {
+				ver, err = versions.ParseVersion(mutator.DefaultVersion())
+			} else {
+				ver, err = versions.ParseVersion(currentVersion)
+			}
+			if err == nil && ver.AtLeast(versions.V2_5.Version()) {
+				mutator.SetTracingType("None")
```
yannuil: I would use `v2.TracerTypeNone` instead of the string `None`.

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -299,3 +320,29 @@ func (m *smcpv2mutator) SetOpenShiftRouteEnabled(value bool) {
...
+		return nil
+	}
+
+	route := gateways.OpenShiftRoute
+
+	if route == nil {
+		return nil
+	}
+
+	return route.Enabled
```
yannuil: Forget to remove?

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -299,3 +320,29 @@ func (m *smcpv2mutator) SetOpenShiftRouteEnabled(value bool) {
...
 }
+
+func (m *smcpv1mutator) IsTracingTypeSpecified() *bool {
+	return nil
+}
+
+func (m *smcpv1mutator) SetTracingType(value string) {}
+
+func (m *smcpv2mutator) IsTracingTypeSpecified() *bool {
+	tracingType := m.smcp.Spec.Tracing.Type
```
yannuil: `m.smcp.Spec.Tracing` maybe `nil`. We want to check it before accessing `Type`.

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -299,3 +320,29 @@ func (m *smcpv2mutator) SetOpenShiftRouteEnabled(value bool) {
...
+
+	if route == nil {
+		return nil
+	}
+
+	return route.Enabled
+}
+
+func (m *smcpv2mutator) SetTracingType(value string) {
+	m.patches = append(m.patches, jsonpatch.NewPatch("add", "/spec/gateways/openshiftRoute/enabled", value))
```
yannuil: `"/spec/gateways/openshiftRoute/enabled"` should be changed accordingly. Setting `/spec/tracing/type` will fail if `/spec/tracing` is `nil`.

yannuil (changes requested):

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -93,6 +93,25 @@ func (v *ControlPlaneMutator) Handle(ctx context.Context, req admission.Request)
...
+		if newOpenShiftTracing == nil {
+			var ver versions.Version
+			var err error
+			// If version is not specified
+			if currentVersion == "" {
+				ver, err = versions.ParseVersion(mutator.DefaultVersion())
+			} else {
+				ver, err = versions.ParseVersion(currentVersion)
+			}
+			if err == nil && ver.AtLeast(versions.V2_5.Version()) {
```
jewertow:
```suggestion
			if err == nil && ver.AtLeast(versions.V2_6.Version()) {
```

jewertow (changes requested):

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -299,3 +320,21 @@ func (m *smcpv2mutator) SetOpenShiftRouteEnabled(value bool) {
...
+
+func (m *smcpv2mutator) IsTracingTypeSpecified() *bool {
+	tracingType := m.smcp.Spec.Tracing
+
+	if tracingType == nil {
+		return nil
+	
+}
+
+func (m *smcpv2mutator) SetTracingType(value string) {
```
eoinfennessy: This function could take a `v2.TracerType` value instead of a string.

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -299,3 +320,21 @@ func (m *smcpv2mutator) SetOpenShiftRouteEnabled(value bool) {
...
+
+func (m *smcpv1mutator) SetTracingType(value string) {}
+
+func (m *smcpv2mutator) IsTracingTypeSpecified() *bool {
+	tracingType := m.smcp.Spec.Tracing
+
+	if tracingType == nil {
+		return nil
+	
+}
```
eoinfennessy: I think returning a `bool` makes more sense than returning `*bool` here. We only need to know whether this field has been explicitly set by a user (`true`) or not specified (`false`). The nil value isn't useful.

Also, you are not checking the `m.smcp.Spec.Tracing.Type` here.

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -299,3 +320,26 @@ func (m *smcpv2mutator) SetOpenShiftRouteEnabled(value bool) {
...
+	tracing := m.smcp.Spec.Tracing
+
+	if tracing == nil {
+		return false
+	}
+
+	if tracing.Type == "" {
+		return false
+	}
+	return true
```
jewertow:
```suggestion
return !(tracing == nil || tracing.Type == "")
```

DeanDoyle1502: Added in this suggestion

mkralik3:
@DeanDoyle1502 Can you please also remove `.spec.tracing.type: Jaeger` (or change it to None) from the sample SMCP which is shown during the creation of SMCP in the OpenShift UI console? (I think, it is located in clusterserviceversion file)

dan3bew:
> @DeanDoyle1502 Can you please also remove `.spec.tracing.type: Jaeger` (or change it to None) from the sample SMCP which is shown during the creation of SMCP in the OpenShift UI console? (I think, it is located in clusterserviceversion file)

Is this only for displaying purpose or for creating new SCMP in the web console? We don't want to directly change the default values. When the user is doing an upgrade to 2.6, the default values get applied and changing the underlying defaults will alter their existing behavior.

mkralik3:
It is prefilled when you creating a new SMCP via web ui, see (e.g. for 2.5)
image

I believe, it is defined in the csv file, maybe here: https://github.com/maistra/istio-operator/blob/maistra-2.6/manifests-servicemesh/2.6.0/servicemeshoperator.v2.6.0.clusterserviceversion.yaml#L36

dan3bew (changes requested):

yannuil:
> @DeanDoyle1502 Can you please also remove `.spec.tracing.type: Jaeger` (or change it to None) from the sample SMCP which is shown during the creation of SMCP in the OpenShift UI console? (I think, it is located in clusterserviceversion file)

Is this only for displaying purpose or for creating new SCMP in the web console? We don't want to directly change the default values. When the user is doing an upgrade to 2.6, the default values get applied and changing the underlying defaults will alter their existing behavior. 

EDITED:
> I believe, it is defined in the csv file, maybe here: [maistra-2.6/manifests-servicemesh/2.6.0/servicemeshoperator.v2.6.0.clusterserviceversion.yaml#L36](https://github.com/maistra/istio-operator/blob/maistra-2.6/manifests-servicemesh/2.6.0/servicemeshoperator.v2.6.0.clusterserviceversion.yaml?rgh-link-date=2024-06-26T06%3A58%3A50Z#L36)

We probably should remove the tracing section.

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -299,3 +320,19 @@ func (m *smcpv2mutator) SetOpenShiftRouteEnabled(value bool) {
...
+func (m *smcpv1mutator) SetTracingType(value v2.TracerType) {}
+
+func (m *smcpv2mutator) IsTracingTypeSpecified() bool {
+	tracing := m.smcp.Spec.Tracing
+
+	return !(tracing == nil || tracing.Type == "")
+}
+
+func (m *smcpv2mutator) SetTracingType(value v2.TracerType) {
+	m.patches = append(m.patches, jsonpatch.NewPatch("add", "/spec/tracing/type", string(value)))
```
yannuil: Setting `/spec/tracing/type` will fail when `/spec/tracing` is empty. We need to construct a `TracingConfig`  while preserving any user defined values.

yannuil (changes requested):

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -93,6 +93,25 @@ func (v *ControlPlaneMutator) Handle(ctx context.Context, req admission.Request)
 		}
 	}
 
+	// removing tracing by default on creation of v2.6 SMCP
+	if req.AdmissionRequest.Operation == admissionv1beta1.Create {
```
jewertow: We already have an "if" statement for "create" events, so please move the body of this condition to the previous one.

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -93,6 +93,25 @@ func (v *ControlPlaneMutator) Handle(ctx context.Context, req admission.Request)
...
+
+		if !newOpenShiftTracing {
+			var ver versions.Version
+			var err error
+			// If version is not specified
+			if currentVersion == "" {
+				ver, err = versions.ParseVersion(mutator.DefaultVersion())
+			} else {
+				ver, err = versions.ParseVersion(currentVersion)
+			}
```
jewertow: Instead of copying and pasting this piece of code, please refactor it as I did it [here](https://github.com/maistra/istio-operator/compare/maistra-2.6...jewertow:istio-operator:OSSM-6693-version-3-backup#diff-0b39680bd1f117e8386c7dd7be4e9b8ff69867f3df17df13037795272ae90f49).

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -169,6 +188,8 @@ type smcpmutator interface {
 	GetPatches() []jsonpatch.JsonPatchOperation
 	IsOpenShiftRouteEnabled() *bool
 	SetOpenShiftRouteEnabled(bool)
+	IsTracingTypeSpecified() bool
+	SetTracingType(v2.TracerType)
```
jewertow: This function always takes one argument - `TracerTypeNone` - so I would change its signature to:
```suggestion
	DisableTracing()
```

jewertow (changes requested):

DeanDoyle1502:
Issues with running gencheck sucessfully on my environment, will have to fix tomorrow

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -169,6 +167,8 @@ type smcpmutator interface {
 	GetPatches() []jsonpatch.JsonPatchOperation
 	IsOpenShiftRouteEnabled() *bool
 	SetOpenShiftRouteEnabled(bool)
+	IsTracingTypeSpecified() bool
+	DisableTracing(v2.TracerType)
```
jewertow: As I said in the previous comment, this function always takes the same argument, so it can take no argument and just always set `v2.TracerTypeNone`.
```suggestion
	DisableTracing()
```

jewertow (changes requested):

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
DeanDoyle1502: Fixed this now

build/manifest-templates/clusterserviceversion.yaml
```diff
@@ -40,24 +40,13 @@ metadata:
...
                "type": "Istiod"
             },
             "telemetry": {
                "type": "Istiod"
             },
             "addons": {
-              "jaeger": {
-                "install": {
-                  "storage": {
-                    "type": "__JAEGER_STORAGE__"
```
eoinfennessy, line 57: You can delete all uses of `JAEGER_STORAGE` and `__JAEGER_STORAGE__` in `build/generate-manifests.sh` too.

yannuil (approved):

jewertow (approved):

PR #1805, opened 2024-06-19
OSSM-6693: Enable Gateway API by default in cluster-wide mode

jewertow:


dgn (commented):
This patch would mean that for GatewayControllerMode to work, you still have to set `techPreview.gatewayAPI.enabled=true` first, which is a bit counterintuitive. Also, the field `values.gatewayAPI.enabled` will show up as set to `false` in the SMCP's status, even when in clusterWide mode, which might again be confusing.

I think the more correct approach would be to actually set the `gatewayAPI.enabled=true` in the conversion code if we're in cluster-wide mode.

jewertow:
> for GatewayControllerMode to work, you still have to set `techPreview.gatewayAPI.enabled`=true

Gateway controller mode was implemented for the ingress controller that is tech-preview too, so I didn't care about that.

> Also, the field `values.gatewayAPI.enabled` will show up as set to false in the SMCP's status, even when in clusterWide mode, which might again be confusing.

Ok, this convince me to implement proper conversion.

dgn (approved):
LGTM, do we need the duplicate tests for `"bool" vs bool` though?

jewertow:
> LGTM, do we need the duplicate tests for "bool" vs bool though?

No, we do not. I removed it.

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -37,6 +37,13 @@ var (
 	_ inject.Client             = (*ControlPlaneMutator)(nil)
 	_ admission.DecoderInjector = (*ControlPlaneMutator)(nil)
 
+	featureDisabled = false
```
rcernich: Should this be `:=`?  Is the address of this used elsewhere?

jewertow: No, you can't use `:=` with `var`.

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -265,37 +274,38 @@ func (m *smcpv2mutator) GetProfiles() []string {
...
 
-func (m *smcpv2mutator) SetOpenShiftRouteEnabled(value bool) {
-	gateways := m.smcp.Spec.Gateways
-
-	if gateways == nil {
-		gateways = &v2.GatewaysConfig{}
+func (m *smcpv2mutator) EnableGatewayAPIIfNotSet() {
+	isGatewayAPISet := func(spec v2.ControlPlaneSpec) bool {
+		if spec.TechPreview != nil {
+			// ignore error - if techPreview.gatewayAPI has wrong type, we still need to check PILOT_ENABLE_GATEWAY_API
```
dgn: hmm you say we need to check it in the comment, but then you don't actually check it - which is fine, I don't think we need to. But we should adjust the comment I guess

dgn (commented):
Hmm. I'm not sure I understand. What was wrong with the previous approach? Why do we explicitly set `techPreview` fields? I'm not sure we should do that, it'll look like we're enabling features that are not GA.

In the previous approach, it would have been enabled by default, and in the status it would have appeared as:

```
status:
  appliedValues:
    gatewayAPI:
      enabled: true
```

whereas now it will also show up in the spec, but under techPreview.
```
spec:
  techPreview:
    gatewayAPI:
      enabled: true
```

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
jewertow: Yes, I forgot to remove that comment.

jewertow:
> What was wrong with the previous approach?

Conversion would enable gateway API in all SMCPs - new and existing.

> I'm not sure we should do that, it'll look like we're enabling features that are not GA.

First of all, we disable IOR the same way and we are going to do the same with Jaeger. Second, this is why I started with changing the template. In that approach, proper env vars were enabled, while tech-preview feature was disabled, but you expected to set techPreview. What's the difference between seeing `spec.techPreview.gatewayAPI.enabled` vs `status.appliedSpec.techPreview.gatewayAPI.enabled`? For me, there is no difference. None of these is less confusing. In both cases you have enabled tech-preview feature.

luksa (approved):

PR #1806, opened 2024-06-20
OSSM-6411: Copy appropriate istio-cni binary based on host OS

eoinfennessy:


resources/helm/overlays/istio_cni/templates/daemonset-v2.4.yaml
```diff
@@ -67,7 +67,52 @@ spec:
...
               path: /readyz
               port: 8000
-          command: ["install-cni"]
+          command:
+            - /bin/sh
+            - -c
+            - |
+              set -xe
+              # Collect host OS information
+              . /host/etc/os-release
```
dgn, line 76: ~I don't have a RHEL instance to test with, but this doesn't work on fedora- should we guard against that? Previously, we also supported running the operator on kind IIRC, that won't work with this I guess~

Nevermind, that was my mistake not removing the `/host` prefix :man_facepalming:

eoinfennessy:
Whoops... Indentation has been flattened here for some reason; fixing now

resources/helm/overlays/istio_cni/templates/daemonset-v2.4.yaml
```diff
@@ -67,7 +67,55 @@ spec:
...
+                    echo "WARNING: Unsupported Fedora variant=${VARIANT_ID}"
+                  fi
+                ;;
+                *)
+                  echo "WARNING: Unsupported OS ID=${ID}"
+                ;;
+              esac
+
+              # Select the appropriate CNI binary to copy based on the OS version
+              sourcebin=
```
luksa: Is this line needed?

resources/helm/overlays/istio_cni/templates/daemonset-v2.4.yaml
```diff
@@ -67,7 +67,55 @@ spec:
...
+              case "${rhelmajor}" in
+                8)
+                  sourcebin=/opt/cni/bin/istio-cni-rhel8
+                ;;
+                9)
+                  sourcebin=/opt/cni/bin/istio-cni-rhel9
+                ;;
+                *)
+                  sourcebin=/opt/cni/bin/istio-cni-rhel9
+                  echo "WARNING: RHEL major version rhelmajor=${rhelmajor} is unsupported; falling back to istio-cni binary compiled on latest RHEL version: sourcebin=${sourcebin}"
```
luksa: This warning is a bit strange on non-RHEL systems, since it says `RHEL major version rhelmajor= is unsupported`. As a user, I would have no idea what this is trying to tell me.

eoinfennessy: Yeah, that makes sense. I could change it to something like...
```
WARNING: The host OS is not supported in FIPS-compliant environments; using istio-cni binary compiled on latest RHEL version: sourcebin=${sourcebin}"
```

resources/helm/overlays/istio_cni/templates/daemonset-v2.4.yaml
eoinfennessy: No, it's not necessary here. I will delete it.

resources/helm/overlays/istio_cni/templates/daemonset-v2.4.yaml
```diff
@@ -67,7 +67,54 @@ spec:
...
+              case "${rhelmajor}" in
+                8)
+                  sourcebin=/opt/cni/bin/istio-cni-rhel8
+                ;;
+                9)
+                  sourcebin=/opt/cni/bin/istio-cni-rhel9
+                ;;
+                *)
+                  sourcebin=/opt/cni/bin/istio-cni-rhel9
+                  echo "INFO: Running in a non-FIPS compliant environment; using istio-cni binary compliled with latest RHEL version: sourcebin=${sourcebin}"
```
luksa:
```suggestion
                  echo "INFO: Running in a non-FIPS compliant environment; using istio-cni binary compiled with latest RHEL version: sourcebin=${sourcebin}"
```

luksa (approved):

PR #1818, opened 2024-06-29
OSSM-6757 Ensure CNI pod starts when using older images

luksa:
Instead of failing with `mv: cannot stat '/opt/cni/bin/istio-cni-rhel9': No such file or directory`, the startup script now emits a warning and tries to use the original `/opt/cni/bin/istio-cni` binary.

### DIFF
--- a/resources/helm/overlays/istio_cni/templates/daemonset-v2.4.yaml
+++ b/resources/helm/overlays/istio_cni/templates/daemonset-v2.4.yaml
@@ -110,9 +110,13 @@ spec:
                 ;;
               esac
 
-              # Move the source binary to where install-cni expects it to be located and clean up unused binaries
-              mv $sourcebin /opt/cni/bin/istio-cni
-              rm /opt/cni/bin/istio-cni-rhel*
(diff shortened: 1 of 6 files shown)


luksa:
@eoinfennessy please review when you're back (even if someone else approves the PR)

yannuil (approved):

eoinfennessy:
@luksa, looks good to me. Thanks for fixing.

Out of curiosity, is this change only for development and/or testing purposes, or is there any way that these DaemonSet templates could end up being used in production with references to older images?

luksa:
@eoinfennessy Yes, this is only needed when using non-product builds (and older builds) of the istio-cni image. Only the new product builds contain the `-rhel8` and `-rhel9` binaries, whereas the non-product build doesn't (see https://github.com/Maistra/istio/blob/09c78459a1d8a410ca229128d1baff1246bbc019/maistra/Dockerfile.istio-cni#L23).

For product builds, I don't think we'll ever need this change, since the config for the 2.6.0 release of the operator will always point to the new CNI images that will contain the two binaries.

PR #1819, opened 2024-07-02
OSSM-6762 Move webhook management from the operator to OLM

luksa:
- The operator will no longer create the Mutating/ValidatingWebhookConfigurations and associated Service, Secret, and ConfigMap, as they will now be created by OLM.
- On startup, the operator will delete these old resources so that they don't conflict with the ones created by OLM.
- Previously, the operator had to restart after creating the resources so that the Secret holding the certificate and private key was injected into the operator pod. As this secret is now created by OLM before the operator starts, the operator no longer needs to restart on its first run.
- The servicemeshcontrolplanes CRD must have CA injection disabled so that OLM and the service-ca operator don't both inject the CA bundle into the CRD.

### DIFF
--- a/build/generate-crds.sh
+++ b/build/generate-crds.sh
@@ -32,7 +32,7 @@ function generateCRDs() {
   # Add maistra-version label
   sed -i -e "s/^  annotations:/  labels:\n    maistra-version: $MAISTRA_VERSION\n\\0/" deploy/crds/*
 
-  sed -i -e 's/^  annotations:/  annotations:\n    service.beta.openshift.io\/inject-cabundle: "true"/' \
+  sed -i -e 's/^  annotations:/  annotations:\n    service.beta.openshift.io\/inject-cabundle: "false"/' \
       deploy/crds/maistra.io_servicemeshcontrolplanes.yaml
 
   for bundle_dir in ${BUNDLE_DIRS}; do
--- a/deploy/maistra-operator.yaml
+++ b/deploy/maistra-operator.yaml
@@ -10999,13 +10999,12 @@ metadata:
 
 ---
 
+# NOTE: this is only needed on vanilla Kubernetes; On OpenShift, the Operator Lifecycle Manager (OLM) creates the service automatically
 apiVersion: v1
 kind: Service
 metadata:
-  name: maistra-admission-controller
+  name: istio-operator-service
   namespace: istio-operator
-  annotations:
-    service.beta.openshift.io/serving-cert-secret-name: maistra-operator-serving-cert
 spec:
   ports:
   - port: 443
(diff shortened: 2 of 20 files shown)


build/generate-crds.sh
```diff
@@ -32,7 +32,7 @@ function generateCRDs() {
   # Add maistra-version label
   sed -i -e "s/^  annotations:/  labels:\n    maistra-version: $MAISTRA_VERSION\n\\0/" deploy/crds/*
 
-  sed -i -e 's/^  annotations:/  annotations:\n    service.beta.openshift.io\/inject-cabundle: "true"/' \
+  sed -i -e 's/^  annotations:/  annotations:\n    service.beta.openshift.io\/inject-cabundle: "false"/' \
       deploy/crds/maistra.io_servicemeshcontrolplanes.yaml
```
luksa, line 36: Instead of removing this annotation, we set it to false to ensure that when the CRD manifest is applied to a cluster that already has this CRD, the annotation is actually overwritten. I believe that if we simply removed this annotation from the manifest, the annotation wouldn't be removed.

deploy/maistra-operator.yaml
```diff
@@ -11103,9 +11103,6 @@ spec:
         - name: operator-olm-config
           mountPath: /etc/operator/olm
           readOnly: true
-        - name: webhook-tls-volume
-          readOnly: true
-          mountPath: /tmp/k8s-webhook-server/serving-certs
```
luksa, line 11108: I'm removing this volume & volumeMount, because OLM will now inject them for us.

luksa:
To anyone reviewing and manually testing this:
- make sure to install an old version of the operator first so that you can confirm that the new operator cleans up the old stuff
- confirm that the mutating, validating, and conversion webhooks all work
- check whether the `istio-operator-service` that is created by OLM is not being continuously deleted and recreated (I saw this in my initial tests, but later fixed the underlying cause);

jewertow (approved):

dgn (approved):
I was able to test this using [splatolm](https://github.com/ecordell/splatolm) and the following script:
```bash
#/bin/bash
oc create -n openshift-operators -f manifests-servicemesh/2.6.0/servicemeshoperator.v2.6.0.clusterserviceversion.yaml
oc get csv -n openshift-operators servicemeshoperator.v2.6.0 -o yaml > csv.yaml
splatolm csv.yaml > rbac.yaml
oc apply -n openshift-operators -f rbac.yaml
```

Note that if you're testing the upgrade, you'll have to remove the old subscription after doing this to trigger the install of 2.6.

All webhooks worked fine, both after a first-time install and an upgrade from 2.5.2.

I did not observe the service recreation problem that Marko mentioned.

PR #1821, opened 2024-07-02
OSSM-6602 Set default security identity type to ThirdParty for new 2.6 SMCP

yannuil:


pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -321,3 +335,34 @@ func (m *smcpv2mutator) DisableTracing() {
...
+	security := m.smcp.Spec.Security
+
+	if security == nil {
+		return false
+	}
+
+	identity := security.Identity
+
+	return !(identity == nil || identity.Type == "")
+}
```
eoinfennessy:
```suggestion
func (m *smcpv2mutator) IsSecurityIdentityTypeSpecified() bool {
	return m.smcp.Spec.Security != nil &&
		m.smcp.Spec.Security.Identity != nil &&
		m.smcp.Spec.Security.Identity.Type != ""
}
```

eoinfennessy: Nitpick; feel free to merge without.

luksa:
Please test this change when using `oc replace`, not just `oc create` or `oc apply`. We've found a similar issue here: https://issues.redhat.com/browse/OSSM-6767

luksa:
Also, do we know of other changes like this in 2.6? They should all be tested with `oc replace` and re-implemented if necessary.

resources/helm/v2.6/istio-control/istio-discovery/values.yaml
```diff
@@ -598,6 +598,13 @@ istio_cni:
   enabled: false
   chained: true
 
+# Gateway Settings
+gateways:
+  # Define the security context for the pod.
+  # If unset, this will be automatically set to the minimum privileges required to bind to port 80 and 443.
+  # On Kubernetes 1.22+, this only requires the `net.ipv4.ip_unprivileged_port_start` sysctl.
+  securityContext: {}
+
```
eoinfennessy: I don't understand why the changes here or changes to files in `resources/helm/v2.6/istio-control/istio-discovery/` are included in this PR. Were these all generated because of changes made in a previous PR?

luksa: Yes. These were pulled by `make gen`. They will disappear now that https://github.com/maistra/istio-operator/pull/1820 has been merged.

eoinfennessy: Thanks!

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
luksa: +1

eoinfennessy (approved):

eoinfennessy:
> Please test this change when using oc replace, not just oc create or oc apply. We've found a similar issue here: https://issues.redhat.com/browse/OSSM-6767
>
> Also, do we know of other changes like this in 2.6? They should all be tested with oc replace and re-implemented if necessary.

Ah, okay. Yes, #1802 is a similar change.

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -60,34 +60,45 @@ func (v *ControlPlaneMutator) Handle(ctx context.Context, req admission.Request)
...
-		if newOpenShiftRoute == nil && effectiveVersion.AtLeast(versions.V2_5.Version()) {
-			mutator.SetOpenShiftRouteEnabled(false)
-		}
+	if req.AdmissionRequest.Operation == admissionv1beta1.Update {
+		oldsmcp := &v2.ServiceMeshControlPlane{}
+		err := v.decoder.DecodeRaw(req.AdmissionRequest.OldObject, oldsmcp)
+		if err != nil {
+			log.Error(err, "error decoding admission request")
+		} else {
+			lastAppliedSpec := oldsmcp.Status.AppliedSpec
```
luksa: We shouldn't read from the status! It's way too fragile.

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -60,34 +60,45 @@ func (v *ControlPlaneMutator) Handle(ctx context.Context, req admission.Request)
...
 			oldVersion := mutator.OldVersion()
 			if currentVersion != oldVersion && oldVersion != versions.InvalidVersion.String() {
 				log.Info("Setting .spec.version to existing value", "version", oldVersion)
-				mutator.SetVersion(oldVersion)
-				effectiveVersion, _ = versions.ParseVersion(oldVersion)
+				currentVersion = oldVersion
 			}
 		}
 	}
+	mutator.SetVersion(currentVersion)
```
luksa: Does this work when the version is already set, given the fact that the patch generated by `SetVersion` adds the `version` field? Previously, we only called this function when the `version` field wasn't set.

resources/smcp-templates/v2.5/base
```diff
@@ -14,7 +14,7 @@ spec:
     egress:
       enabled: true
     openshiftRoute:
-      enabled: true
+      enabled: false
```
eoinfennessy: What happens in cases such as this:
1. A user has been using IOR using SMCP with version 2.4, without explicitly enabling IOR
2. Prior to the introduction of your new UPDATE logic, they upgrade to SMCP v2.5
3. After this change gets released and their operator updated they either...
    1. `kubectl replace` the SMCP, or
    2. `kubectl apply` some unrelated change to the SMCP

In this case I'm pretty sure `kubectl replace` would cause IOR to become disabled. I'm not sure about `kubectl apply`?

pkg/controller/servicemesh/webhooks/mutation/controlplane.go
```diff
@@ -60,34 +60,45 @@ func (v *ControlPlaneMutator) Handle(ctx context.Context, req admission.Request)
...
-	if req.AdmissionRequest.Operation == admissionv1beta1.Create {
-		// As we are deprecating IOR, on creating a v2.5 SMCP we want to disable IOR if not specified explicitly
-		newOpenShiftRoute := mutator.IsOpenShiftRouteEnabled()
-
-		if newOpenShiftRoute == nil && effectiveVersion.AtLeast(versions.V2_5.Version()) {
-			mutator.SetOpenShiftRouteEnabled(false)
-		}
+	if req.AdmissionRequest.Operation == admissionv1beta1.Update {
+		oldsmcp := &v2.ServiceMeshControlPlane{}
+		err := v.decoder.DecodeRaw(req.AdmissionRequest.OldObject, oldsmcp)
```
luksa: This assumes that the webhook is mutating a v2 SMCP, which isn't always the case. I might be wrong, but it looks like this is why we have the `smcpmutator` (so that the we can have two implementations - for v1 and for v2).

yannuil:
Closed as the road map for setting third-party as default move to 3.0

PR #1827, opened 2024-07-05
OSSM-6774 Use the existing kube clients to create metrics resources

luksa:
Previously, during startup, the operator instantiated a separate kube client and a separate discovery client to create the Service and ServiceMonitor for metrics. Now, the existing clients provided by the controller manager are used to create these two resources.

However, since these clients are only initialized when the manager is started, the code that creates the two metrics resources is now invoked by a runnable registered with the manager via `mgr.Add()`.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -11,7 +11,6 @@ import (
 	"strings"
 	"time"
 
-	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	"github.com/magiconair/properties"
 	"github.com/mitchellh/mapstructure"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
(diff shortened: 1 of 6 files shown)


cmd/manager/main.go
```diff
@@ -183,8 +182,14 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Add the Metrics Service
-	addMetrics(ctx, cfg)
+	err = mgr.Add(manager.RunnableFunc(func(stop <-chan struct{}) error {
+		createMetricsResources(context.Background(), cfg, mgr.GetClient(), dc)
+		return nil
+	}))
```
eoinfennessy, line 188: Can you please tell me what the purpose of `manager.RunnableFunc` is?

I read [its docs](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/manager#Runnable) but I don't understand why this is used here instead of calling `createMetricsResources` directly. Also, when is `Start` called on the `RunnableFunc`?

eoinfennessy (approved):
LGTM -- I just added one comment with a question about something I don't fully understand.

cmd/manager/main.go
eoinfennessy: Oh, sorry, I didn't see this was explained in the description; Makes sense now.

luksa: If we call `createMetricsResources()` directly (instead of having the manager call it), the function will fail when it tries to read from the manager's client. This happens because the client's caches aren't synced until the manager starts. The only reason we could previously call the function directly was because it was creating its own client.

In contrast, when the function is called via a `RunnableFunc` registered via `mgr.Add()`, the manager waits for the [caches to be synced before invoking the registered runnables](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.6.3/pkg/manager/internal.go#L579-L586). 

So, if you want to use the manager's client, you have to do it in a runnable registered with the manager.

eoinfennessy: Thank you for explaining! Great to know about this.

PR #1829, opened 2024-07-09
Simplify setting defaults

yannuil:


manifests-maistra/2.6.0/maistraoperator.v2.6.0.clusterserviceversion.yaml
mkralik3, line 1: Doesn't this return `Jaeger` type into the samle SMCP shown during the installation via OpenShift UI?
e.g. :
image
we shouldn't have this value there since it is not default value for 2.6 . (see my comment here https://github.com/maistra/istio-operator/pull/1802#issuecomment-2190958105 )

yannuil: Updated

resources/smcp-templates/v2.6/base
```diff
@@ -34,7 +34,7 @@ spec:
 
   security:
     identity:
-      type: Kubernetes
+      type: ThridParty
```
dgn: typo

dgn:
gencheck test failure is fixed in https://github.com/maistra/istio-operator/pull/1831 - sorry

resources/smcp-templates/v2.6/base
```diff
@@ -34,7 +34,7 @@ spec:
 
   security:
     identity:
-      type: Kubernetes
+      type: ThirdParty
```
rcernich, line 37: This change will break existing meshes if a control plane is updated from an older version to 2.6.  Existing proxies will not be able to update their serving certificates for mtls until they are reinjected.

resources/smcp-templates/v2.6/base
```diff
@@ -43,7 +43,7 @@ spec:
     type: None
 
   tracing:
-    type: Jaeger
+    type: None
```
rcernich, line 46: This will cause users updating from an older version to have Jaeger disappear.

yannuil:
Won't make change to 2.6.

PR #1830, opened 2024-07-12
OSSM-6812: Detect FIPS mode and enable `COMPLIANCE_POLICY` if necessary

jewertow:


pkg/controller/versions/strategy_v2_6.go
```diff
@@ -461,6 +462,18 @@ func (v *versionStrategyV2_6) Render(ctx context.Context, cr *common.ControllerR
 		return nil, fmt.Errorf("could not set field meshConfig.ingressControllerMode: %v", err)
 	}
 
+	// Enable FIPS automatically if necessary
+	data, err := ioutil.ReadFile("/proc/sys/crypto/fips_enabled")
+	if err != nil {
+		return nil, fmt.Errorf("could not read file /proc/sys/crypto/fips_enabled: %v", err)
```
luksa: Do all kernels expose this file?

pkg/controller/versions/strategy_v2_6.go
```diff
@@ -461,6 +462,18 @@ func (v *versionStrategyV2_6) Render(ctx context.Context, cr *common.ControllerR
...
+	}
+	if string(data) == "1" {
+		err = spec.Istio.SetField("global.fipsEnabled", true)
```
luksa, line 471: Is it okay to assume that all nodes are fips enabled based on only the node that runs the operator?

pkg/controller/versions/strategy_v2_6.go
jwendell: This and the other errors could also be converted to just a warning. If that file doesn't exist, it should not block other things to happen.

pkg/controller/versions/strategy_v2_6.go
jwendell: If I understand [this page](https://docs.openshift.com/container-platform/4.16/installing/installing-fips.html#installing-fips-mode_installing-fips) correctly, fips is installed at cluster level, meaning all nodes are either fips-enabled or disabled.

pkg/controller/versions/strategy_v2_6.go
jewertow: Thanks! It wouldn't work on regular clusters. I logged a message on info level, because our logger does not support warnings.

rcernich:
I thought this was already fixed in proxy itself.

Also, shouldn't we be checking the node on which the proxy will be running?  I'm not sure if it's possible to have a cluster where only some nodes are running FIPS, so this may not be an issue.

dgn:
> @jewertow: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> Test name 	Commit 	Details 	Required 	Rerun command
> ci/prow/gencheck 	[b3eb6ce](https://github.com/maistra/istio-operator/commit/b3eb6cea56df4ed9df548df6b2c2636100a918e2) 	[link](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/maistra_istio-operator/1830/pull-ci-maistra-istio-operator-maistra-2.6-gencheck/1811726477129945088) 	true 	`/test gencheck`
> 
> [Full PR test history](https://prow.ci.openshift.org/pr-history?org=maistra&repo=istio-operator&pr=1830). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:jewertow).

test failure is fixed in https://github.com/maistra/istio-operator/pull/1831

PR #1833, opened 2024-07-31
OSSM-6857 Ensure validating webhook works when both 2.x and 3.x are installed

luksa:
When you install 3.x, the default version for some CRDs changes from e.g. v1beta1 to v1. This causes kubectl to default to using the v1 version. But this caused the validating webhook in 2.x to fail, because it was invoked with the v1 version of the resource, which it doesn't understand. The root cause was that the validatingwebhookconfiguration previously specified that it supports all apiVersions.

In this commit, we specify the actual apiVersions that the webhook supports. For example, for DestinationRules, the updated validatingwebhookconfiguration causes kubernetes to convert the DestinationRule from v1 to v1beta before invoking the webhook.

### DIFF
--- a/resources/helm/overlays/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml
+++ b/resources/helm/overlays/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml
@@ -4,7 +4,6 @@ kind: ValidatingWebhookConfiguration
 metadata:
   name: istio-validator-{{ .Values.revision | default "default" }}-{{ .Release.Namespace }}
   labels:
-    maistra-version: "2.6.1"
     app: istiod
     release: {{ .Release.Name }}
     istio: istiod
--- a/build/patch-charts.sh
+++ b/build/patch-charts.sh
@@ -99,51 +99,6 @@ gatewayAPI:
 }
 
 function patchGalley() {
-  echo "patching Galley specific Helm charts"
-  # galley
-  # Webhooks are not namespaced!  we do this to ensure we're not setting owner references on them
-  # add namespace selectors
-  # remove define block
-  webhookconfig=${HELM_DIR}/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml
-
-  # remove original objectSelector
-  sed_wrap -i '/.*objectSelector:/,/.*{{- end }}/d' "${webhookconfig}"
-
-  # replace namespaceSelector and insert maistra objectSelector
-  sed_wrap -i -e 's|\(\(^ *\)rules:\)|\2namespaceSelector:\
-\2  matchExpressions:\
-\2  - key: maistra.io/member-of\
-\2    operator: In\
-\2    values:\
-\2    - {{ .Release.Namespace }}\
-\2objectSelector:\
-\2  matchExpressions:\
-\2  - key: maistra-version\
-\2    operator: DoesNotExist\
-\1|' "${webhookconfig}"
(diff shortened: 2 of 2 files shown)


resources/helm/overlays/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,141 @@
...
+      matchExpressions:
+      - key: maistra.io/member-of
+        operator: In
+        values:
+        - {{ .Release.Namespace }}
+    objectSelector:
+      matchExpressions:
+      - key: maistra-version
+        operator: DoesNotExist
+    rules:
```
yxun, line 35:
```suggestion
    matchPolicy: Equivalent
    rules:
```

Should we add `matchPolicy: Equivalent` as sridharg suggested in the Slack thread ?

resources/helm/v2.6/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml
```diff
@@ -52,21 +52,86 @@ webhooks:
         - "*"
         resources:
         - "*"
+
       - operations:
           - CREATE
           - UPDATE
         apiGroups:
           - security.istio.io
+        apiVersions:
```
yxun, line 61: Same question about adding `matchPolicy: Equivalent` above

yxun (approved):

yxun (changes requested):
I think we may also include apiVersion v1 for Telemetry resources

See latest istio doc example :
https://istio.io/latest/docs/reference/config/telemetry/

resources/helm/overlays/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,141 @@
...
+        resources:
+          - proxyconfigs
+
+      - operations:
+          - CREATE
+          - UPDATE
+        apiGroups:
+          - telemetry.istio.io
+        apiVersions:
+          - v1alpha1
```
yxun, line 120:
```suggestion
          - v1alpha1
          - v1
```

resources/helm/v2.6/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml
```diff
@@ -52,21 +52,86 @@ webhooks:
...
+        resources:
+          - proxyconfigs
+
+      - operations:
+          - CREATE
+          - UPDATE
+        apiGroups:
           - telemetry.istio.io
+        apiVersions:
+          - v1alpha1
```
yxun, line 121:
```suggestion
          - v1alpha1
          - v1
```

resources/helm/overlays/istio-control/istio-discovery/templates/validatingwebhookconfiguration.yaml
luksa: The [CRD manifest](https://github.com/Maistra/istio-operator/blob/maistra-2.6/resources/helm/v2.6/istio-init/files/telemetries.telemetry.istio.io.crd.yaml#L37) contains only version v1alpha1. Maybe v1 was introduced in a later Istio version.

yxun (approved):

jewertow (approved):

PR #1835, opened 2024-07-31
[maistra-2.4] OSSM-6857 Ensure validating webhook works when both 2.x and 3.x are installed

luksa:
When you install 3.x, the default version for some CRDs changes from e.g. v1beta1 to v1. This causes kubectl to default to using the v1 version. But this caused the validating webhook in 2.x to fail, because it was invoked with the v1 version of the resource, which it doesn't understand. The root cause was that the validatingwebhookconfiguration previously specified that it supports all apiVersions.

In this commit, we specify the actual apiVersions that the webhook supports. For example, for DestinationRules, the updated validatingwebhookconfiguration causes kubernetes to convert the DestinationRule from v1 to v1beta before invoking the webhook.

### DIFF
--- a/build/patch-charts.sh
+++ b/build/patch-charts.sh
@@ -90,51 +90,6 @@ gatewayAPI:
 }
 
 function patchGalley() {
-  echo "patching Galley specific Helm charts"
-  # galley
-  # Webhooks are not namespaced!  we do this to ensure we're not setting owner references on them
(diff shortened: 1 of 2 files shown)


yxun (approved):

yxun:
~I am going to fix the 2.4 duplicated jobs.~

PR #1863, opened 2025-01-30
Use CA rather than the server cert in patching webhook

nrfox:
When running the istio-operator with a 3.0 istiod, both will try to patch the validator webhook's caBundle. This isn't a problem if both try to patch the same bundle but with the `istiod-tls` secret present, the 2.6 operator will actually patch the istiod server cert `tls.crt` rather than the CA cert `ca.crt`. This patches changes the operator to patch the CA cert `ca.crt`.


dgn (approved):

pkg/controller/servicemesh/webhookca/controller.go
```diff
@@ -97,7 +97,7 @@ var autoRegistrationMap = map[string]CABundleSource{
 			},
 			{
 				SecretName: istiodCertManagerSecretName,
-				Key:        common.IstiodTLSSecretCertKey,
+				Key:        common.IstiodTLSSecretCertCAKey,
```
FilipB, line 100: Should this be done also for mutating web hook?

nrfox: It can be although it does not seem like the OSSM2 operator tries to insert the istiod tls cert into the OSSM3 webhook. I'm guessing because the ossm2 webhook has a different naming scheme:
```
istio-sidecar-injector-ossm3     2          20h
istiod-basic-istio-system        1          22h
```

We should probably update for `cacerts` as well since that will likely have the same problem.

nrfox: @FilipB I updated the rest of the istiod webhooks and updated the tests as well

jewertow (approved):

nrfox:
@FilipB here's the maistra-test-tool test for this: https://github.com/maistra/maistra-test-tool/pull/773. Covers `ClusterWide` mode but not `MultiTenant` yet.

FilipB:
We had some failures when running MTT against this PR. Checking.

FilipB:
MTT is passing now. Failures were caused by env misconfiguration.

FilipB (approved):

PR #1880, opened 2025-12-15
set Kiali CR cluster_wide_access to false always

jmazzitelli:
fixes: https://issues.redhat.com/browse/OSSM-11635

The Kiali CRD Schema sets the default for cluster_wide_access to true. We need it always set to false to support multi-tenancy. Since accessible_namespaces is never set to "**" in Maistra/OSSM 2.x, setting this to false always is appropriate.

PR #1881, opened 2026-01-08
[OSSM-11635] Set kiali cluster_wide_access value according to SMCP mode

mkralik3:
Fix regression after https://github.com/maistra/istio-operator/pull/1880 where istio-operator generates invalid Kiali CR
```
  accessible_namespaces:
      **
  cluster_wide_access:  false
```
when SMCP mode is set to `ClusterWide`.
Since istio-operator manages also `accessible_namespaces` here https://github.com/maistra/istio-operator/blob/maistra-2.6/pkg/controller/servicemesh/memberroll/controller.go#L378 , hardcoded `cluster_wide_access:  false` in KialiCR is not possible.

The istio-operator should set `.spec.deployment.cluster_wide_access` in the created/managed KialiCR according to the value in SMCP `.spec.mode`, so:

| SMCP .spec.mode | Kiali .spec.deployment.cluster_wide_access |
| ------------- | ------------- |
| ClusterWide | true  |
| MultiTenant or empty  | false |

==============

Tested (with generated operator image quay.io/mkralik3/istio-operator:kiali) that
- [x] Correct Kiali CR (cluster_wide_access:  false) when SMCP `.spec.mode` is empty or `MultiTenant`
- [x] Correct Kiali CR (cluster_wide_access:  true) when SMCP `.spec.mode` is empty or `ClusterWide`
- [x] That update SMCP `.spec.mode` causes an update of Kiali CR
- [x] That older Kiali operator (`1.17.10`) can work with this as well ( since some customers can be on older Kiali operator ) 
- [x] Standalone Kiali CR is not touched

### DIFF
--- a/pkg/controller/servicemesh/controlplane/addons.go
+++ b/pkg/controller/servicemesh/controlplane/addons.go
@@ -56,6 +56,17 @@ func (r *controlPlaneInstanceReconciler) patchKiali(ctx context.Context, grafana
 		},
 	}
 
+	// SMCP mode cluster wide access
+	if r.Instance.Spec.Mode == maistrav2.ClusterWideMode {
+		if err := updatedKiali.Spec.SetField("deployment.cluster_wide_access", true); err != nil {
+			return common.RequeueWithError(errorOnSettingValueInKialiCR("deployment.cluster_wide_access", err))
+		}
+	} else {
+		if err := updatedKiali.Spec.SetField("deployment.cluster_wide_access", false); err != nil {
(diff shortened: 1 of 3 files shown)


dgn (commented):
Generally LGTM but would appreciate a test -- we should have probably added one when doing the initial work in #1880

jmazzitelli (approved):
this looks correct now

dgn (approved):
approving to unblock. tests can be added in follow-up PR

PR #1884, opened 2026-01-14
Add test for setting cluster_wide_access

mkralik3:
Added tests after PR https://github.com/maistra/istio-operator/pull/1881 

- Fixed Kiali CR cluster_wide_access field to be correctly set based on SMCP mode during creation
- Added 5 new test cases for addon_test covering:
  - SMCP with ClusterWide mode
    - check that Kiali CR was created with `cluster_wide_access=true`
  - SMCP with MultiTenant mode
    - check that Kiali CR was created with `cluster_wide_access=false`
  - SMCP without any mode (no new test, this is implicitly test by the other tests)
    - check that Kiali CR was created with `cluster_wide_access=false`
  - Create Kiali CR with `cluster_wide_access=true` and than create SMCP with MultiTenant mode
    - check that Kiali CR was patched to  `cluster_wide_access=false`
  - Create Kiali CR with `cluster_wide_access=false` and than create SMCP with ClusterWide mode
    - check that Kiali CR was patched to  `cluster_wide_access=true`
  - Create Kiali CR without any `cluster_wide_access` and than create SMCP with MultiTenant mode
    - check that Kiali CR was patched to  `cluster_wide_access=false`

### DIFF
--- a/pkg/controller/servicemesh/controlplane/addons_test.go
+++ b/pkg/controller/servicemesh/controlplane/addons_test.go
@@ -114,7 +114,7 @@ func TestAddonsInstall(t *testing.T) {
 			},
 			create: IntegrationTestValidation{
 				Verifier: Verify("create").On("kialis").Named(kialiName).In(controlPlaneNamespace).
-					Passes(ExpectedKialiCreate(jaegerExistingName, domain, prometheusPasswordExpected)),
+					Passes(ExpectedKialiCreate(jaegerExistingName, domain, prometheusPasswordExpected, false)),
 				Assertions: ActionAssertions{
 					Assert("create").On("kialis").Named(kialiName).In(controlPlaneNamespace).IsSeen(),
 					Assert("create").On("jaegers").Named(jaegerName).In(controlPlaneNamespace).IsNotSeen(),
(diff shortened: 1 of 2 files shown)


pkg/controller/servicemesh/controlplane/addons_test.go
```diff
@@ -114,7 +114,7 @@ func TestAddonsInstall(t *testing.T) {
 			},
 			create: IntegrationTestValidation{
 				Verifier: Verify("create").On("kialis").Named(kialiName).In(controlPlaneNamespace).
-					Passes(ExpectedKialiCreate(jaegerExistingName, domain, prometheusPasswordExpected)),
+					Passes(ExpectedKialiCreate(jaegerExistingName, domain, prometheusPasswordExpected, false)),
```
jewertow: When reading this function I had to jump to the function definition to understand what this `false` parameter means, so I would suggest to follow existing convention with self-describing variables, e.g. `clusterWideModeDisabled` / `clusterWideModeEnabled`.
```suggestion
					Passes(ExpectedKialiCreate(jaegerExistingName, domain, prometheusPasswordExpected, clusterWideDisabled)),
```

pkg/controller/servicemesh/controlplane/hooks.go
```diff
@@ -170,6 +171,12 @@ func (r *controlPlaneInstanceReconciler) patchKialiConfig(ctx context.Context, o
 	if err != nil {
 		return fmt.Errorf("could not set tracing password in kiali CR: %s", err)
 	}
+	// SMCP mode cluster wide access
+	clusterWideAccess := r.Instance.Spec.Mode == maistrav2.ClusterWideMode
```
jewertow: This comment would not be need if the variable name included `smcp`, e.g.
```suggestion
	smcpClusterWideAccess := r.Instance.Spec.Mode == maistrav2.ClusterWideMode
```

jewertow (approved):
LGTM, left only non-blocking comments.

mkralik3:
@jewertow thanks for the review. Pushed changes

PR #1886, opened 2026-01-15
[OSSM-12223] Fix tmp invalid KialiCR after update MultiTenanty to ClusterWide in SMCP

mkralik3:
The memberroll controller sets `accessible_namespaces` and `api.namespaces.exclude` according to the SMCP mode here: https://github.com/maistra/istio-operator/blob/maistra-2.6/pkg/controller/servicemesh/memberroll/controller.go#L377-L381 

When a new SMCP is created, there is no issue:
- The Maistra operator generates the Kiali Custom Resource (CR) without `cluster_wide_access` set. (since it is not in kiali cr helm template)
- The memberroll controller populates `accessible_namespaces`.
- The `addons.go` logic then sets the `cluster_wide_access` field explicitly. This sequence does not cause issues.

However, a temporary invalid Kiali CR state can occur when an existing MultiTenant SMCP is updated to ClusterWide mode. In this scenario:
- The Kiali CR already exists with `cluster_wide_access=false`.
- When the memberroll controller updates `accessible_namespaces` to `**`, the Kiali CR becomes temporarily invalid from the Kiali operator's perspective. ( because contains accessible_namespaces=** as well as cluster_wide_access=false )
- This invalid state persists until the `addons.go` logic patches the Kiali CR to set `cluster_wide_access=true`.

**Fix:**
Every time the memberroll controller sets `accessible_namespaces="**"`, it should set explicitly also `cluster_wide_access=true`

### DIFF
--- a/pkg/controller/servicemesh/memberroll/controller.go
+++ b/pkg/controller/servicemesh/memberroll/controller.go
@@ -711,7 +711,23 @@ func (r *defaultKialiReconciler) reconcileKiali(ctx context.Context, kialiCRName
 		return pkgerrors.Wrapf(err, "cannot set deployment.accessible_namespaces in Kiali CR %s/%s", kialiCRNamespace, kialiCRName)
 	}
 
-	err = updatedKiali.Spec.SetStringSlice("api.namespaces.exclude", excludedNamespacesSet.List())
+	// if the accessible namespaces include "**", set the cluster wide access flag to true
+	// since other combination (like accessible_namespaces="**" and cluster_wide_access=false) are not allowed in Kiali operator and reconsilation fails
+	if accessibleNamespacesSet.Has("**") {
+		reqLogger.Info("Updating Kiali CR deployment.cluster_wide_access to true since accessible namespaces include **", "deployment.cluster_wide_access", true)
+		err = updatedKiali.Spec.SetField("deployment.cluster_wide_access", true)
+		if err != nil {
+			return pkgerrors.Wrapf(err, "cannot set deployment.cluster_wide_access in Kiali CR %s/%s", kialiCRNamespace, kialiCRName)
+		}
+	} else {
+		reqLogger.Info("Setting Kiali CR deployment.cluster_wide_access to false (no ** in accessible namespaces)",
+			"deployment.cluster_wide_access", false)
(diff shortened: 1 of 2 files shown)


pkg/controller/servicemesh/memberroll/controller.go
```diff
@@ -711,7 +711,23 @@ func (r *defaultKialiReconciler) reconcileKiali(ctx context.Context, kialiCRName
 		return pkgerrors.Wrapf(err, "cannot set deployment.accessible_namespaces in Kiali CR %s/%s", kialiCRNamespace, kialiCRName)
 	}
 
-	err = updatedKiali.Spec.SetStringSlice("api.namespaces.exclude", excludedNamespacesSet.List())
+	// if the accessible namespaces include "**", set the cluster wide access flag to true
+	// since other combination (like accessible_namespaces="**" and cluster_wide_access=false) are not allowed in Kiali operator and reconsilation fails
```
jewertow, line 715: This comment makes sense, but I am slightly confused, because the PR description says the opposite:
```
Every time the memberroll controller sets accessible_namespaces="**", it should set explicitly also cluster_wide_access=false
```
Could you please update the PR description?

pkg/controller/servicemesh/memberroll/controller_test.go
```diff
@@ -810,6 +813,86 @@ func TestKialiResource(t *testing.T) {
...
+		{
+			name:                      "cluster-wide-access-false-when-regular-namespaces",
+			accessibleNamespaces:      []string{"bookinfo", "foo"},
+			expectedClusterWideAccess: false,
+		},
+		{
+			name:                      "cluster-wide-access-true-when-double-asterisk-with-other-namespaces",
+			accessibleNamespaces:      []string{"**", "bookinfo"},
+			expectedClusterWideAccess: true,
+		},
```
jewertow, line 836: Could we add a test case for the empty accessible namespace set?

pkg/controller/servicemesh/memberroll/controller.go
mkralik3: yes, the PR description was incorrect. Fixed

mkralik3:
@jewertow thanks for the review. Pushed changes

jewertow (approved):
LGTM

PR #1887, opened 2026-01-17
OSSM-12244: scope validating webhook management to maistra.io/managed=true

jewertow:


pkg/controller/servicemesh/webhookca/maistra2040_test.go
```diff
@@ -513,65 +260,21 @@ func webhookName(prefix, namespace string) string {
...
-	}
-}
-
 func create2xValidatingWebhook() *arv1beta1.ValidatingWebhookConfiguration {
 	return &arv1beta1.ValidatingWebhookConfiguration{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: webhookName(istiodValidatingWebhookNameTestPrefix, testNamespace),
+			Labels: map[string]string{
+				maistraManagedLabel: "true",
+			},
```
nrfox, line 270: Does this label only get added to newly created SMCPs? What about existing ones?

jewertow: The label was added to v2.6 helm chart, so it will be applied to all v2.6 webhooks - existing and newly created ones.

FilipB (approved):

